PR #8, opened 2020-03-23
Add datasource with beds by county

crwilcox:
Starting with csv for US Hospital Capacity. Below is the script I used. The only additional modification was to column names, though that could be scripted also [Colab](https://colab.research.google.com/drive/1yPvym1egd1PXEZK_MURz8TSZGW7ga12E)

```
# -*- coding: utf-8 -*-
"""
Essentially: SELECT county, state, SUM(staffed_beds), SUM(licensed_beds), SUM(icu_beds) FROM hospitals GROUP BY county, state
"""

import pandas

df = pandas.read_csv("/content/US Hospital Capacity - Not for Distribution - hospitals.csv")

df.head()

def clean_county_name(name):
  return name.split(' - ')[-1]
df['HQ_COUNTY'] = df['HQ_COUNTY'].apply(clean_county_name)

select_columns = df[['HQ_STATE', 'HQ_COUNTY', 'STAFFED_BEDS', 'LICENSED_BEDS', 'ICU_BEDS']]
select_columns = select_columns.rename({'HQ_STATE':'state', 'HQ_COUNTY':'county'}, axis='columns')
select_columns = select_columns.rename(str.lower, axis='columns')
select_columns.head()

grouped = select_columns.groupby(['state', 'county']).sum()
grouped = grouped.astype({'staffed_beds': int, 'licensed_beds': int, 'icu_beds': int})
grouped = grouped.rename(str.lower, axis='columns')
grouped.rename(str.lower, axis='index')

grouped.to_csv("hospital_beds_by_county.csv")


```


jredding:
The data doesn't need to be cleaned up on use, right? You are just showing that you did a little clean up of the data via this PR, is that correct?

crwilcox:
@jredding yep. just documenting what I did to do this for repeat use if needed again.

davidstrauss:
Awesome work!

davidstrauss:
> The data doesn't need to be cleaned up on use, right?

This should meet our needs pretty directly, but we did a little cleanup. The main goal of this transformation here is to only publish what the data source has authorized us to release. Definitive Healthcare has authorized us to release the county-level aggregates.

PR #13, opened 2020-03-25
Use Mortality Data for Infection Estimation [WIP]

davidstrauss:


notebook_runtime.ipynb
```diff
@@ -122,20 +83,263 @@
...
-    "            Dataset.get_population_by_country_state('USA', state)\n",
-    "        )\n"
+    "]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [
```
jredding: I suggest that we clear all cells prior to committing. This will make diffs easier to read. 

Unless the intent is to check in computed data.

davidstrauss:
We're mothballing this model, so closing this improvement.

PR #15, opened 2020-03-25
Automate Deploying Dataset

fridiculous:
This helps automate the serving of datasets, on a regular basis using a scheduled (hourly) lambda function to make a dataset for the DoD.

This is only configured to work on staging atm.

After an hour has passed post-deployment, the new dataset will be available publically at https://covidactnow-models-staging.s3.amazonaws.com/my-key


raygilbert:
Looks good. Is versioning enabled on the S3 bucket?

PR #20, opened 2020-03-25
Restore Binder Workflow [WIP]

davidstrauss:


davidstrauss:
This is currently broken, but we should restore it when possible.

davidstrauss:
Wow, it's working now. No changes other than re-enabling it.

davidstrauss:
Since this is suddenly working now, I'm going to merge this to restore the workflow I removed.

PR #22, opened 2020-03-25
Fix issues I hit.

mikelehen:


mikelehen:
This fixes the issues I hit (I think).

davidstrauss:
LGTM

Issue #7, opened 2020-03-22
Testing -automated and smoke tests


jredding:
We need to think about basic testing within our scripts to ensure we are creating everything correctly. A few items we can test

- Integrity of input files in the dataset 
- State count of output .json file 
- size of the .json files (i.e. not null) 
- date span (not dropping dates) 

let's start with some basic smoke tests (ex. ensuring 1+1 per model .json for each state)

ghop02:
There are a few areas I would like to address, but aren't in this PR:
 - Adding a linting github action (my guess is that we're going to want to be fairly loose with the rules)
 - Automatic code formatting - In other projects, I use a git commit hook to run `black`, but there might be a way to do it with github actions? Though I don't feel totally comfortable with automatic changes of that nature.
 - More than one test :)
 - More standardized environment setup instructions.  I personally ❤️pyenv + pyenv-virtualenv, though recognize that people run into issues with it.  Having everyone on the same python version (or in a set of python versions) reduces one more variable when debugging.
 - Should the cvoid_data_model be a proper python package? Not quite sure - I lean towards yes, but that probably depends a lot more on how/where it is used.

PR #25, opened 2020-03-26
Adding basic automated test infrastructure

ghop02:
- Updates requirements files + adds test requirements file
- Adds a SETUP.md file for instructions on setting up environment
- Adds a makefile with some basic commands.  Right now I added super loose linting, no style checks at all right now.
- Adds a github action that runs tests


.github/workflows/python_unit_tests.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
```
playermanny2: LGTM --

Question: This seems like a duplicate dependency installation as the make file also handles requirements install?

Also where is matrix.python-version env coming from?

ghop02: and It is duplicate in the sense of the code is duplicated. I can just use a make command to simplify!

Probably don't need the matrix here, but theoretically we could run on multiple python environments
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

playermanny2: Ahh i see. I'm good with this, thanks for handling this!

#7

requirements_test.txt
```diff
@@ -0,0 +1,8 @@
+# Requirements file containing packages needed for testing.
+# These should *not* be installed in non-test environments.
```
mikelehen: FWIW- I think everybody running the model should also be running tests, so I think separating out test vs. non-test requirements is probably overkill, though not particularly harmful either.  So I'm fine if you want to keep them separated.

.github/workflows/python_unit_tests.yml
```diff
@@ -0,0 +1,25 @@
+name: Python Unit Tests
```
mikelehen: Just to sanity-check, have you been able to test this somehow?  I'm not very familiar with GitHub workflows, so not sure how you go about vetting them before committing.

requirements_test.txt
ghop02: Oh definitely agree around everyone running tests.  

The non-test requirements is helpful when you start deploying to production environments.  Sometimes you could have packages that you may want locally but not on production.

crwilcox:
This LGTM in general. If folks are open to it I can make a PR to move from make to nox or tox. Mostly to help keep things to virtualenvs on folks machines. But this is a big step forward!

requirements_test.txt
mikelehen: Yeah.  There's no notion of "deploying to production" at this point.  But it doesn't hurt to keep this in place.

.github/workflows/python_unit_tests.yml
ghop02: yep - it actually runs as soon as you add a file - so you can see in the results below

ghop02:
Yeah I'm open to either! I haven't used tox too much, so will defer to you @crwilcox !

PR #28, opened 2020-03-26
Switch data source to data repo

davidstrauss:


davidstrauss:
This doesn't yet provide the ability to "pin" to a specific revision, but you can probably see how that's easy to add in a later iteration.

libs/CovidDatasets.py
```diff
@@ -277,26 +278,40 @@ def parse_country(country):
...
+            logging.info('Loading URL: {}'.format(csv_url))
+
+            # For each data file
+            try:
+                df = pd.read_csv(csv_url)
+            except urllib.error.HTTPError as e:
+                if e.code == 404:
+                    logging.info('Received a 404 for date {}. Ending iteration.'.format(snapshot_date))
+                    break
+                raise
```
davidstrauss: I attempted to use the GitHub content API, but I was having issues getting the listing for the folder. Because this only ends on 404, other errors should still get raised.

libs/CovidDatasets.py
```diff
@@ -277,26 +278,40 @@ def parse_country(country):
...
+            # Parse the 'US' entries into 'USA'
+            df[self.COUNTRY_FIELD] = df[self.COUNTRY_FIELD].apply(parse_country)
+            # Append the record dates by converting the file name into a datetime object
+            # Only process the csv files
+            day_reports.append(df)
+            snapshot_date += datetime.timedelta(days=1)
         full_report = pd.concat(day_reports)  # Concat said reports into a single DataFrame
         full_report = full_report.reset_index()
+
+        full_report.info()
```
mikelehen: Just double-checking we want to leave this in?  (I'm not at all opposed; just flagging in case it was meant to be temporary for debugging)

libs/CovidDatasets.py
mikelehen: Sounds good.

I wonder a bit if fetching from raw.githubusercontent.com is the best option long-term.  Besides this enumeration problem, it may be a bit slow. A couple potential alternatives:
* Have a shell script or similar that clones the covid-data-public repo (at a revision) and then points run.py and other scripts at that location. Then we can do (fast) local filesystem access.
* We could _consider_ embedding the covid-data-public repo in covid-data-model as a git submodule, but TBH I don't know how suitable that is and the learning curve on git submodules may make it a subpar solution, even if it would work.

But this seems fine to me for now.

PR #29, opened 2020-03-26
Update dod.py to also generate states.csv

gsoltis:
County data is aggregated by state, lat/lng comes from first county in a state. Intervention data is pulled from master in covid-projections repo (will need to be updated w/ monorepo change), and projections is currently pulled from checked-in csv.


dod.py
```diff
@@ -55,3 +75,42 @@
...
+state_col_remap = {
+    'state_x': 'Province/State',
+    'Projection1': '4-day Hospitalizations Prediction',
+    'Projection2': '8-day Hospitalizations Prediction',
+    'intervention': 'Intervention'
+}
+states_remapped = states_abbrev.rename(columns=state_col_remap)
+states_remapped.info()
+# TODO: filter out county-specific columns
+state_cols = cols + ['Intervention', '4-day Hospitalizations Prediction', '8-day Hospitalizations Prediction']
```
fridiculous: to me, this looks like there is a standard interface for these columns, and that interface should be seperated from the logic above

dod.py
```diff
@@ -1,14 +1,34 @@
...
+abbrev_df = pd.DataFrame(
+    list(us_state_abbrev.items()),
+    columns=['state', 'abbreviation']
+)
+abbrev_df.info()
+
+projections_df = pd.read_csv('projections_03-26-2020.csv')
+projections_df.info()
+
 url = 'https://github.com/CSSEGISandData/COVID-19/raw/master/csse_covid_19_data/csse_covid_19_daily_reports/{}.csv'.format(
```
fridiculous: up to this point, it looks like it's mostly obtaining data.  i can definitely see a helper method.

PR #31, opened 2020-03-27
Fix JHU Calculation 03/22-Current

crwilcox:
Later time ranges are formatted as counties. As we were using the last state format date, we were stopping soon.

### DIFF
--- a/libs/CovidDatasets.py
+++ b/libs/CovidDatasets.py
@@ -204,6 +204,7 @@ class Dataset:
             if len(state_data[state_data[self.DATE_FIELD] == curr_date]) == 0:
                 county_data_for_date = copy(county_data[county_data[self.DATE_FIELD] == curr_date])
                 if len(county_data_for_date) == 0:  # If there's no county data, we're SOL.
+                    logging.info("NO COUNTY DATA: {}".format(curr_date))
                     continue  # TODO: Revisit. This should be more intelligent
                 county_data_for_date = county_data_for_date.iloc[0]
                 new_state_row = copy(state_data.iloc[0])  # Copy the first row of the state data to get the right format
@@ -263,11 +264,12 @@ class JHUDataset(Dataset):
         # Compile a list of all of the day reports available
         def parse_county(state):
             if ',' in state:
-                return state.split(',')[0]
+                return state.split(',')[0].strip()
 
         def parse_state(state):
+            state = state.strip()
             if ',' in state:
-                state = state.split(',')[1]
+                state = state.split(',')[1].strip()
             if state in us_state_abbrev:
                 return us_state_abbrev[state]
             return state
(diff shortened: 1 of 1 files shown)


libs/CovidDatasets.py
```diff
@@ -196,14 +196,15 @@ def combine_state_county_data(self, country, state):
             [self.DATE_FIELD, self.COUNTRY_FIELD, self.STATE_FIELD], as_index=False
         )[[self.CASE_FIELD, self.DEATH_FIELD, self.RECOVERED_FIELD]].sum()
         # Now we fill in whatever gaps we can in the state data using the county data
-        curr_date = state_data[self.DATE_FIELD].max()  # Start on the last date of state data we have
+        curr_date = max(state_data[self.DATE_FIELD].max(), county_data[self.DATE_FIELD].max()) # Start on the last date of state data we have
```
tsebens: This is an interesting thought, one I hadn't considered.

In general we try to use the state data as our first choice and fill in the holes with county data. I don't imagine that's likely for the county data goes later than the state data, but I think we should have a plan in place for when it does. 

I think that we should end when the state data ends, because we are projecting at a state level, and often times summing counties provides a lower total than the state data. In this highly hypothetical situation, it is likely that the final cycle or two of just county data would artificially drag down the model.

In other words, I don't think we should do this, just wanted to talk through why 😄

libs/CovidDatasets.py
```diff
@@ -196,14 +196,15 @@ def combine_state_county_data(self, country, state):
...
         county_data_to_insert = []
         while curr_date > self._START_DATE:
             curr_date -= datetime.timedelta(days=1)
             # If there is no state data for a day, we need to get some country data for the day
             if len(state_data[state_data[self.DATE_FIELD] == curr_date]) == 0:
                 county_data_for_date = copy(county_data[county_data[self.DATE_FIELD] == curr_date])
                 if len(county_data_for_date) == 0:  # If there's no county data, we're SOL.
+                    logging.info("NO COUNTY DATA: {}".format(curr_date))
```
tsebens: I think we should take this one step further and raise an exception. If we reach this point in execution, that means that for a certain date, there is no state data and no county data. That's a serious problem, which _will_ crash the model later on in execution, 100% of the time.

The exception should also indicate the date for which there is no state or county data

libs/CovidDatasets.py
davidstrauss: > summing counties provides a lower total than the state data

It also sometimes yields a higher one, as is the case in Washington state.

davidstrauss: For JHU, though, I think each data set has exclusively been states or counties. I could be wrong, though.

libs/CovidDatasets.py
davidstrauss: I don't want to block merging on this, but I'll open an issue for how we handle this concern. I'm leaning toward not throwing an exception and having the consumer of the data be responsible for ensuring it has the necessary data points to operate.

davidstrauss: The issue: https://github.com/covid-projections/covid-data-model/issues/35

PR #34, opened 2020-03-27
Create local copy of data to avoid repeat downloads from GitHub

crwilcox:
While trying to verify changes with `run.py` I encountered high failure rates for downloading from GitHub. There are a large number of repeated downloads for the same run. Rather than downloading separately, download a copy of the public data once at the beginning of a run.


libs/CovidDatasets.py
```diff
@@ -67,14 +70,39 @@
...
+public_data_base_url = 'file://localhost' + local_public_data.name + '/covid-data-public-master'
+
+
+def create_local_copy_public_data():
+    """
+    Creates a local copy of the public data repository. This is done to avoid
+    downloading the file again for each intervention type.
+    """
+    github_zip_url = "https://github.com/covid-projections/covid-data-public/archive/master.zip"
+    logging.info("Creating a Local Copy of {}".format(github_zip_url))
```
ghop02: Can use f-strings here~ `"fCreating a Local copy of {github_zip_url}"`

libs/CovidDatasets.py
```diff
@@ -238,7 +266,9 @@ def get_beds_by_country_state(self, country, state):
 class JHUDataset(Dataset):
     # The date of the first JHU data snapshot.
     _FIRST_JHU_DATE = datetime.date(2020, 1, 22)
-    _JHU_URL_TEMPLATE = 'https://raw.githubusercontent.com/covid-projections/covid-data-public/master/data/cases-jhu/csse_covid_19_daily_reports/{}.csv'
+    if not local_public_data_populated:
```
ghop02: Can you run this somewhere else than in the class definition? It would surprise me if I imported this class and it started downloading data

libs/CovidDatasets.py
```diff
@@ -288,6 +319,11 @@ def parse_country(country):
                     logging.info('Received a 404 for date {}. Ending iteration.'.format(snapshot_date))
                     break
                 raise
+            except urllib.error.URLError as e:
+                if isinstance(e.reason, FileNotFoundError):
+                    logging.info('Received a 404 for date {}. Ending iteration.'.format(snapshot_date))
```
ghop02: Does this actually log an error message? 

Update: Whoaaa didn't realize that it just logged to the root logger.

No real need, but would be good to make a logger instance for this module.  I generally do:

```
```
right after import statements at the top.

libs/CovidDatasets.py
```diff
@@ -67,14 +70,39 @@
...
+
+    local_public_data_populated = True
 
 
+class Dataset:
+    """Base class of the Dataset objects. Standardizes the output so that data
+    from multiple differenmt sources can be fed into the model with as little
+    hassle as possible."""
+
+    _POPULATION_URL = public_data_base_url + '/data/misc/populations.csv'
```
ghop02: It would be great if we could pass this in as an argument to the init function (since it can change).

I've sometimes made these default arugments:

```
class Dataset:
    def __init__(self, start_date, filter_past_date=None, population_url=DEFAULT_URL):
        self.population_url = population_url
        ...

```

run.py
```diff
@@ -122,10 +127,11 @@ def model_state(country, state, interventions=None):
     },
 ]
 
-Dataset = CDSDataset()
+Dataset = _Dataset()
```
ghop02: Can we try to reserve capitalized variables for ClassNames and change this variable from `Dataset` -> `dataset`

I realize this was there before, but helps make it clear when you're reading code whether or not you're accessing a static/classmethod `Dataset.get_something` vs an instance method `dataset.get_something`.

libs/CovidDatasets.py
tsebens: I agree. It should probably be a conditional call in each of the get_data functions (beds, population, timeseries) whichever one calls it first downloads the local copy

libs/CovidDatasets.py
crwilcox: Oh, can I? I am game, I just was concerned perhaps someone was using an older interpreter.

ghop02: Yeah! I think we're all down for 3.7+

libs/CovidDatasets.py
crwilcox: I had thought about this, but didn't want to make the call. The other thought I had was to use an env_var to allow switching. It is pretty easy to shift this around.

libs/CovidDatasets.py
crwilcox: Can make this change.

run.py
crwilcox: I saw this. I was trying not to fix too many things I wasn't touching :) I can refactor this.

libs/CovidDatasets.py
crwilcox: I moved the url behind a getter. on first get it will be downloaded.

davidstrauss:
This implementation looks good, but should we consider a higher-level approach to this problem by wrapping the HTTP request in a cache? If we used Requests to fetch the files, we could choose to cache [according to RFC](https://httpcache.readthedocs.io/en/latest/) or [just for an arbitrary length of time](https://pypi.org/project/requests-cache/).

run.py
crwilcox: I dropped these changes from this pr.

crwilcox:
@tsebens previous to this I was using a context manager variety of temp directory.I added back now that I got the rest settled. This will create at the start and delete once out of scope. 

@davidstrauss I like the idea of httpcache, though I think read_csv is using urlib directly, not requests.

tsebens:
👍

PR #37, opened 2020-03-27
Adding a proper CLI around run, load data urls from property

ghop02:
Loading the various Data URLs cased the root logger to get called before calling in import.  
This messed up logging. 

As a fix, I moved the urls to instance properties rather than on the class.  This delays calling the function to download the data locally.  

Also - I added a slight cli wrapper around the run-model function.  I found it a bit tricky to know what i was actually running.  This allows you to choose a couple dataset sources and change the output path.

```
./run.py run-model
```
will do the trick.


ghop02:
I updated this to remove the `run.py` changes.  This should be independent of #38 now.

The primary change is to not download the dataset on module import

requirements.txt
```diff
@@ -6,3 +6,4 @@ plotly==4.5.4
 jupyter==1.0.0
 simplejson==3.17.0
 boto3==1.12.28
+click==7.1.1
```
gsoltis: This probably belongs in the next change w/ the `run.py` changes.

ghop02: ah yes. foreshadowing :)

PR #38, opened 2020-03-27
Add a sanity check for runs of run.py

gsoltis:
`validate.py` wraps a run of `run.py`, validating that `run.py` does not crash and produces the expected number of non-empty result files. 

Will need to be updated as support for additional territories are added (Guam, American Samoa, Northern Mariana Islands)

### DIFF
--- a/run.py
+++ b/run.py
@@ -4,6 +4,7 @@ logging.basicConfig(level=logging.INFO)
 import datetime
 import time
 import simplejson
+from libs.build_params import r0, OUTPUT_DIR, INTERVENTIONS
 from libs.CovidTimeseriesModel import CovidTimeseriesModel
 from libs.CovidDatasets import CDSDataset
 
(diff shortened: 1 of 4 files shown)


gsoltis:
I shuffled a few things out of `run.py` and `CovidDatasets.py` so I could import them without triggering downloads of the datasets.

run.py
```diff
@@ -4,6 +4,7 @@
 import datetime
 import time
 import simplejson
+from libs.build_params import r0, OUTPUT_DIR, INTERVENTIONS
```
ghop02: I would recommend just importing the module:
`from libs import build_params`

it's helpful for keeping track of where variables come from

gsoltis:
Might be better to wait for https://github.com/covid-projections/covid-data-model/pull/37 to land, then we can probably do this more easily w/o using a subprocess

validate.py
```diff
@@ -0,0 +1,71 @@
...
+    os.unlink(os.path.join(result_dir, f))
+
+
+UNSUPPORTED_REGIONS = [
+  'AS',
+  'GU',
+  'MP'
+]
+
+def validate_results(result_dir: str) -> None:
```
ghop02: This is great, I think it's worth figuring out how to run this regularly in pytest as part of the normal tests.  But I think we can do that in another PR

run.py
gsoltis: Done.

validate.py
gsoltis: Sounds good, will do

validate.py
```diff
@@ -0,0 +1,71 @@
+import argparse
+from datetime import datetime
+import os
+import subprocess
+import sys
+
+from libs import build_params
+
+def run_run_py(datasource: str) -> None:
```
mikelehen: Run, run!

PR #39, opened 2020-03-27
Snapshot testing

tsebens:
Snapshot framework to compare the model's current output to outputs that it's had in the past. Also added a method to easily update the snapshots.

Ran the tests and everything passes.


jredding:
Do we need to be checking in the test data? This is PR is 225 files large ;).

test/CovidTimeseriesModel.py
```diff
@@ -47,21 +48,79 @@ def test_calculate_r(self):
...
+
+    def _build_test_parameter_sets(self):
+        dataset = CDSDataset()
+        states = dataset.get_all_states_by_country('USA')
+        MODEL_INTERVAL = 4
+        sets = []
+        for state in states[:2]:
+            for intervention in self.interventions:
+                sets.append(
+                    CovidUtil().initialize_model_parameters({
```
jredding: These would need to match up with whatever is in run.py for the pipeline and I can imagine them getting out of sync over time. For now probably fine, but I anticipate that this will needed to be updated - I also know there was talk about having different assumptions for each location, which could change dramatically based on location.

tsebens: Those files were from the old snapshot framework, which depended on them. I removed them because the new framework uses a different format (with a single file). I would argue that removing them simultaneous to the snapshot testing refactor is appropriate.

tsebens:
All those files being deleted were from the old snapshot framework, which depended on them. I removed them because the new framework uses a different format (with a single file). I would argue that removing them simultaneous to the snapshot testing refactor is appropriate.

davidstrauss:
This is definitely something I'd like to land as part of supporting the new modeling work.

fridiculous:
@tsebens Can we revisit this PR in the context of the changes being proposed here? https://docs.google.com/document/d/1VhQqMxtVYHQMIj9zFeb1DUQZ1aq1k4xIc_bYoTIqUb4/edit?ts=5e8a3a1e#

fridiculous:
Closing as this branch is no longer relevant

PR #40, opened 2020-03-27
Update run function and add test

ghop02:
A few things:
 - Moving validate code into run cli + libs/validate.py
 - Adding ability to run by state
 - Added a test that runs one state and make sure it works from the cli.

I have to step away for a bit - need to fully test the validate functionality.


mikelehen:
Sorry, but for me to review this it needs to be split into smaller, more isolated PRs.  As general rules:
1) PRs should be small and do only one thing.
2) You should always try to split out pure refactors (renaming files / variables / etc.) from actual code changes.  Otherwise the refactors add a lot of noise to the diff that make it easy to overlook actual code changes.  If I know all you did is rename a variable in a bunch of places, it's easy to scan through and sanity-check.  But if you also made other code changes, it's a lot more tedious and error-prone. 
 And in this case, you changed validate.py enough that git doesn't even recognize it as a rename anymore, so I have to try to manually compare them.

If your work builds upon itself so that you can't easily do it as separate PRs against master in parallel, you can either created PRs chained off of each other (which admittedly is a little awkward with github) or do it in a single PR but make sure that the PR consists of separate commits for each reviewable piece, e.g.:

* Rename files to my preferred directory structure / naming convention.
* Rename constants to my preferred naming convention.
* Switch to a dedicated logger instead of logging.info() directly.
* Add comments to significant methods.
* Switch to click for command-line processing.
* ...

Then I can quickly click through the commits and review them in isolation.

ghop02:
Hi Michael, thanks for the feedback! Will do!

gsoltis:
The changes LGTM, but I will defer to Michael's feedback.

ghop02:
Will end up fixing these in other PRs!

PR #41, opened 2020-03-28
Add shapefile producing functions to build_dod_dataset.py

vincentwoo:
Hi this is my first PR so I tagged a few reviewers.

This PR does a few things:

* Imports a new library, pyshp, to handle reading and writing shapefiles
* Commits ~150MB of data into Git LFS that comprises, effectively, a map of the states and counties of the USA. This comes from the census.
* Adds a couple functions to `build_dod_dataset.py` that uses the above to produce what are effectively copies of the census shapefiles with additional COVID-19 information. The information in question is basically exactly what was being produced in the CSVs for the DoD previously.


vincentwoo:
Also, image for flavor

image

davidstrauss:
We're making some effort to retire the "data" folder in this repository in favor of tracking raw or minimally processed data in [a new repository](https://github.com/covid-projections/covid-data-public). Do you have any ideas for making this work better with that approach?

vincentwoo:
Ya it could work, but you'd sort of have to hardcode an assumption about where the other repo is gonna be. You want I should do that and open 2 PRs?

davidstrauss:
> Ya it could work, but you'd sort of have to hardcode an assumption about where the other repo is gonna be.

I think that's okay, as the repository is intended as a library of raw data for remote access.

> You want I should do that and open 2 PRs?

Let's see what others think about this, but I'm leaning towards yes.

vincentwoo:
K lemme know - my assumption right now is `covid-data-model` and `covid-data-public` are sibling directories

fridiculous:
Yea, open 2 PRs.  This looks great, i'm merging it in.

igorkofman:
@davidstrauss is the plan to read from the data repo locally or via https? cc @mikelehen

mikelehen:
I think I'm out-of-the-loop on the DoD stuff at this point.  All else being equal, I think accessing via https is nicer. But I'm not sure exactly what we're talking about.

PR #45, opened 2020-03-28
Adding click cli to run

ghop02:
This updates `./run.py` to use click - a great CLI interface developed by the same person (or team) as flask.

I've found it much easier than using argparse - and something more lightweight than invoke (if anyone has used that before).

This just updates the main function to let you run the model using a couple easier steps.

For example, running on a single state:
```
./run.py run-model --state CA
```
Run model using CDS data
```
./run.py run-model CDS
```
Run Model using JHU data:
```
./run.py run-model JHU
```
Run model to a different output dir:
```
./run.py run-model -o my_favorite_dir
```


run.py
```diff
@@ -1,11 +1,23 @@
+#!/usr/bin/env python
+"""
+Run Covid data model.
+Execution:
+    ./run.py run-model [JHU|CDS] [--output-dir]
```
mikelehen: 1. Why do I need `run-model`?  What else does this script do?  Ideally it would do only one thing, I think. :-)
2. Why is it `JHU` and `CDS` instead of `--jhu` and `--cds` ?  I don't care super strongly, but I'm not sure if it makes sense.  FWIW we may end up with multiple model implementations, more datasets to choose from, and end up writing code to merge/blend them, etc.  So I don't know exactly how this will evolve.
3. From a documentation perspective, should this be `[--output-dir path]` to make it clear you have to pass an additional arg after `--output-dir`?

run.py
```diff
@@ -74,16 +85,49 @@ def model_state(country, state, interventions=None):
...
+
+
+@main.command()
+@click.argument('dataset_name', type=click.Choice(['JHU', 'CDS']), default="CDS")
+@click.option(
+    '--output-dir', '-o', default='results/test', help="Model results output directory"
+)
+@click.option('--country', default='USA')
+@click.option('--state', '-s', multiple=True)
+def run_model(dataset_name, output_dir, country, state):
```
mikelehen: I guess `@click` is doing some magic to automatically run this function if you pass `run_model` in on the command-line?  That's neat, but is non-obvious (to somebody like me at least).  I have a slight preference against using libraries that do too much magic until you need to.

Though if we rework the CLI interface to remove "run-model" from the command-line, as I suggested above, maybe this will go away?

In any case, I'm kind of on the fence about this...

run.py
```diff
@@ -74,16 +85,49 @@ def model_state(country, state, interventions=None):
...
+    # filter_past_date = datetime.date(2020, 3, 19)
+    filter_past_date = None
```
mikelehen: Why did you change this to None?  (this is the kind of change I don't want to accidentally sneak in along with refactoring changes, etc. 😛 )

run.py
mikelehen: It looks like there's a state parameter too, but you haven't documented that here?

test/run_cli_test.py
```diff
@@ -0,0 +1,27 @@
...
+from libs import build_params
+
+
+@pytest.mark.parametrize('dataset_name', ['JHU', 'CDS'])
+def test_runs_all_interventions_on_one_state(tmpdir, dataset_name):
+    runner = CliRunner()
+
+    runner.invoke(run.main, [
+        'run-model', dataset_name, '--state', 'MA', '--output-dir', tmpdir
+    ], catch_exceptions=False)
```
mikelehen: FWIW I think this test is more likely to slow us down (as we add additional models, rework how datasets work, etc.) than to catch a regression.

In other words, I expect the "spec" for run.py to be rapidly evolving.  I'm not sure it's the best place to be testing right now.

run.py
ghop02: 1. Down to remove that! I think i was intending on providing a single entry-point for various commands.  Other examples would be `.run.py validate` - validates data.  `./run.py download-local-data`.  
2. Generally the flags there would be additive.  For instance right now it wouldn't make sense to have `./run.py run-model --jhu --cds`.  Another solution would be `./run.py run-model --dataset [jhu|cds]`
3. yup!

It may not be good to keep the example here, i've found that it tends to get out of date quite quickly.  Running `./run.py run-model --help` shows the various command options.

run.py
ghop02: heh, in the original implementation, there were multiple calls to create the dataset:
In model_state:
```
Dataset = CDSDataset(filter_past_date=datetime.date(2020, 3, 19))
```
in the main function:
```
Dataset = CDSDataset()
```
This gives one source of truth for the dataset.

What I wanted to do was to add another input, `--filter-past-date` but didn't want to spend time with that parsing logic (also, in python I love using Arrow for datetimes, but that is much more controversial than I wanted to point to)

test/run_cli_test.py
ghop02: Yeah I go back and forth - this test feels high level enough to me to serve as a stop gap - I think we should make sure that no changes break the basic command line run function.

I think the test is easy enough to update that there shouldn't be a ton of maintenance? happy to figure out another way to go about it though

ghop02:
I'm going to close this for now until we reach a bit more stability with outstanding models.  May revisit at a later point!

PR #56, opened 2020-03-29
Add logging for each intervention, only use one dataset source

ghop02:


ghop02:
Oh there will be some merge conflicts - but there already are i believe. I can do that though!

gsoltis:
Adding my +1 to have this merged, as the current data in `covid-data-public` breaks `run.py`

PR #59, opened 2020-03-30
New Dataset Format + County level data [WIP]

ghop02:
There is still a bit more validation to do here, but wanted to get this up.

This will essentially deprecate `CovidDatasets.py`, replacing them with more separated classes.

This splits up the dataset source (JHU, CDS, etc.) from the dataset that the Model queries (in this PR, called a `TimeseriesDataset`).  

I modified run.py to use county data - the model runs! 

Will need to verify + do a bit more cleanup tomorrow, but think this is in a good spot!


ghop02:
Closing this in favor of #64

Issue #60, opened 2020-03-30
Current method of copying covid-public-data locally doesn't support Git LFS 


gsoltis:
Downloading a zip of the repository downloads the pointers to large files tracked with git lfs, but not the files themselves. Since the zip is also not itself a repository, it is not immediately possible to fetch the large files. We should consider either what needs to be done to fetch the large files after the zip is downloaded, or the possibility of cloning enough of the repo to also download the large files as we need them.

This is currently breaking support for the `COVID_DATA_PUBLIC` environment variable on the `dod` branch, since it depends on shapefiles stored via git lfs in `covid-data-public`.

gsoltis:
The current plan is:
 * when COVID_DATA_PUBLIC var is set, do nothing, use value of the var for the path
   * The person setting this path will be responsible for rehydrating lfs objects
 * when COVID_DATA_PUBLIC is *not* set, do a full clone and rehydrate lfs data, then set the path to that.

gsoltis:
Eventually, we need to add support for git urls to the environment variable so that we can point tools at a specific hash and get the correct behavior

davidstrauss:
Closing this for two reasons:

- I believe we now implement `COVID_DATA_PUBLIC` as specified here.
- The runtime of the current model is sufficiently long that optimizing raw data sourcing won't substantially improve the ability to run the current model "off the shelf" in a quick way (which is the main limitation to the current system).

PR #61, opened 2020-03-30
Try using shallow clone to fetch repo for cache

gsoltis:
@vincentwoo Can you give this a try and see if it works for you? Note the new dependency to add as well in `requirements.txt`. I switched to cloning the repo for the local cache, rather than downloading a zip file. I *think* that fixes the shapefile / git-lfs issues, but I also didn't see an exception earlier, so I'd like to verify that it's working.

DO NOT MERGE. Still requires verification.

Tries to address #60

### DIFF
--- a/libs/CovidDatasets.py
+++ b/libs/CovidDatasets.py
@@ -1,4 +1,5 @@
 import datetime
+import git
 import logging
 import math
 from copy import copy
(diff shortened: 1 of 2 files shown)


gsoltis:
Closing in favor of https://github.com/covid-projections/covid-data-model/pull/62

PR #64, opened 2020-03-31
52 county level data

ghop02:


libs/datasets/data_source.py
```diff
@@ -0,0 +1,51 @@
...
+    POPULATION_FIELD_MAP = None
+
+    # Name of dataset source
+    SOURCE_NAME = None
+
+    def __init__(self, data: pd.DataFrame):
+        self.data = data
+
+    @classmethod
+    def build_from_local_github(cls) -> "cls":
```
crwilcox, line 34: bit of a nit, but possibly just `local_git` or maybe just `local`? The source doesn't have to be a git clone right?

libs/datasets/dataset_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    jhu_usa_data = data.get_subset(
+        AggregationLevel.COUNTY, country="USA", after="2020-03-01"
+    ).data
+
+    data = cds_data_source.to_generic_timeseries()
+    cds_usa_data = data.get_subset(
+        AggregationLevel.COUNTY, country="USA", after="2020-03-01"
+    ).data
+
+    # TODO(chris): Better handling of counties that are not consistent.
```
crwilcox, line 74: worth filing an issue?

run.py
```diff
@@ -1,11 +1,19 @@
...
-import simplejson
+import logging
 import time
 import datetime
+
+import simplejson
 import pandas as pd
-import logging
+
+# from libs.CovidDatasets import CDSDataset, JHUDataset
```
crwilcox, line 8: remove?

libs/datasets/sources/dh_beds.py
```diff
@@ -0,0 +1,145 @@
...
+        for state, county, fips in fips_combos
+    }
+    counties_by_state = defaultdict(list)
+
+    for state, county, _ in fips_combos:
+        counties_by_state[state].append(
+            county.lower()
+            .replace("'", "")
+            .replace("-", " ")
+            .replace("ó", "o")
```
crwilcox, line 36: would it be worth prefetching on some other accented chars? 
Also, is there a reason ó => o but ñ => ""? At least at glance it seems it should also go to a non-accented form?

ghop02: oh yeah, this was super hacky - I'll look into a better solution for this.  

https://stackoverflow.com/a/2633310

Huh this is cool (example from SO):
```python
accented_string = u'Málaga'
# accented_string is of type 'unicode'
import unidecode
unaccented_string = unidecode.unidecode(accented_string)
# unaccented_string contains 'Malaga'and is of type 'str'
```

crwilcox:
Took another pass, looking good.

libs/datasets/sources/dh_beds.py
```diff
@@ -0,0 +1,145 @@
...
+        data[cls.Fields.AGGREGATE_LEVEL] = "county"
+
+        # Backfilling FIPS data based on county names.
+        # TODO: Fix all missing cases
+        fips_data = dataset_utils.build_fips_data_frame()
+        data = match_county_to_fips(data, fips_data)
+        return data
+
+    @classmethod
+    def local(cls) -> "JHUTimeseriesData":
```
davidstrauss, line 143: Why does this beds dataset map to JHUTimeseriesData here?

ghop02: ah it does not, copy pasta error

PR #66, opened 2020-03-31
Continued SEIR Refinement

davidstrauss:


run.py
```diff
@@ -102,70 +102,61 @@ def record_results(
...
-        # If True use the harvard parameters directly, if not calculate off the above
-        "use_harvard_params": True,
-        # If True use the parameters that make R0 2.4, if not calculate off the above
-        "fix_r0": False,
-        # If True use the harvard model inputs for inital conditions and N (recreate their graph)
-        "use_harvard_init": False,
+        "population": dataset.get_population_by_country_state(country, state),
+    }
+
+    MODEL_PARAMETERS = {
```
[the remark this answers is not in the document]
ghop02: Fwiw, I think this would be a great case for [dataclasses](https://docs.python.org/3/library/dataclasses.html)
can define it similar to a class:

```
from dataclasses import dataclass

@dataclass
class ModelParameters(object):
    model: str = "seir"
    use_harvard_params: bool = False
    fix_r0: bool = False

    @property
    def case_fatality_rate_hospitals_overwhelmed(self):
        return self.hospitalization_rate / self.hospitalized_cases_requiring_icu_care

# Generates defaults but overriding fix_r0
model_parameters = ModelParameters(fix_r0=True)
```

Definitely not advocating for the change here, but wanted to put the idea out there :).
If you were interested, it would definitely be better as it's own pr, but it does give a bit more structure and easier documentation!

libs/CovidTimeseriesModelSIR.py
```diff
@@ -243,93 +248,93 @@ def iterate_model(self, model_parameters):
...
+        )
 
-            # move the actual infected numbers into infected_a where its NA
-            combined_df["infected_a"] = combined_df["infected_a"].fillna(
-                combined_df["infected"]
+        if model_parameters["model"] == "seir":
+
+            # make infected total represent the sum of the infected stocks
+            # but don't overwrite the historical
+            combined_df.loc[:, "infected_tmp"] = (
```
ghop02: I don't think you need to add this to the dataframe:

```
infected_tmp = (...)
combined_df.loc[:, "infected"].fillna(infected_tmp)
```
should work!

run.py
```diff
@@ -103,70 +101,64 @@ def write_results(data, directory, name):
...
+        # hospitalized = case load from timeseries on last day of data / 4
+        # mild = hospitalized / hospitalization_rate
+        # icu = hospitalized * hospitalized_cases_requiring_icu_care
+        # expoosed = exposed_infected_ratio * mild
+        "presymptomatic_period": 3,  # Time before exposed are infectious, In days
+        "duration_mild_infections": 6,  # Time mildly infected poeple stay sick, In days
         "hospital_time_recovery": 11,  # Duration of hospitalization, In days
         "icu_time_death": 7,  # Time from ICU admission to death, In days
-        "case_fatality_rate": 0.0109341104294479,
         "beta": 0.5,
```
maxwhenderson, line 139: `"beta": 0.6,`

libs/CovidTimeseriesModelSIR.py
traethethird: This infected sum does change for each row... the thing I'm trying to achieve is fill infected for the modeled rows without overwriting infected in the historical rows where the _a _b _c stocks are empty (thus causing the historical to read 0). Fillna from the column is clunky, but the best way I could think to achieve that.

PR #69, opened 2020-04-01
Tweaks from Igor / Max

igorkofman:


traethethird:
As discussed, these interventions can result in improbable or in some cases even negative betas, but approving the merge from a code perspective

PR #79, opened 2020-04-02
Fixing pool processing

ghop02:


run.py
```diff
@@ -20,7 +20,12 @@
 
 _logger = logging.getLogger(__name__)
 
-pool = mp.Pool(max(mp.cpu_count() - 1, 1))
+
+def get_pool(num_cores=None) -> multiprocessing.Pool:
+    if not num_cores:
+        num_cores = max(multiprocessing.cpu_count() - 1, 1)
```
fridiculous, line 26: can you do `-2` instead of `-1`

PR #80, opened 2020-04-02
Add easier cli for model

ghop02:
This adds a more direct cli - couple advantages:
 1. Can choose if you want to run the county or state model.
 2. Can pass `--deploy` flag which will save files to the public data directory in `covid-projections`
 3. You can choose a specific state to run on

example:
```
./run_model.py state --deploy
```

### DIFF
--- a/run_model.py
+++ b/run_model.py
@@ -5,7 +5,7 @@ import logging
 import click
 from libs import build_params
 import run
-
+print(run)
 WEB_DEPLOY_PATH = "../covid-projections/public/data"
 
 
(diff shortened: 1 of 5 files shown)


ghop02:
One thing I wasn't sure about - do we want this in run.py? Should the file be named differently? fine moving it there too - i just don't love how crowded run.py is getting.

run_model.py
```diff
@@ -0,0 +1,62 @@
...
+def run_state(state=None, deploy=False):
+    """Run State level model."""
+    min_date = datetime.datetime(2020, 3, 7)
+    max_date = datetime.datetime(2020, 7, 6)
+
+    output_dir = build_params.OUTPUT_DIR
+    if deploy:
+        output_dir = WEB_DEPLOY_PATH
+
+    run.run_county_level_forecast(
```
crwilcox, line 55: state level?

fridiculous:
tested and validated

PR #83, opened 2020-04-02
Add support for a COVID_PUBLIC_DATA_HASH env variable

gsoltis:
Uses a `contextmanager` to set the local copy of covid-data-public to the specified hash, if one is supplied. Resets to the existing branch on exit, and errors if the working tree is not clean.

Currently only supported for `run.py`, but if there's interest I could wire it into other entry points as well.

### DIFF
--- a/run.py
+++ b/run.py
@@ -16,7 +16,7 @@ from libs.build_params import OUTPUT_DIR, get_interventions
 from libs.datasets import JHUDataset
 from libs.datasets import FIPSPopulation
 from libs.datasets import DHBeds
-from libs.datasets.dataset_utils import AggregationLevel
+from libs.datasets.dataset_utils import AggregationLevel, public_data_hash
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -384,9 +384,11 @@ def run_state_level_forecast(min_date, max_date, country="USA", state=None):
 
 if __name__ == "__main__":
     logging.basicConfig(level=logging.INFO)
-    # @TODO: Remove interventions override once support is in the Harvard model.
-    min_date = datetime.datetime(2020, 3, 7)
-    max_date = datetime.datetime(2020, 7, 6)
-    # build_county_summary()
-    run_county_level_forecast(min_date, max_date)
-    run_state_level_forecast(min_date, max_date)
+    with public_data_hash(os.getenv('COVID_DATA_PUBLIC_HASH', None)) as git_hash:
```
ghop02, line 387: Can you add this to `run_model.py` as well?

gsoltis: Assuming you mean `run_old_model.py`, I don't think this exact approach will work the same way, since without `COVID_DATA_PUBLIC` also being specified, that code will clone a fresh repo and use that. I can look at unifying the two approaches in another PR though.

PR #85, opened 2020-04-02
Take max of licensed,staffed beds

ghop02:


ghop02:
This was fixed in #82

igorkofman:
Was it? I thought this was about taking max?

PR #97, opened 2020-04-03
Sgoldblatt/countie projection data

igorkofman:


vincentwoo:
this PR adds a gitignore for various files (counties.csv) but also checks them in - the latter is probably a mistake

libs/build_dod_dataset.py
```diff
@@ -135,6 +256,9 @@ def get_usa_by_county_df():
     # handle serializing FIPS without trailing 0
     final_df['State/County FIPS Code'] = final_df['State/County FIPS Code'].astype(str).str.replace('\.0','')
 
+    # There should be a leading zero (FIPS codes are of length 5). Ideally this would be a char
+    final_df['State/County FIPS Code'] = final_df['State/County FIPS Code'].astype(str).str.rjust(5, '0')
```
vincentwoo, line 260: if you want this behavior I think you should just delete the line above, that way we dont have to do any padding

libs/build_dod_dataset.py
```diff
@@ -203,33 +327,43 @@ def get_usa_by_states_df():
...
+    # Just adding some understanding of errors
+    failed_dictionary = {}
+
     for shapeRecord in shapeRecords:
         try:
+            # Gets the row of the dataframe that matches the FIPS codes for a state/county
             row = df[df[pivot_df_column] == shapeRecord.record[pivot_shp_field]].iloc[0]
-        except:
+        except Exception as e:
+            state_fips = shapeRecord.record[pivot_shp_field][:2] # state fips is the first two chars of the state/county fips
```
vincentwoo, line 354: the FIPS for the state is also available as a separate entry (if you dont want to string split)

run2.py
```diff
@@ -0,0 +1,69 @@
+import logging
+import argparse
+
+from libs.build_params import OUTPUT_DIR, get_interventions
+from datetime import datetime, timedelta
+from libs.datasets import JHUDataset
+from libs.datasets.dataset_utils import AggregationLevel
+import os.path
+import simplejson
+from libs.us_state_abbrev import us_state_abbrev, us_fips
```
fridiculous, line 10: should we delete this file?

.gitignore
```diff
@@ -3,7 +3,17 @@ results/*.csv
...
+results/county/*
+
+# don't upload csvs/files
+counties.csv
+counties.dbf
+counties.shp
+counties.shx
+states.csv
+states.dbf
+states.shp
```
fridiculous, line 15: we may have to delete the files since they are added to gitignore, right?

libs/build_dod_dataset.py
```diff
@@ -105,6 +112,120 @@ def get_projections_df():
...
+
+    # get 16 and 32 days out from now
+    today = datetime.datetime.now()
+    sixteen_days = today + datetime.timedelta(days=16)
+    thirty_two_days = today + datetime.timedelta(days=32)
+
+    #save results in a list of lists, converted to df later
+    results = []
+
+    # get the state and fips so we can get the files (TODO: this is trash)
```
fridiculous, line 152: 😆

libs/build_dod_dataset.py
```diff
@@ -105,6 +112,120 @@ def get_projections_df():
...
+                    hosp_32_days, short_fall_32_days = get_hospitals_and_shortfalls(projection, thirty_two_days)
+                    hospitalizations = [int(row[9]) for row in projection]
+                    deaths = [int(row[11]) for row in projection]
+                    new_deaths = list(range(len(deaths)))
+                    new_deaths[0] = 0
+                    for i in range(1, len(deaths)):
+                        new_deaths[i] = int(deaths[i]) - int(deaths[i-1])
+                    mean_hospitalizations = math.floor(statistics.mean(hospitalizations))
+                    mean_deaths = math.floor(statistics.mean(new_deaths))
+                    peak_hospitalizations_date = find_peak(projection, 9)[1]
```
fridiculous, line 173: my main recommendation here would be to reference columns by name as opposed to by number

libs/build_dod_dataset.py
```diff
@@ -105,6 +112,120 @@ def get_projections_df():
     "Unassigned": NULL_VALUE
 }
 
+def get_state_and_counties():
+    # go to the JHUDataset to get this, for w/e reason
+    timeseries = JHUDataset.local().timeseries()
```
ghop02, line 117: If you use the `FIPSPopulation` dataset source, it has more "ground truth" county data - don't think all of the counties are in the JHU dataset.

Also, unrelated, over the week i've been thinking about ways to make the interface for datasets better - even though I wrote it, i've really come to hate the `.local().timeseries()`.  
Do you have an initial reaction to calling `TimeseriesDataset.load_jhu_dataset()` to get the data? Willingly taking any and all suggestions

libs/build_dod_dataset.py
```diff
@@ -105,6 +112,120 @@ def get_projections_df():
...
+
+    #save results in a list of lists, converted to df later
+    results = []
+
+    # get the state and fips so we can get the files (TODO: this is trash)
+    missing = 0
+    for state, counties in get_state_and_counties().items():
+        for county, fips in counties:
+            file_name = f"{state}.{fips}.{intervention_type}.json"
+            path = os.path.join(OUTPUT_DIR_COUNTIES, file_name)
```
ghop02, line 157: Zero need to change, this but figured I'd throw it out there because I absolutely have come to adore [pathlib](https://docs.python.org/3/library/pathlib.html) (apart from urls...) and now have become its evangelist. the pathlib version of this would be:
```
import pathlib

path = pathlib.Path(OUTPUT_DIR_COUNTIES) / filename
if path.exists():
    with path.open("r) as projections:
```

libs/build_dod_dataset.py
```diff
@@ -105,6 +112,120 @@ def get_projections_df():
...
+
+    counties_remapped = counties_decorated.rename(columns=state_col_remap)
+
+    new_cols = list(set(output_cols + list(state_col_remap.values())))
+    counties = pd.DataFrame(counties_remapped, columns=new_cols)
+    counties = counties.fillna(NULL_VALUE)
+    counties.index.name = 'OBJECTID'
+    counties['State/County FIPS Code'] = counties['State/County FIPS Code'].astype(str).str.rjust(5, '0')
+    # assert unique key test
+    assert counties['Combined Key'].value_counts().max() == 1
```
ghop02, line 226: Ah this is great!

libs/build_dod_dataset.py
ghop02: I guess this isn't in this change, but if you do:
```
data = pd.read_csv(file, dtype={"FIPS": str})
data = data['State/County FIPS Code'].str.zfill(5)
```
I think that would work!

libs/build_dod_dataset.py
```diff
@@ -6,14 +6,21 @@
...
 from urllib.parse import urlparse
+from collections import defaultdict
 
-from .build_params import OUTPUT_DIR
+from .enums import NO_INTERVENTION
+from .build_params import OUTPUT_DIR, OUTPUT_DIR_COUNTIES
 from .CovidDatasets import get_public_data_base_url
 from .us_state_abbrev import us_state_abbrev, us_fips
+from .datasets import JHUDataset
+from .datasets.dataset_utils import AggregationLevel
```
ghop02, line 20: if possible, can you use the entire package name here? the [style guide](https://google.github.io/styleguide/pyguide.html#23-packages) explains a bit more for why to use the full name. i like that it gives a bit more context without having to remember where this file is.

run2.py
goldblatt: ah yeah sorry forgot

libs/build_dod_dataset.py
goldblatt: cool these are json but made a constant for clarity

libs/build_dod_dataset.py
goldblatt: oh perfect i had been looking for this

libs/build_dod_dataset.py
goldblatt: cool cool was just following the file conventionbut i changed it

.gitignore
goldblatt: yes yes

libs/datasets/sources/fips_population.py
```diff
@@ -40,6 +40,7 @@ class Fields(object):
 
     def __init__(self, path):
         data = pd.read_csv(path, dtype={"fips": str})
+        data['fips'] = data.fips.str.zfill(5)
```
goldblatt, line 43: i'm pretty sure this is the only one that needed to be changed

libs/build_dod_dataset.py
```diff
@@ -101,14 +107,140 @@ def get_projections_df():
...
+    counties['State/County FIPS Code'] = counties['State/County FIPS Code'].astype(str).str.rjust(5, '0')
+    # assert unique key test
+    assert counties['Combined Key'].value_counts().max() == 1
+    return counties
+
 def get_usa_by_county_df():
     url = '{}/data/cases-jhu/csse_covid_19_daily_reports/{}.csv'.format(
         get_public_data_base_url(), latest.strftime("%m-%d-%Y"))
-    raw_df = pd.read_csv(url)
+    raw_df = pd.read_csv(url, dtype={"FIPS": str})
```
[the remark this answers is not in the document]
goldblatt: ya i was spot checking some counties and it does seem like some counties we have data for (like carson montana (fips code 30009) exists in the website, but doesn't show up in this exported data). I think we should address that in a different PR though?

fridiculous:
Overall, awesome work getting all this stuff working and up.

Only major Hygenic change is we should probably commit the fips.csv to https://github.com/covid-projections/covid-data-public instead

PR #98, opened 2020-04-03
Add latest summary

ghop02:
Quick script to add summary files for latest state and county data

### DIFF
--- a/run_data.py
+++ b/run_data.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+
+import pathlib
+import json
+import logging
+import click
+from libs import build_params
+from libs.datasets import JHUDataset
+from libs.datasets import FIPSPopulation
+from libs.datasets.dataset_utils import AggregationLevel
+
+WEB_DEPLOY_PATH = pathlib.Path("../covid-projections/public/data")
+
+
+_logger = logging.getLogger(__name__)
+
+
+@click.group()
+def main():
+    pass
+
+
+@main.command("latest")
+@click.option("--state", "-s")
+@click.option(
+    "--deploy",
+    is_flag=True,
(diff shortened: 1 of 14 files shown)


ghop02:
#99

run_data.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+import pathlib
+import json
+import logging
+import click
+from libs import build_params
+from libs.datasets import JHUDataset
+from libs.datasets.dataset_utils import AggregationLevel
+
+WEB_DEPLOY_PATH = pathlib.Path("../covid-projections/public/data")
```
fridiculous, line 15: would be cool if this could be set as opposed as part of the cli or via an env var

run_data.py
```diff
@@ -0,0 +1,74 @@
...
+    """Get latest case values from JHU dataset."""
+    output_dir = pathlib.Path(build_params.OUTPUT_DIR)
+    if deploy:
+        output_dir = WEB_DEPLOY_PATH
+
+    output_folder = output_dir / "case_summary"
+    output_folder.mkdir(exist_ok=True)
+    timeseries = JHUDataset.local().timeseries().get_subset(None, country="USA")
+
+    timeseries.data.cases = timeseries.data.cases.fillna(0)
```
fridiculous, line 42: if possible, it would be great that logic like formatting or transformation doens't live here

ghop02: Yeah definitely true.  

I've been punting on this issue for a bit - should we treat these missing values as zeros rather than NaNs across the board? If so, I can modify the timeseries class to fill those in as such

run_data.py
ghop02: yeah i think that's true, i can address that in a follow up pr if that's cool

PR #103, opened 2020-04-03
Adding in a max_bed_count column + simple tests on beds dataset

ghop02:
Moved the logic for choosing a bed count from querying functions to the initial data initialization.

Did this to:
 * make it easier to join on other datasets and not have to work to come up with the correct beds number.
 * Reduce touch points for the max calculation.

Also adds in the ability to quickly query just state or county level data.

### DIFF
--- a/libs/datasets/beds.py
+++ b/libs/datasets/beds.py
@@ -1,7 +1,5 @@
-from typing import List, Iterator
-import enum
+from typing import Optional
 import pandas as pd
-import datetime
 from libs.datasets import dataset_utils
 from libs.datasets.dataset_utils import AggregationLevel
 from libs.datasets import custom_aggregations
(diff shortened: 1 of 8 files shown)


libs/datasets/beds.py
```diff
@@ -1,56 +1,90 @@
...
         data = source.data
+
         to_common_fields = {value: key for key, value in source.BEDS_FIELD_MAP.items()}
         final_columns = to_common_fields.values()
         data = data.rename(columns=to_common_fields)[final_columns]
         data[cls.Fields.SOURCE] = source.SOURCE_NAME
         data[cls.Fields.GENERATED] = False
 
-        group = [cls.Fields.SOURCE, cls.Fields.AGGREGATE_LEVEL, cls.Fields.STATE, cls.Fields.GENERATED]
-        data = custom_aggregations.update_with_combined_new_york_counties(
```
gsoltis, line 45: The reason we don't need this aggregation any more is because we're only operating at the state level? If so, I think we can remove the import.

PR #104, opened 2020-04-04
Add git hash support to run_model.py

gsoltis:
Also output a version file when the model is run. Git hash support is in the form of the `--git-hash` commandline argument, as that seems to be how we are configuring these runs.

### DIFF
--- a/libs/datasets/dataset_utils.py
+++ b/libs/datasets/dataset_utils.py
@@ -1,9 +1,13 @@
 from contextlib import contextmanager
+from datetime import datetime
 import enum
 import git
+import json
 import logging
+import os
 import pathlib
 import pandas as pd
+import pytz
 from typing import Optional
 from libs import build_params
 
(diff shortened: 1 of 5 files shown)


libs/datasets/dataset_utils.py
```diff
@@ -13,6 +17,42 @@
 
 _logger = logging.getLogger(__name__)
 
+class DataVersion(object):
```
ghop02, line 20: Do you think you could move the data version related code out into it's own file? 

dataset_utils is already a bit overloaded and I think the data version code has enough functionality that it merits its own file!

run_model.py
```diff
@@ -26,21 +27,34 @@ def main():
...
+    help='''
+    | Git hash of the commit in covid-data-public to use.
+    | If provided, covid-data-public must have no pending changes.
+    | If omitted, the repository will be used as-is'''
+)
+def run_county(state=None, deploy=False, summary_only=False, git_hash=None):
     """Run county level model."""
-    min_date = datetime.datetime(2020, 3, 7)
-    max_date = datetime.datetime(2020, 7, 6)
+    with data_version(git_hash) as version:
```
ghop02, line 40: I don't love having to wrap all code that uses the version in a context manager - I think this would be a good chance for a decorator to enclose this context manager. 

I gave this approach a shot (#106 ) - let me know what you think!

libs/datasets/dataset_utils.py
gsoltis: Done.

run_model.py
gsoltis: Merged and very lightly updated.

run_model.py
```diff
@@ -26,7 +28,8 @@ def main():
     is_flag=True,
     help="Only runs the county summary if true.",
 )
-def run_county(state=None, deploy=False, summary_only=False):
+@data_version.with_git_version_click_option
+def run_county(version: data_version.DataVersion, state=None, deploy=False, summary_only=False):
```
ghop02, line 32: 💯 for type

run_model.py
```diff
@@ -61,6 +70,12 @@ def run_state(state=None, deploy=False):
     run.run_state_level_forecast(
         min_date, max_date, country="USA", state=state, output_dir=output_dir
     )
+    logging.info(f'Wrote output to {output_dir}')
+    # only write the version if we saved everything
+    if state is None :
+        version.write_file('states', output_dir)
```
ghop02, line 76: Maybe in the future we can write this in the json file itself

run_model.py
```diff
@@ -61,6 +70,12 @@ def run_state(state=None, deploy=False):
     run.run_state_level_forecast(
         min_date, max_date, country="USA", state=state, output_dir=output_dir
     )
+    logging.info(f'Wrote output to {output_dir}')
```
ghop02, line 73: Would prefer if we used a module level logger to give a bit more context to where the log is coming from.

gsoltis: Done.

run_model.py
gsoltis: I like this idea, but we would obviously need to verify that the consumers don't break.

PR #108, opened 2020-04-04
Tweaks for performance & readability

fridiculous:


libs/build_dod_dataset.py
```diff
@@ -16,7 +16,9 @@
 from libs.build_params import OUTPUT_DIR, OUTPUT_DIR_COUNTIES
 from libs.CovidDatasets import get_public_data_base_url
 from libs.us_state_abbrev import us_state_abbrev, us_fips
-from libs.datasets import FIPSPopulation
+from libs.datasets import FIPSPopulation, JHUDataset
```
goldblatt, line 19: don't need JHUDataset here

libs/build_dod_dataset.py
```diff
@@ -16,7 +16,9 @@
...
+from libs.datasets.dataset_utils import AggregationLevel
```
goldblatt, line 20: also don't need AggregationLevel

libs/build_dod_dataset.py
```diff
@@ -68,19 +103,18 @@ def get_projections_df():
...
+            df = read_json_as_df(path)
+            df['short_fall'] = df.apply(calc_short_fall, axis=1)
 
-                hosp_16_days, short_fall_16_days = get_hospitals_and_shortfalls(projection, sixteen_days)
-                hosp_32_days, short_fall_32_days = get_hospitals_and_shortfalls(projection, thirty_two_days)
+            hosp_16_days, short_fall_16_days = get_hospitals_and_shortfalls(df, sixteen_days)
+            hosp_32_days, short_fall_32_days = get_hospitals_and_shortfalls(df, thirty_two_days)
 
-                results.append([state, hosp_16_days, hosp_32_days, short_fall_16_days, short_fall_32_days])
+            results.append([state, hosp_16_days, short_fall_16_days, short_fall_16_days, short_fall_32_days])
```
goldblatt, line 117: should still be hosp_32_days here right

libs/build_dod_dataset.py
```diff
@@ -158,26 +186,28 @@ def get_county_projections():
...
+
+            peak_hospitalizations_date = df.iloc[df.all_hospitalized.idxmax()].date
+            peak_deaths_date = df.iloc[df.dead.idxmax()].date
+
+            results.append([state, fips, hosp_16_days, hosp_32_days, short_fall_16_days, short_fall_32_days,
+                    mean_hospitalizations, mean_deaths, peak_hospitalizations_date, peak_deaths_date])
         else:
             missing = missing + 1
-    print(f'Models missing for {missing} counties')
+    print(f'Models missing for {missing} states')
```
goldblatt, line 210: i think this should still say county

libs/build_dod_dataset.py
```diff
@@ -158,26 +186,28 @@ def get_county_projections():
...
+            df['short_fall'] = df.apply(calc_short_fall, axis=1)
+
+            hosp_16_days, short_fall_16_days = get_hospitals_and_shortfalls(df, sixteen_days)
+            hosp_32_days, short_fall_32_days = get_hospitals_and_shortfalls(df, thirty_two_days)
+
+            #hospitalizations = [int(row[9]) for row in projection]
+            #deaths = [int(row[11]) for row in projection]
+            df['new_deaths'] = df.dead - df.dead.shift(1)
+
+            mean_hospitalizations = df.all_hospitalized.mean().round(0)
```
goldblatt, line 199: average_per_day? idk it just seems kinda unhelpful

libs/build_dod_dataset.py
```diff
@@ -158,26 +186,28 @@ def get_county_projections():
...
+            # mean_hospitalizations = math.floor(statistics.mean(hospitalizations))
+            mean_deaths = df.new_deaths.mean()
```
goldblatt, line 201: idk this also seems odd because it's like flattening the data a level since the rate of change is interesting?

libs/build_dod_dataset.py
```diff
@@ -158,26 +186,28 @@ def get_county_projections():
...
+
+            peak_hospitalizations_date = df.iloc[df.all_hospitalized.idxmax()].date
+            peak_deaths_date = df.iloc[df.dead.idxmax()].date
```
goldblatt, line 204: should this be new_dead? or is dead the daily (instead of the cumulative)

libs/build_dod_dataset.py
```diff
@@ -43,16 +43,49 @@ def get_abbrev_df():
...
+    "l",
+    "population",
+    "m",
+    "n",
+]
+
+exclude_cols = ["a","b","c","d","e","f","g","i","j","k","l","m","n"]
+
+def read_json_as_df(path):
+    df = pd.DataFrame.from_records(simplejson.load(open(path,'r')),
```
ghop02, line 73: What's the reason for simplejson vs json here?

libs/build_dod_dataset.py
```diff
@@ -43,16 +43,49 @@ def get_abbrev_df():
...
+    columns=all_cols, exclude=exclude_cols)
 
-        if row_time >= days_out:
-            hospitalizations = int(row[ALL_HOSPITALIZED])
-            beds = int(row[BEDS])
-            short_fall = abs(beds - hospitalizations) if hospitalizations > beds else 0
-            return hospitalizations, short_fall
-    return 0, 0
+    df['date'] = pd.to_datetime(df.date)
```
ghop02, line 76: Probably more efficient to define this in the `from_records`:
```
parse_dates=['date'],
dtype={
    'all_hospitalized': int,
    'beds': int,
    'dead': int
}|

```

libs/build_dod_dataset.py
goldblatt: ah no i see it shouldn't

libs/build_dod_dataset.py
fridiculous: it's faster for w/e reason

libs/build_dod_dataset.py
```diff
@@ -241,6 +269,7 @@ def get_usa_by_county_df():
     url = '{}/data/cases-jhu/csse_covid_19_daily_reports/{}.csv'.format(
         get_public_data_base_url(), latest.strftime("%m-%d-%Y"))
     raw_df = pd.read_csv(url, dtype={"FIPS": str})
+    raw_df.to_csv('deleteme.csv', index=False)
```
goldblatt, line 272: ?

libs/build_dod_dataset.py
fridiculous: i tried this but it didn't work!

ghop02: 🙀

libs/build_dod_dataset.py
```diff
@@ -241,6 +269,7 @@ def get_usa_by_county_df():
     url = '{}/data/cases-jhu/csse_covid_19_daily_reports/{}.csv'.format(
         get_public_data_base_url(), latest.strftime("%m-%d-%Y"))
     raw_df = pd.read_csv(url, dtype={"FIPS": str})
```
goldblatt, line 271: hmm maybe we need to add the padding?

libs/build_dod_dataset.py
fridiculous: you were right, it should be new_deaths

libs/build_dod_dataset.py
fridiculous: it's what they asked for

libs/build_dod_dataset.py
```diff
@@ -158,26 +184,28 @@ def get_county_projections():
         path = os.path.join(OUTPUT_DIR_COUNTIES, file_name)
         # if the file exists in that directory then process
         if os.path.exists(path):
-            with open(path, "r") as projections:
```
goldblatt, line 161: i wonder if my `fips = fips_row['fips']` trims the leading zerro

ghop02: pretty unrelated but should try to save our fips codes in csvs surrounded by quotes so we don't always have to fill em in

libs/build_dod_dataset.py
ghop02: the jhu data does have a few fips (not most of them) without leading zeros i'm pretty sure

PR #111, opened 2020-04-04
Add output mapper

erccarls:
Map from pyseir output to the webui for counties. States not done yet.


pyseir/ensembles/ensemble_runner.py
```diff
@@ -141,8 +142,7 @@ def _generate_compartment_arrays(model_ensemble):
         value_stack: array[n_samples, time steps]
             Array with the stacked model output results.
         """
-        compartments = {key: [] for key in model_ensemble[0].results.keys() if key not in ('t_list')}
-
+        compartments = {key: [] for key in model_ensemble[0].results.keys() if key not in ('t_list', 'county_metadata')}
```
erccarls, line 145: This is just adding some metadata.

fridiculous:
hey @erccarls - given our discussion and the suggested structure here? https://github.com/covid-projections/covid-data-model/issues/145 can we close out this branch?

pyseir/cli.py
```diff
@@ -60,24 +61,42 @@ def _generate_state_reports(state=None):
...
-def _run_all(state=None):
+def _run_all(state=None, run_mode='default', generate_reports=True, output_interval_days=4, skip_download=False):
     exceptions = []
-    cache_all_data()
+
+    if not skip_download:
+        cache_all_data()
 
     if state:
         _impute_start_dates(state.title())
```
zhang-xinyu, line 84: What is difference between state.title and state.name.title? Should this be state.name.title or the state.name.title should be updated to this? This comment applies throughout this PR.

pyseir/cli.py
```diff
@@ -108,5 +129,15 @@ def generate_state_report(state):
 
 @entry_point.command()
 @click.option('--state', default='', help='State to generate files for. If no state is given, all states are computed.')
-def run_all(state=None):
-    _run_all(state)
+@click.option('--output-interval-days', default='', type=int, help='Number of days between outputs for the WebUI payload.')
```
zhang-xinyu, line 132: What is output-interval-days? Time forward for forecasting?

pyseir/ensembles/ensemble_runner.py
```diff
@@ -58,17 +73,102 @@ def __init__(self, fips, n_years=2, n_samples=250,
...
+
+    def init_run_mode(self):
+        """
+        Based on the run mode, generate suppression policies and ensemble
+        parameters.  This enables different model combinations and project
+        phases.
+        """
+        self.suppression_policies = dict()
+
+        min_date_data = datetime.datetime.today() - datetime.timedelta(days=30)
```
zhang-xinyu, line 102: Does this define the earliest date extract case data or predictions?

pyseir/ensembles/ensemble_runner.py
```diff
@@ -58,17 +73,102 @@ def __init__(self, fips, n_years=2, n_samples=250,
...
+        covid_data = timeseries.get_subset(AggregationLevel.COUNTY, after=min_date_data, country='USA', state=self.state_abbr)\
+                  .get_data(state=self.state_abbr, country='USA', fips=self.fips)
+
+        if self.run_mode is RunMode.CAN_BEFORE:
+            self.n_samples = 1
+
+            for scenario in ['no_intervention', 'flatten_the_curve', 'full_containment', 'social_distancing']:
+                R0 = 3.6
```
zhang-xinyu, line 110: why is this hard coded? I thought covidactnow model is inferring R0.

pyseir/ensembles/ensemble_runner.py
```diff
@@ -94,27 +194,27 @@ def run_ensemble(self):
...
                 N_samples=self.n_samples,
                 t_list=self.t_list,
-                suppression_policy=generate_empirical_distancing_policy(
-                    t_list=self.t_list,
-                    fips=self.fips,
-                    future_suppression=suppression_policy
-                )).sample_seir_parameters()
-
+                suppression_policy=suppression_policy)
+            parameter_ensemble = parameter_sampler.sample_seir_parameters(override_params=self.override_params)
```
zhang-xinyu, line 205: what is override_params?

pyseir/models/suppression_policies.py
```diff
@@ -69,6 +69,92 @@ def generate_triggered_suppression_model(t_list, lockdown_days, open_days, reduc
...
+        today = datetime.utcnow()
+
+        if scenario == 'no_intervention':
+            rho.append(1)
+
+        elif scenario == 'flatten_the_curve':
+            if actual_date <= today:
+                rho.append(1)
+            elif (actual_date - today).days <= 30:
+                rho.append(1.3 / R0)
```
zhang-xinyu, line 121: Are rhos values selected by covidactnow? I am confused how this and the date together translates R_eff.

pyseir/cli.py
```diff
@@ -60,24 +61,42 @@ def _generate_state_reports(state=None):
         report.generate_report()
     else:
         for state in us.states.STATES:
-            _generate_state_reports(state.title())
+            _generate_state_reports(state.name.title())
```
zhang-xinyu, line 64: Should this be state.name.title or state.title? Seems you are using state.title somewhere else.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def generate_state(self):
+        """
+        Generate for each county in a state, the output for the webUI.
+        """
+        df = load_data.load_county_metadata()
+        all_fips = df[df['state'].str.lower() == self.state.lower()].fips
+        p = Pool()
+        p.map(self.map_fips, all_fips)
+        p.close()
```
ghop02, line 108: Do you not have to join? my multiprocessing knowledge is so seasonal

pyseir/ensembles/ensemble_runner.py
```diff
@@ -2,19 +2,31 @@
...
 
+from libs.datasets import JHUDataset
+from libs.datasets import FIPSPopulation
+from libs.datasets.dataset_utils import AggregationLevel
+timeseries = JHUDataset.local().timeseries()
+
+
+class RunMode(Enum):
+    DEFAULT = 'default'  # 4 basic scenarios.
+    CAN_BEFORE = 'can-before' # 4 basic scenarios.
```
ghop02, line 28: What's the difference here?

pyseir/ensembles/ensemble_runner.py
```diff
@@ -58,17 +73,102 @@ def __init__(self, fips, n_years=2, n_samples=250,
...
+        self.suppression_policies = dict()
+
+        min_date_data = datetime.datetime.today() - datetime.timedelta(days=30)
+        covid_data = timeseries.get_subset(AggregationLevel.COUNTY, after=min_date_data, country='USA', state=self.state_abbr)\
+                  .get_data(state=self.state_abbr, country='USA', fips=self.fips)
+
+        if self.run_mode is RunMode.CAN_BEFORE:
+            self.n_samples = 1
+
+            for scenario in ['no_intervention', 'flatten_the_curve', 'full_containment', 'social_distancing']:
```
ghop02, line 109: We have these defined now in `libs.enums.Intervention` I think

pyseir/ensembles/ensemble_runner.py
```diff
@@ -94,27 +194,27 @@ def run_ensemble(self):
         Run an ensemble of models for each suppression policy nad generate the
         output report / results dataset.
         """
+        for suppression_policy_name, suppression_policy in self.suppression_policies.items():
+            logging.info(f'Running simulation ensemble for {self.county_metadata["county"]}, {self.county_metadata["state"]} {suppression_policy_name}')
```
ghop02, line 198: Can you use a module level logger? 
```
import logging
```

pyseir/load_data.py
```diff
@@ -333,7 +337,7 @@ def cache_all_data():
     cache_county_case_data()
     cache_hospital_beds()
     cache_mobility_data()
-    cache_public_implementations_data()
+    # cache_public_implementations_data()
```
ghop02, line 340: ?

pyseir/models/suppression_policies.py
```diff
@@ -69,6 +69,92 @@ def generate_triggered_suppression_model(t_list, lockdown_days, open_days, reduc
...
+    Returns
+    -------
+    suppression_model: callable
+        suppression_model(t) returns the current suppression model at time t.
+    """
+    # Rho is the multiplier on the contact rate. i.e. 1 = no change, 0 = no transmission
+    rho = []
+    for t in t_list:
+        actual_date = t0 + timedelta(days=t)
+        today = datetime.utcnow()
```
ghop02, line 112: thank you for utc 👍

pyseir/parameters/parameter_ensemble_generator.py
```diff
@@ -1,6 +1,13 @@
 import numpy as np
 import pandas as pd
 from pyseir import load_data
+from libs.datasets import FIPSPopulation
+from libs.datasets import DHBeds
+import us
+
+
+beds_data = DHBeds.local().beds()
+population_data = FIPSPopulation.local().population()
```
ghop02, line 10: Can you move this to the __init__ function? .I'd prefer not to have any side effects files on import

pyseir/ensembles/ensemble_runner.py
ghop02: no need to change though, we can work these in later

pyseir/cli.py
erccarls: Clarified. One was an object the other is a string.

pyseir/cli.py
erccarls: You can read the help string there. It specifies the output data interval that gets consumed by the Web UI.

pyseir/ensembles/ensemble_runner.py
erccarls: Outdated. No longer needed, but it is the earliest data extracted.

pyseir/ensembles/ensemble_runner.py
erccarls: This is the "inferred" value, but it was hardcoded in theirs as well.  This week we will switch ours to infer using the model, not the exponential growth phase.

pyseir/ensembles/ensemble_runner.py
erccarls: See the docstring for the consuming method. It overrides the parameter sampler values.

erccarls: This is temporary anyway since you are going to generate configs.

pyseir/models/suppression_policies.py
erccarls: Our suppression policy is usually a number between 0 - 1. They hardcoded R_eff rather than XX percent reduction. This fixes that mapping to reproduce an R_eff after intervention.

pyseir/cli.py
erccarls: fixed.

pyseir/deployment/webui_data_adaptor_v1.py
erccarls: Close does a join and then shuts down the pool.

pyseir/ensembles/ensemble_runner.py
erccarls: Oops. Fixed the docstring.   This will be moved to config files soon, but currently the run modes are different parameter sets.

pyseir/load_data.py
erccarls: mistake. Fixed.

pyseir/ensembles/ensemble_runner.py
erccarls: Yes, but if you don't mind let me do this tomorrow across all modules.  Do you have a specific preference for why? We currently output filename and line no in the root logger.

pyseir/parameters/parameter_ensemble_generator.py
erccarls: I agree except that it is detrimental to performance since these loads are heavyweight... Maybe I can use globals which is almost as ugly, but fewer side effects.

ghop02: yeah, fine with that solution for now, but we should try to figure out in the run structure how to just define this once and pass it in where needed.

pyseir/ensembles/ensemble_runner.py
ghop02: Mostly because it gives the user of the package more granular log control.  I'll sometimes set logging levels at different levels depending on the package:
for example:
```
logging.getLogger("botocore").setLevel(logging.INFO)
logging.getLogger("boto3.resources").setLevel(logging.INFO)
logging.getLogger("urllib3.connectionpool").setLevel(logging.INFO)
```
if its on the root logger you don't get that control

pyseir/parameters/parameter_ensemble_generator.py
erccarls: Can we add that caching to the data source abstraction itself? Otherwise it seems the code is needlessly complex given that we want to load these and other metadata in several places in a given run.

PR #112, opened 2020-04-04
Fridiculous/crazy fips

fridiculous:


libs/datasets/sources/fips_population.py
```diff
@@ -40,6 +40,7 @@ class Fields(object):
 
     def __init__(self, path):
         data = pd.read_csv(path, dtype={"fips": str})
+        data['fips'] = data.fips.str.zfill(5)
```
ghop02, line 43: Can you move this into the `standardize_data` function? would prefer to have all of that logic there

ghop02: Also, i'm curious which fips didn't have padded zeros? 
when I run, it shows they're all length 5 (or na for states):
```
pop = FIPSPopulation.local().population().data.fips
pop.groupby(pop.apply(lambda x: x if pd.isna(x) else len(x))).size()
```

fridiculous: this whole needs to be re-done

ghop02: I would rather have a check that explicitly breaks if the fips isn't len 5 I think. but just my 0.02

fridiculous:
> ohh wait i think this should be balcklisted!

is that because it's a dupe key?

libs/build_dod_dataset.py
```diff
@@ -347,6 +352,7 @@ def get_usa_by_states_df():
     states_final = states_final.fillna(NULL_VALUE)
     states_final['Combined Key'] = states_final['Province/State']
     states_final['State/County FIPS Code'] = states_final['Province/State'].map(us_fips)
+    states_final['State/County FIPS Code'] = states_final['State/County FIPS Code'].astype(str).str.zfill(5)
```
vincentwoo, line 355: is this necessary if we have performed the padding earlier in this function at raw_df?

fridiculous: yea, it's probably overkill, then again, most of this code is.

PR #115, opened 2020-04-05
Adding error logging to catch errors from parallel calls

ghop02:
Doesn't yet address the actual error in https://github.com/covid-projections/covid-projections/issues/419, but this adds error logging that would've helped catch it sooner.

### DIFF
--- a/run.py
+++ b/run.py
@@ -330,7 +330,6 @@ def forecast_each_state(
         _logger.warning(f"Missing population for {state}")
         return
 
-    all_website_data = []
     for i, intervention in enumerate(get_interventions()):
         _logger.info(f"Running intervention {i} for {state}")
         results = model_state(cases, beds, population, intervention)
@@ -338,10 +337,7 @@ def forecast_each_state(
             results, cases, population, min_date, max_date, interval=4
         )
 
-        all_website_data.append(website_data)
         write_results(website_data, output_dir, f"{state}.{i}.json")
-
-    return all_website_data
 
 
 @catch_and_return_errors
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -320,14 +373,18 @@ def forecast_each_county(
...
+    for i, intervention in enumerate(interventions):
         results = model_state(cases, beds, population, intervention)
         website_data = prepare_data_for_website(
             results, cases, population, min_date, max_date, interval=4
         )
-
+        all_website_data.append(website_data)
         write_results(website_data, output_dir, f"{state}.{fips}.{i}.json")
 
+    return all_website_data
```
mikelehen, line 386: You changed these functions to return this `all_website_data` thing, but I don't see where it is ultimately used.  I'm probably just confused.

run.py
```diff
@@ -397,19 +458,11 @@ def run_state_level_forecast(
...
-if __name__ == "__main__":
-    logging.basicConfig(level=logging.INFO)
-    with public_data_hash(os.getenv('COVID_DATA_PUBLIC_HASH', None)) as git_hash:
-        # @TODO: Record git hash in output data for reproducibility
-        # @TODO: Remove interventions override once support is in the Harvard model.
-        min_date = datetime.datetime(2020, 3, 7)
-        max_date = datetime.datetime(2020, 7, 6)
-        # build_county_summary()
-        run_county_level_forecast(min_date, max_date)
-        run_state_level_forecast(min_date, max_date)
```
mikelehen, line 415: Er, why is all of this deleted?

ghop02: ah, i'll leave it for now actually.  Most of this logic for running the models now lives in `./run_model.py`

run.py
ghop02: ah yeah, I'll also remove this until we actually need the output result.  It felt appropriate to have the function return the results of running the model, but don't yet need it for anything.

PR #127, opened 2020-04-06
States Fields + Validations + Interventions

goldblatt:
- Updated the states fields to include the updated fields we use for counties
- loop through the interventions and upload them all
- changed where the build_dod_dataset output lands for file structure joy
- added a validations folder that doesn't let you upload if things are bad (ie it's not the right size-ish) or it's missing states/expected headers


libs/constants.py
```diff
@@ -1,11 +1,24 @@
 NO_INTERVENTION = 0
+FLATTEN = 1
+FULL_CONTAINMENT = 2
+SOCIAL_DISTANCING = 3
 
+INTERVENTIONS = [
```
ghop02, line 6: I think this is a good use case for the [enum](https://docs.python.org/3/library/enum.html#creating-an-enum). 

```
import enum
class Intervention(enum.Enum):
    NO_INTERVENTION = 0
    FLATTEN = 1
    FULL_CONTAINMENT = 2
    SOCIAL_DISTANCING = 3
```

Then you can access it like:
```
Intervention.SOCIAL_DISTANCING.value  # 1
Intervention.SOCIAL_DISTANCING.name  # "SOCIAL_DISTANCING"

```

libs/validate_results.py
```diff
@@ -0,0 +1,112 @@
+import csv 
+import sys
+
+from libs.us_state_abbrev import us_state_abbrev
+
+"""
+A set of functions to validate the datasets we prepare
```
ghop02, line 7: Thoughts on renaming this file `libs/validate_dod_results.py` for now?  Better to just be more explicit about that for the time being, since we don't quite validate all of the datasets just yet here.

Also, we have some of this logic in `libs/datasets/dataset_utils.py` (definitely not all of it!).  do you think we could combine the validation logic functions from `libs/datasets/dataset_utils.py` with these? maybe `libs/datasets/validation.py`?

libs/validate_results.py
```diff
@@ -0,0 +1,112 @@
...
+EXPECTED_MISSING_STATES_FROM_COUNTES = set([
+    'District of Columbia'
+])
+
+class DataExportException(Exception): 
+    def __init__(self, key, message):
+        self.key = key
+        self.message = message
+
+def __raise_error_if_incorrect_headers(key, headers_list, df):
```
ghop02, line 51: in general in python single underscores are the standard here.  Double underscores do some weird name mangling (though that might only be in classes) so they are usually shied away from.

libs/validate_results.py
```diff
@@ -0,0 +1,112 @@
...
+
+def validate_counties_df(key, counties_df):
+    # assert the headers are what we expect
+    __raise_error_if_incorrect_headers(key, OUTPUT_HEADERS_COUNTIES, counties_df)
+
+    # assert data from each of the states
+    __raise_error_if_not_data_from_all_states(key, counties_df, EXPECTED_MISSING_STATES.union(EXPECTED_MISSING_STATES_FROM_COUNTES))
+
+    # assert no duplicated counties
+    if len(counties_df['State/County FIPS Code'].unique()) != len(counties_df['State/County FIPS Code']):
```
ghop02, line 85: just for reference:
https://github.com/covid-projections/covid-data-model/blob/master/libs/datasets/dataset_utils.py#L93
this does a similar thing here

deploy_dod_dataset.py
```diff
@@ -49,39 +56,55 @@ def persist(self):
...
-    DatasetDeployer(key='states.dbf', body=states_dbf.getvalue()).persist()
-
-    counties_shp = BytesIO()
-    counties_shx = BytesIO()
-    counties_dbf = BytesIO()
-    get_usa_county_shapefile(counties_shp, counties_shx, counties_dbf)
-    DatasetDeployer(key='counties.shp', body=counties_shp.getvalue()).persist()
-    DatasetDeployer(key='counties.shx', body=counties_shx.getvalue()).persist()
-    DatasetDeployer(key='counties.dbf', body=counties_dbf.getvalue()).persist()
+def deploy(should_run_validation):
```
ghop02, line 68: I think this is a good case to use a keyword argument.  Generally with booleans I prefer it, since:
`deploy(True)` is a bit more confusing just skimming through than `deploy(should_run_validation=True)`

deploy_dod_dataset.py
```diff
@@ -1,10 +1,17 @@
...
 import os
+import logging
 
+from libs.build_params import OUTPUT_DIR_DOD
+from libs.constants import INTERVENTIONS
+from libs.validate_results import validate_states_df, validate_counties_df, validate_states_shapefile, validate_counties_shapefile
 from libs.build_dod_dataset import get_usa_by_county_df, get_usa_by_states_df, get_usa_county_shapefile, get_usa_state_shapefile
 from libs.build_dod_dataset import get_usa_by_county_with_projection_df
 
+logger = logging.getLogger()
```
ghop02, line 12: `logger = logging.getLogger(__name__)` is preferred since it'll prepend the logs with the module name so you know where it came from

deploy_dod_dataset.py
```diff
@@ -1,10 +1,17 @@
 from io import BytesIO
 import boto3
 import os
+import logging
 
+from libs.build_params import OUTPUT_DIR_DOD
+from libs.constants import INTERVENTIONS
+from libs.validate_results import validate_states_df, validate_counties_df, validate_states_shapefile, validate_counties_shapefile
```
ghop02, line 8: this is a nit fersure, but I think it's better to make this `from libs import validate_results` - that way in the code, it's clear where calling those functions comes from.
https://google.github.io/styleguide/pyguide.html#22-imports has a good piece about that.

libs/build_dod_dataset.py
```diff
@@ -84,19 +100,41 @@ def calc_short_fall(x):
...
+    """
+    Given a file path, return the calculations we perform for that file. 
+    Note in the future maybe return a data type to keep type clarity
+    """
     # get 16 and 32 days out from now
     today = datetime.datetime.now()
     sixteen_days = today + datetime.timedelta(days=16)
     thirty_two_days = today + datetime.timedelta(days=32)
 
+    if os.path.exists(file_path):
```
ghop02, line 116: Maybe a good opportunity to return early here!

```
if not os.path.exists(file_path):
    return None

# This code then isn't indented!
```

libs/constants.py
goldblatt: ya sorry was being lazy

libs/validate_results.py
goldblatt: cool i think the duplication is fine here

libs/validate_results.py
```diff
@@ -0,0 +1,112 @@
...
+    states_in_df = set(df['Province/State'].unique())
+
+    missing_states_in_df = states - states_in_df - expected_missing
+    extra_states_in_df = states_in_df - states
+    if missing_states_in_df or extra_states_in_df: 
+        raise DataExportException(key, f"Missing Data from states: {missing_states_in_df}. Have extra states in df {extra_states_in_df}")
+
+def validate_states_df(key, states_df):
+    # assert the headers are what we expect
+    _raise_error_if_incorrect_headers(key, OUTPUT_HEADERS_STATES, states_df)
```
fridiculous, line 68: so cool!

libs/build_dod_dataset.py
```diff
@@ -139,7 +176,7 @@ def get_projections_df():
     # "People Tested",
 ]
 
-county_output_cols = ["Province/State",
+our_output_cols = ["Province/State",
```
fridiculous, line 179: nit: what is our?  i.e.
```suggestion
api_output_cols = ["Province/State",
```

fridiculous:
Overall solid work.  Would love to get that validation in there.

Issue #125, opened 2020-04-06
Create PR template to remind folks to document schema changes.


mikelehen:
To make sure API schema documentation is kept up-to-date and we audit API changes we want to have a PR template that reminds folks to document any changes to schema (new endpoints, json fields, etc.).

PR #128, opened 2020-04-06
Add initial PR Template

fridiculous:
This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/125

### Please Check
- [ ] Are you using data source loaders instead of reading data raw?
- [ ] Are tests passing?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,14 +1,6 @@
-This PR addresses
-https://github.com/covid-projections/covid-data-model/issues/< >
+This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/< >
 
 
-Please Check
-- [ ] Are you using data source loaders instead of reading data raw?
-- [ ] Is `api/README.md` up to date with the latest *schema*?
+### Please Check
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+This PR addresses
+https://github.com/covid-projections/covid-data-model/issues/< >
+
+
+Please Check
+- [ ] Are you using data source loaders instead of reading data raw?
```
mikelehen, line 6: Will everybody working in this codebase know exactly what this means?  (FWIW- I do not.)

Perhaps it could be expanded to something like:
> If you are adding a new data source, did you add a loader for it in `some/repo/path/` rather than reading it raw?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+- [ ] Is `api/README.md` up to date with the latest *schema*?
+- [ ] Are tests passing?
+
+#TODO
+# - [ ] Are you using data models for commonly used cases?
+# - [ ] Is the resultant output's format valid?
+# - [ ] Is the resultant output's data integral?
+# - [ ] Are you using the api models to persist data?
```
mikelehen, line 14: I've never messed with templates before.  Will these TODO's end up showing up in people's PR descriptions? Is that intentional?

Issue #123, opened 2020-04-06
Create run.sh script that generates all published data artifacts.


mikelehen:
The goal is to write a single script that runs "everything" we need to run in order to produce all artifacts that we want to publish on data.covidactnow.org.  This includes our website models and our DoD artifacts.  Per @fridiculous we prefer this to be a bash script to keep it simple and discourage any kind of data processing, etc. in it.  The resulting files in the output directory should match our API schema (which will be more formally defined as part of https://github.com/covid-projections/covid-data-model/issues/124).

Syntax should be something like:
```
run.sh <location of covid-data-public repo> <output location>
```

As we clean up our internal plumbing, what the script does may change but for now we need to probably:
* run run.py
* run deploy_dod_dataset.py pointed at output directory.
* Generate a version.json file that contains the current date/time, the git hash of covid-data-public, and the git hash of covid-data-model.

Context: https://docs.google.com/document/d/1VhQqMxtVYHQMIj9zFeb1DUQZ1aq1k4xIc_bYoTIqUb4

mikelehen:
This is mostly done (we now have deploy_website.sh and deploy_dod.sh that are very similar to what we want for run.sh).  I'll grab this and #126 tomorrow morning unless somebody beats me to it.

PR #129, opened 2020-04-06
Fix deploy subdirectory

ghop02:
Changing the deploy directory wasn't quite working properly, so this standardizes a bit of that.

Also, added a `deploy.sh` file to illustrate how to deploy the data to the website.  It doesn't quite cover #123 , but I think it's a start and at least documents the current process.

I named it `deploy.sh` rather than `run.sh` because it's pretty specific to what you need to do to actually deploy to the website, rather than just run the models for testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ See [covid-data-public](https://github.com/covid-projections/covid-data-public) 
 ## Running
 
 ### Run website data deploy
+
+This will run all model and data needed for the webiste, outputting to ``../covid-projections/public/data``.
 ```bash
 ./deploy.sh
 ```
--- a/README.md
+++ b/README.md
@@ -26,9 +26,9 @@ See [covid-data-public](https://github.com/covid-projections/covid-data-public) 
 
 ## Running
 
-### Create JSON for UI
+### Run website data deploy
 ```bash
-python run.py
+./deploy.sh
 ```
 
 ### Create JSON for API
--- a/libs/build_params.py
+++ b/libs/build_params.py
@@ -32,6 +32,7 @@ def get_interventions(start_date=datetime.now().date()):
 
 
 OUTPUT_DIR = 'results/test'
+OUTPUT_DIR_COUNTIES = 'results/county'
 
 # Relative directory to where all output data is deployed to for the website
 WEB_DEPLOY_PATH = pathlib.Path("../covid-projections/public/data")
--- a/libs/build_params.py
+++ b/libs/build_params.py
@@ -3,6 +3,7 @@ Constants to use for a build. In a separate file to avoid
 auto-importing a dataset when we don't necessarily need to.
 '''
 
+import pathlib
 from datetime import datetime, timedelta, date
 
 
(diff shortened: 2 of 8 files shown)


libs/build_params.py
```diff
@@ -29,11 +30,27 @@ def get_interventions(start_date=datetime.now().date()):
         },
     ]
 
+
 OUTPUT_DIR = 'results/test'
-OUTPUT_DIR_COUNTIES = 'results/county'
+COUNTY_SUBDIR = 'county'
+COUNTY_SUMMARY_SUBDIR = 'county_summaries'
+# Output directory for latest case data by state
+STATE_SUMMARY_SUBDIR = "case_summary"
```
mikelehen, line 38: Rather than hardcoding all of this in python-land, can we make the python scripts accept an arbitrary output directory?  i.e.

run_model.py --output ../covid-projections/public/data/

FWIW- We are probably going to tweak this directory structure as part of cleaning up our "API"...

fridiculous: +1 this

ghop02: yeah I like this too.

README.md
```diff
@@ -26,9 +26,9 @@ See [covid-data-public](https://github.com/covid-projections/covid-data-public)
 
 ## Running
 
-### Create JSON for UI
+### Run website data deploy
```
mikelehen, line 29: Can you add a note explicitly saying that this will run all models and write them to appropriate locations in ../covid-projections/public/data/ ?

run_data.py
```diff
@@ -25,27 +22,30 @@ def main():
...
     state_summaries = dataset_export.latest_case_summaries_by_state(timeseries)
 
     for state, state_summary in state_summaries:
-        output_file = output_folder / f"{state}.summary.json"
+        output_file = output / f"{state}.summary.json"
         with output_file.open("w") as f:
             _logger.info(f"Writing latest data for {state}")
             json.dump(state_summary, f)
 
+    if not state:
```
gsoltis, line 44: I don't think we want this condition here. `state` will always be defined, and I think we always want to write the version file since there doesn't appear to be an option to do a partial run.

run_model.py
```diff
@@ -20,63 +17,76 @@ def main():
...
     min_date = datetime.datetime(2020, 3, 7)
     max_date = datetime.datetime(2020, 7, 6)
 
-    output_dir = build_params.OUTPUT_DIR
-    if deploy:
-        output_dir = WEB_DEPLOY_PATH
+    run.run_county_level_forecast(
+        min_date, max_date, output, country="USA", state=state
+    )
+    if not state:
```
gsoltis, line 40: Missing `summary_only` check. Is there a reason we want to remove it?

libs/build_params.py
```diff
@@ -29,11 +30,16 @@ def get_interventions(start_date=datetime.now().date()):
         },
     ]
 
+
 OUTPUT_DIR = 'results/test'
 OUTPUT_DIR_COUNTIES = 'results/county'
 
+# Relative directory to where all output data is deployed to for the website
+WEB_DEPLOY_PATH = pathlib.Path("../covid-projections/public/data")
```
mikelehen, line 38: Should be unused now?

libs/build_params.py
```diff
@@ -3,6 +3,7 @@
 auto-importing a dataset when we don't necessarily need to.
 '''
 
+import pathlib
```
mikelehen, line 6: probably unused now?

run_data.py
ghop02: oh good call

run_model.py
ghop02: Oh, i didn't properly clean that up.  Basically the summary is moved to a different command to make output directories more configurable.

run.py
```diff
@@ -457,15 +463,3 @@ def run_state_level_forecast(
...
-if __name__ == "__main__":
-    logging.basicConfig(level=logging.INFO)
-    with public_data_hash(os.getenv('COVID_DATA_PUBLIC_HASH', None)) as git_hash:
-        # @TODO: Record git hash in output data for reproducibility
-        # @TODO: Remove interventions override once support is in the Harvard model.
-        min_date = datetime.datetime(2020, 3, 7)
-        max_date = datetime.datetime(2020, 7, 6)
-        # build_county_summary()
-        run_county_level_forecast(min_date, max_date)
-        run_state_level_forecast(min_date, max_date)
```
mikelehen, line 471: So you deleted this (again).  I don't know if that is okay.  Is this going to break other people's workflow?  And why does run.py exist if it doesn't have a main entry point?  

If run.py is deprecated / whatever, maybe you should make it print a message saying, "run.py is deprecated.  Run this instead: ..." or something?

ghop02: Now that we have a consistent way of running the deploys (./deploy.sh) it's easier to point people away from this, so i feel better about it now. 

I'm down with the deprecation method print message though.

run.py always did a lot of things, in the plan we have, those different tasks have new homes.  If it doesn't exist, I can make a ticket to split run.py up.

ghop02: Warning added!

PR #141, opened 2020-04-07
Add dod artifacts to deploy sh

vincentwoo:
Helps address #123, may conflict with @goldblatt PR.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -2,10 +2,11 @@
 
 PUBLIC_DATA_PATH=../covid-projections/public/data
 
-
 # Run State and County level models
 ./run_model.py state -o ${PUBLIC_DATA_PATH}
 ./run_model.py county -o ${PUBLIC_DATA_PATH}/county
 ./run_model.py county-summary -o ${PUBLIC_DATA_PATH}/county_summaries
 # Generate the latest state case summary data.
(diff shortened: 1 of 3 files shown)


deploy.sh
```diff
@@ -2,10 +2,11 @@
...
 
-
 # Run State and County level models
 ./run_model.py state -o ${PUBLIC_DATA_PATH}
 ./run_model.py county -o ${PUBLIC_DATA_PATH}/county
 ./run_model.py county-summary -o ${PUBLIC_DATA_PATH}/county_summaries
 # Generate the latest state case summary data.
 ./run_data.py latest -o ${PUBLIC_DATA_PATH}/case_summary
+
+./deploy_dod_dataset.py -i ${PUBLIC_DATA_PATH} -o ${PUBLIC_DATA_PATH}
```
ghop02, line 12: can you output this to a folder in `covid-data-model` instead of the public data path for the website?

mikelehen, line 12: This is going to copy files into the website repo which we probably don't want.  Can you change `${PUBLIC_DATA_PATH}` to `.` for now?

This will get reworked once we turn this into the run.sh script that's not hardcoded to deploy to the website. :)

PR #147, opened 2020-04-07
first pass at reorging the files

goldblatt:
Hmm something is a lil wrong with the output, but this runs so that's something

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/datasets/model_output_schema.py
+++ b/libs/datasets/model_output_schema.py
@@ -1,0 +1,25 @@
+MODEL_OUTPUT_SCHEMA = [
+    "day_num",
+    # ^ for index column
+    "date",
+    "a",
+    "b",
+    "c",
+    "d",
+    "e",
+    "f",
+    "g",
+    "all_hospitalized",
+    "all_infected",
+    "dead",
+    "beds",
+    "i",
+    "j",
+    "k",
+    "l",
+    "population",
+    "m",
+    "n",
+]
+
+MODEL_OUTPUT_SCHEMA_EXCLUDED_COLUMNS = ["a","b","c","d","e","f","g","i","j","k","l","m","n"]
--- a/libs/datasets/projections_schema.py
+++ b/libs/datasets/projections_schema.py
@@ -1,0 +1,31 @@
+OUTPUT_COLUMN_REMAP = {
+    'state_x': 'Province/State',
+    'intervention': 'State Intervention',
+    '16-day_Hospitalization_Prediction': '16d-HSPTLZD',
+    '32-day_Hospitalization_Prediction': '32d-HSPTLZD',
+    '16-day_Beds_Shortfall': '16d-LACKBEDS',
+    '32-day_Beds_Shortfall': '32d-LACKBEDS',
+    "Mean Hospitalizations": 'MEAN-HOSP',
+    "Mean Deaths": 'MEAN-DEATHS',
+    "Peak Hospitalizations On": 'PEAK-HOSP',
+    "Mean Deaths On": 'PEAK-DEATHS',
+    "Deaths": "Current Deaths",
+    "Confirmed": "Current Confirmed",
+    "Recovered": "Current Recovered",
+    "Active": "Current Active",
+}
+
+
+CALCULATED_PROJECTION_HEADERS_SHARED = [
+    '16-day_Hospitalization_Prediction',
+    '32-day_Hospitalization_Prediction',
+    '16-day_Beds_Shortfall',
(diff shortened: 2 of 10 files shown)


libs/datasets/model_output_schema.py
```diff
@@ -0,0 +1,25 @@
+MODEL_OUTPUT_SCHEMA = [
```
ghop02, line 1: It would be great to have this closer to the model that this schema applies to.  or maybe just a comment around which model this actually is.  many different files with the name model in it

libs/datasets/model_output_schema.py
```diff
@@ -0,0 +1,25 @@
...
+    "i",
+    "j",
+    "k",
+    "l",
+    "population",
+    "m",
+    "n",
+]
+
+MODEL_OUTPUT_SCHEMA_EXCLUDED_COLUMNS = ["a","b","c","d","e","f","g","i","j","k","l","m","n"]
```
ghop02, line 25: What are these excluded from?

libs/datasets/projections_schema.py
```diff
@@ -0,0 +1,31 @@
+OUTPUT_COLUMN_REMAP = {
```
ghop02, line 1: What's the input/output here? like what datasets are these mapping between?

libs/functions/calculate_projections.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    Given a file path, return the calculations we perform for that file. 
+    Note in the future maybe return a data type to keep type clarity
+    """
+    # get 16 and 32 days out from now
+    today = datetime.datetime.now()
+    sixteen_days = today + datetime.timedelta(days=16)
+    thirty_two_days = today + datetime.timedelta(days=32)
+
+    if os.path.exists(file_path):
```
ghop02, line 40: This path should exist right?

libs/functions/calculate_projections.py
```diff
@@ -0,0 +1,99 @@
...
+
+def _get_hospitals_and_shortfalls(df, date_out):
+    first_record_after_date = df[(df.date > date_out)].iloc[0]
+    hospital_short_fall_columns = ['all_hospitalized', 'short_fall']
+    return tuple(first_record_after_date[hospital_short_fall_columns].values)
+
+def _read_json_as_df(path):
+    # TODO: read this from a dataset class
+    df = pd.DataFrame.from_records(simplejson.load(open(path,'r')),
+    columns=MODEL_OUTPUT_SCHEMA, exclude=MODEL_OUTPUT_SCHEMA_EXCLUDED_COLUMNS)
```
ghop02, line 22: Just thinking here - I think this could be it's own `DataSource` that just has columns defined (like `JHUDataset`)

(definitely not for this pr lol)

libs/functions/calculate_projections.py
```diff
@@ -0,0 +1,99 @@
...
+        path = os.path.join(input_dir, 'state', file_name)
+
+        # if the file exists in that directory then process
+        projection_data = _calculate_projection_data(path)
+        if projection_data:
+            results.append([state] + projection_data)
+    return pd.DataFrame(results, columns=CALCULATED_PROJECTION_HEADERS_STATES)
+    
+def get_county_projections_df(input_dir, intervention_type):
+    # for each state in our data look at the results we generated via run.py
```
ghop02, line 76: Can you make these all docstrings while we're moving?

libs/functions/calculate_projections.py
goldblatt: nah i think we could have fips that don't have data for right?

libs/functions/calculate_projections.py
goldblatt: ya i think ideally we would use datasource for this that have the headers named and defined in the class so we wouldn't have lists of strings

libs/functions/calculate_projections.py
goldblatt: done

libs/datasets/model_output_schema.py
goldblatt: when we pull in the data we ignore them. they should be named lol

libs/datasets/projections_schema.py
goldblatt: updated the names

libs/datasets/model_output_schema.py
goldblatt: ranemd to CAN_MODEL_OUTPUT_SCHEMA

PR #150, opened 2020-04-08
Add run.sh script to generate all API artifacts.

mikelehen:


run.sh
```diff
@@ -0,0 +1,89 @@
...
+
+set -o nounset
+set -o errexit
+
+DEFAULT_DATA_SOURCES_DIR="../covid-data-public"
+DEFAULT_API_RESULTS_DIR="api_results"
+
+main () {
+  # Go to repo root (where run.sh lives).
+  cd "$(dirname "$0")"
```
ghop02, line 14: you could do this using `pushd` and `popd` so that it doesn't change your directory after the run

run.sh
```diff
@@ -0,0 +1,89 @@
...
+  # TODO: Move DoD onto a more general-purpose schema rather than treat them custom.
+  DOD_DIR="${API_OUTPUT_DIR}/custom1"
+
+  echo ">>> Generating state models to ${STATES_DIR}"
+  # TODO(#148): We need to clean up the output of these scripts!
+  ./run_model.py state -o "${API_OUTPUT_DIR}" > /dev/null
+
+  echo ">>> Generating county models to ${COUNTIES_DIR}"
+  # TODO(#148): We need to clean up the output of these scripts!
+  ./run_model.py county -o "${COUNTIES_DIR}" > /dev/null
```
ghop02, line 64: by running this inside a function, i don't think it'll automatically fail on a non zero response.  
a) is this true?
b) if so we should probably check the return value and exit if necessary

run.sh
mikelehen: `cd` within a bash script doesn't affect the current directory of the caller.  So `pushd` / `popd` isn't necessary.

run.sh
mikelehen: a) not true AFAIK!

My understanding is that the `set -o errexit` at the top will automatically fail the entire script if any command fails.  Feel free to prove me wrong. 😛

run.sh
ghop02: time to go brush up on my bash 😀

run.sh
ghop02: you are right!

.gitignore
```diff
@@ -1,5 +1,6 @@
 .~lock.*
 results/*
+api_results/
```
fridiculous, line 3:
```suggestion
api_results/*
```

run.sh
```diff
@@ -0,0 +1,89 @@
...
+  # requirement that the county model JSON is in a /county subdirectory of the
+  # states?
+  COUNTIES_DIR="${API_OUTPUT_DIR}/county";
+  COUNTY_SUMMARIES_DIR="${API_OUTPUT_DIR}/county_summaries";
+  CASE_SUMMARIES_DIR="${API_OUTPUT_DIR}/case_summary"
+
+  # TODO: Move DoD onto a more general-purpose schema rather than treat them custom.
+  DOD_DIR="${API_OUTPUT_DIR}/custom1"
+
+  echo ">>> Generating state models to ${STATES_DIR}"
```
fridiculous, line 58: it seems like up to this point, we're just setting vars.  could we just make that as part of a `prepare()` function?

run.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# Helper for getting absolute paths.
+function absPath() {
+  (
+  cd "$(dirname $1)"
+  echo "$PWD/$(basename $1)"
+  )
+}
+
+main "$@"
```
fridiculous, line 89: How would someone know which arguments to send?   Should we put that in a readme?

Although i'm realizing that this script takes either no arguments.  Would it make sense to make it invoke a usage method if empty instead?

```
usage() {
  echo "USAGE: ${BASH_SOURCE[0]} <.> ..."
  echo ""
  echo "Example:
  ${BASH_SOURCE[0]} <a> ....
  ${BASH_SOURCE[0]} <b> ....
  "
  exit 0
}
```

PR #152, opened 2020-04-08
running black formatter on functions

ghop02:
This just runs the autoformatter on the functions file! 

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/functions/generate_shapefiles.py
+++ b/libs/functions/generate_shapefiles.py
@@ -5,24 +5,45 @@ from urllib.parse import urlparse
 from libs.CovidDatasets import get_public_data_base_url
 from libs.constants import NULL_VALUE
 
+
 def _file_uri_to_path(uri: str) -> str:
     return urlparse(uri).path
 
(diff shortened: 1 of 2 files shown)


libs/functions/generate_shapefiles.py
```diff
@@ -36,34 +57,54 @@ def join_and_output_shapefile(df, shp_reader, pivot_shp_field, pivot_df_column,
             # Gets the row of the dataframe that matches the FIPS codes for a state/county
             row = df[df[pivot_df_column] == shapeRecord.record[pivot_shp_field]].iloc[0]
         except Exception as e:
-            state_fips = shapeRecord.record[pivot_shp_field][:2] # state fips is the first two chars of the state/county fips
-            failed_dictionary.setdefault(state_fips, []).append(shapeRecord.record[pivot_shp_field])
+            state_fips = shapeRecord.record[pivot_shp_field][
```
goldblatt, line 60: this one hurts me

ghop02: yeah, for these I generally try to rewrite them so that it doesn't happen (in this case it would just be moving the comment on top)

Issue #145, opened 2020-04-07
Deep Clean & Re-Structure of Repo


fridiculous:
As suggested from the [data api doc]( https://docs.google.com/document/d/1VhQqMxtVYHQMIj9zFeb1DUQZ1aq1k4xIc_bYoTIqUb4/edit?ts=5e8a3a1e#) we need a separation of concerns between different parts of the application.

In general, the repo structure should treat the following implementations independently:
1) How datasets get defined, loaded, and sanitized - `libs/datasets/sources`
2) How datasets get defined as a canonical source for that type of data - `libs/datasets/models`
3) How functions and general transformations are defined - `libs/functions/transforms`
4) How statistical functions are defined `libs/functions/models`
5) How pipelines are defined as steps for functions and dataset steps - `libs/pipelines`
6) How the resultant API schema is defined and written to - `api`
7) How the pipelines are triggered - `cli`
8) How triggerable pipelines are automated - `.github/workflows/`
9) Where notebooks & exploratory analyses are kept - `analyses`

As a result, I propose we use the following structure to more cleanly separate the varying responsibilities needed within this repo.
```bash
├── .github/workflows
├── analyses
├── api
├── cli
├── libs
│   ├── datasets
│   │   ├── sources
│   │   └── api
│   └── models
│   │   ├── ensemble
│   │   ├── inference
│   │   └── seir
│   └── pipelines
└── tests
    ├── data
    └── results
```

Much of this work would require moving the existing repo to match this structure, including moving some functions around to live in their target homes.

EDIT:  Taking in notes and updating the hierarchy above

ghop02:
Couple questions/comments:
 - `functions` feels a bit broad do we need this structure? 
 - I'm a bit leery of separating "transforms" away from data sources/data models. I'm just not 100% sure what I expect to live there.
 - Why are models in functions? If i'm looking at `libs/` without knowing anything, seeing `datasets` and `models` at the top level feels clearer than models living in functions. 
 - What goes in `dataset/models`? the word model will always be overloaded in a project that involves actual models :P.

In general I liked the flatter structure a bit more, to me the data structures that will define these matter a bit more than the folder structure, and I don't quite have a handle on that. 

It might be helpful for someone to do a quick run through of applying this structure (or other) to the dod "pipeline" as a draft PR to see a bit more concretely how those changes manifest/what feels right.

BrettBoval:
done

PR #154, opened 2020-04-08
moving datset deployer out of deploy_dod_dataset

ghop02:
More work on 145 https://github.com/covid-projections/covid-data-model/issues/145

Moves logic out of deploy_dod_dataset + adds a helper function to deploy all shapefiles

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/dataset_deployer.py
+++ b/libs/dataset_deployer.py
@@ -1,0 +1,94 @@
+import os
+import io
+import logging
+import boto3
+
+_logger = logging.getLogger(__name__)
+
+
+class DatasetDeployer(object):
+    """Common deploy operations for persisting files to s3 or local folder.
+
+    Deploys to S3 if the BUCKET_NAME environment variable is set.
+    """
+
+    def __init__(self, key="filename.csv", body="a random data", output_dir="."):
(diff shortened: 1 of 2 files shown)


libs/dataset_deployer.py
```diff
@@ -0,0 +1,94 @@
+import os
+import io
+import logging
+import boto3
+
+_logger = logging.getLogger(__name__)
```
fridiculous, line 6: what do you think if this lives in api or some persistance folder?

ghop02: I don't think there is much advantage to having the functionality for deploying live in the "api" folder just yet. For instance, I'm not sure exactly what logic will actually live in the API folder.  Some of that logic I think belongs in pipelines? If the API folder was just the schema objects that would make sense to me.

Until the folder more completely encapsulate the behavior, I'd prefer not creating them?

I saw something once that you don't need a folder until there are more than 8 files in a directory and I kinda like.  Once a folder starts getting overloaded i think you can identify similar functionality a bit easier.

fridiculous: Noted.

PR #156, opened 2020-04-08
moving core of dod dataset generation to its own pipeline

ghop02:
This PR addresses more of  https://github.com/covid-projections/covid-data-model/issues/155.

Adds a `pipelines` directory where the bulk of the logic now lives.  The CLI is now a very thin wrapper. 

### Please Check
- [ ] Are tests passing?


deploy_dod_dataset.py
```diff
@@ -1,58 +1,30 @@
...
-from libs import dataset_deployer
+from libs.pipelines import dod_pipeline
 logger = logging.getLogger(__name__)
 PROD_BUCKET = "data.covidactnow.org"
 
 
 @click.command()
 @click.option('--run_validation', '-r', default=True, help='Run the validation on the deploy command')
-@click.option('--input', '-i', default='results', help='Input directory of state/county projections')
+@click.option('--input-file', '-i', default='results', help='Input directory of state/county projections')
```
fridiculous, line 13: file or directory?

libs/pipelines/dod_pipeline.py
```diff
@@ -0,0 +1,87 @@
...
+PROD_BUCKET = "data.covidactnow.org"
+
+DodInterventionResult = namedtuple(
+    "DodInterventionResult", ["key", "projection_df", "shapefile_data"]
+)
+
+
+def run_projections(
+    input_file, intervention: Intervention, run_validation=True
+) -> Tuple[DodInterventionResult, DodInterventionResult]:
```
fridiculous, line 21: it would be good to know which one is for state or county, or if can be abstracted by geos

fridiculous:
Ty for starting this.  The thing that this prompts in me, is what is the internal datastructure of a pipeline.  is it simply an array, aka a series of steps at this point?

Should we take it a step further by separating out tasks and order?

ghop02:
Yeah that's a good question.  I kinda feel like we'll better figure that out as we incrementally add more structure?

libs/pipelines/dod_pipeline.py
ghop02: yeah i originally had this as "run_county_projections" and "run_state_projections", would you prefer that? I dont' think we're quite stable enough to have the same code runs on just a different aggregation

PR #158, opened 2020-04-08
adding the current intervention logic

goldblatt:
Adds the current intervention to the output. Plus some extra fields (which were still kind of annoying to add TBH).


libs/functions/calculate_projections.py
```diff
@@ -78,11 +81,20 @@ def _calculate_projection_data(file_path):
             mean_deaths,
             peak_hospitalizations_date,
             peak_deaths_date,
+            hospitals_overcapacity_date,
```
fridiculous, line 84: overcapacity == short_fall?

libs/functions/generate_shapefiles.py
```diff
@@ -33,6 +33,7 @@ def join_and_output_shapefile(
         "State Intervention",
         "PEAK-HOSP",
         "PEAK-DEATHS",
+        "Hospital Over Capacity On",
```
fridiculous, line 36: Not sure if this is the suggested field, but
```suggestion
        "Hospital Overcapacity Date",
```

libs/functions/calculate_projections.py
goldblatt: changed the name

PR #176, opened 2020-04-09
Add deploy-api Github Action workflow.

mikelehen:
* Runs run.sh and publishes the results to data.covidactnow.org/snapshot/<run-id>
* Scheduled for 12:30 daily, which should be after the relevant covid-data-public
  update.

@fridiculous Before we merge this we need to add the AWS secrets for the data bucket to github (as AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY).  Can you do that?

### DIFF
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -1,0 +1,75 @@
+# To debug, it's recommended you modify / use the version in the test-actions repo:
+# https://github.com/covid-projections/test-actions/blob/master/.github/workflows/deploy_api.yml
+
+name: Build / Publish API artifacts to data.covidactnow.org
+
+on:
+  # JHU posts new data at 11:30 (UTC).
+  # covid-data-public fetches it at 12:00 (UTC).
+  # So we rebuild / publish the API at 12:30 (UTC).
+  schedule:
+   - cron: '30 12 * * *'
+
+  # Hook to trigger a manual run.
+  # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
+  repository_dispatch:
+    types: publish-api
+
+jobs:
+  publish-api:
+    runs-on: ubuntu-latest
+
+    env:
+      AWS_S3_BUCKET: 'data.covidactnow.org'
+      # To pin to an old data set or model code, put the branch/tag/commit here:
+      COVID_DATA_PUBLIC_REF: 'master'
+      COVID_DATA_MODEL_REF: 'master'
+      SNAPSHOT_ID: ${{github.run_number}}
+
+    steps:
+    - name: Checkout covid-data-model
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_api.yml
```diff
@@ -0,0 +1,75 @@
...
+# https://github.com/covid-projections/test-actions/blob/master/.github/workflows/deploy_api.yml
+
+name: Build / Publish API artifacts to data.covidactnow.org
+
+on:
+  # JHU posts new data at 11:30 (UTC).
+  # covid-data-public fetches it at 12:00 (UTC).
+  # So we rebuild / publish the API at 12:30 (UTC).
+  schedule:
+   - cron: '30 12 * * *'
```
mikelehen, line 11: Pretty sure this is right, but I don't usually write cron jobs, so please double check me.  The intention is to run it at 12:30 UTC (30 minutes after noon) daily.

.github/workflows/deploy_api.yml
```diff
@@ -0,0 +1,75 @@
+# To debug, it's recommended you modify / use the version in the test-actions repo:
+# https://github.com/covid-projections/test-actions/blob/master/.github/workflows/deploy_api.yml
+
+name: Build / Publish API artifacts to data.covidactnow.org
```
fridiculous, line 4:
```suggestion
name: Build & Publish API artifacts to data.covidactnow.org
```

.github/workflows/deploy_api.yml
fridiculous: looks good!

.github/workflows/deploy_api.yml
```diff
@@ -0,0 +1,75 @@
...
+
+    - name: Deploy Artifacts to S3 (https://data.covidactnow.org/snapshot/${{env.SNAPSHOT_ID}}/).
+      uses: jakejarvis/s3-sync-action@master
+      with:
+        args: --acl public-read --follow-symlinks --delete
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        SOURCE_DIR: './api_results/'
+        DEST_DIR: 'snapshot/${{env.SNAPSHOT_ID}}'
```
fridiculous, line 73: for whatever reason, having the api structure defined here makes me feel a bit uneasy but i'm ok with it for now

mikelehen: Yeah.  That's fair.  I thought about having run.sh spit out files under `/snapshot/123/` itself, but the upload to S3 needs to be for the specific destination directory (so we don't overwrite previous snapshots, etc.).  So I think it kinda' has to live here.

Issue #110, opened 2020-04-04
Make built_dod_dataset.py not read CSVs (poorly) directly.  Use JHUDataset -> Timeseries dataset


igorkofman:
As is, it's fragile to what date it is vs what date the latest dataset is from and likely other issues

mikelehen:
This broke another automation run:  https://github.com/covid-projections/covid-data-model/actions/runs/80276073

Can we get a fix (even if it's hacky) in soooooon?

mikelehen:
Fixed by #254.

PR #186, opened 2020-04-10
Pulling actual case values from JHUDataSource

ghop02:
This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/110

Still need to clean this up a bit more, but basically swaps out reading the jhu files from the latest file to using the JHU data source as a common timeseries. 

As a result some of the column names changed and needed to add country level fips.

I still need to verify the output to make sure it matches with the existing/surface any differences. 

### Please Check
- [ ] Are tests passing?


libs/build_dod_dataset.py
```diff
@@ -49,145 +52,82 @@ def _get_abbrev_df():
...
         .merge(interventions_df, left_on="state", right_on="state", how="inner")
     )
-
     counties_remapped = counties_decorated.rename(
         columns=OUTPUT_COLUMN_REMAP_TO_RESULT_DATA
     )
     counties = pd.DataFrame(counties_remapped, columns=RESULT_DATA_COLUMNS_COUNTIES)
     counties = counties.fillna(NULL_VALUE)
     counties.index.name = "OBJECTID"
+    counties["Province/State"] = counties["Province/State"].map(abbrev_us_state)
```
goldblatt, line 90: i mean would love to use the class vars here if we have them?

libs/build_dod_dataset.py
```diff
@@ -49,145 +52,82 @@ def _get_abbrev_df():
...
-            "Active": "sum",
-            "Country/Region": "first",
-            "Latitude": "first",
-            "Longitude": "first"
-            # People tested is currently null
-            #'People Tested': 'sum'
-        }
-    )
-
-    # basically the states_agg has full state names, the interventions have abbreviation so we need these to be merged
```
goldblatt, line 166: why does this change?

ghop02: This is all handled in the Data source/timeseries preprocessing, the datasets coming out of it are already one row per state

libs/build_dod_dataset.py
ghop02: word. I think I'll be able to use your latest PR for that right?

libs/build_dod_dataset.py
```diff
@@ -49,145 +52,82 @@ def _get_abbrev_df():
...
+    return data.latest_values(AggregationLevel.STATE)
 
 
 def get_usa_by_county_with_projection_df(input_dir, intervention_type):
     us_only = _get_usa_by_county_df()
-    fips_df = FIPSPopulation.local().data  # used to get interventions
-    interventions_df = (
-        _get_interventions_df()
-    )  # used to say what state has what interventions
+    fips_df = FIPSPopulation.local().data
```
goldblatt, line 68: do we need to specify this is county data?

ghop02: it works okay because the left join is only starting with counties, but might be good to make it explicit.

PR #188, opened 2020-04-10
Move to common cli

ghop02:
This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/145

Okay - a bit larger change.

This just hooks in all of the various command lines into one master cli so it's easier to get around and gives a common home for all cli functionality. 

I didn't do any restructuring right now, just wanted to keep the current structure as closely as possible.

I'm running all of the deploy scripts right now to verify that everything hooks together.

```
(covid-data-model) ➜  covid-data-model git:(move-to-common-cli) ✗ ./run.py
Usage: run.py [OPTIONS] COMMAND [ARGS]...

  Entry point for covid-data-model CLI.

Options:
  --help  Show this message and exit.

Commands:
  data        Generate reports from raw data.
  deploy-dod  Generates and runs dod data projections from model outputs.
  model       Run models
```

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/cli/run_model.py
+++ b/cli/run_model.py
@@ -9,8 +9,9 @@ from libs.pipelines import can_model_pipeline
 _logger = logging.getLogger(__name__)
 
 
-@click.group()
+@click.group('model')
 def main():
+    """Run models"""
     pass
 
 
--- a/cli/run_data.py
+++ b/cli/run_data.py
@@ -15,8 +15,9 @@ from libs.datasets import data_version
 _logger = logging.getLogger(__name__)
 
 
-@click.group()
+@click.group('data')
 def main():
+    """Generate reports from raw data."""
     pass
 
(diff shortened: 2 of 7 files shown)


cli/run_model.py
```diff
@@ -9,8 +9,9 @@
 _logger = logging.getLogger(__name__)
 
 
-@click.group()
+@click.group('model')
 def main():
+    """Run models"""
```
erccarls, line 14: Since you are using click you can just add my modeling as a subcommand directly right now. 

If you have an existing `group` you can use `group.add_command(model)` where model is another group that can be imported from my libs. This allows for the same modeling subcommands to be used as well.

ghop02: yeah, excited for that :)

Issue #185, opened 2020-04-10
Generate schema for existing county summary data pipeline with tests


ghop02:
This is for generating a schema flow for outputting county summary data.
It can take some ideas from [this pr](https://github.com/covid-projections/covid-data-model/pull/159/files).

 - Independent python file containing only the pydantic object for outputting data.
 - command (preferably in the cli) to generate all schemas and save them as json-schema formats.
 - pipeline file to take dataframes and write output data in the expected schema

BrettBoval:
Can we close @ghop02?

PR #189, opened 2020-04-10
Add state summary schema

ghop02:
This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/185

Adds the state summary schema as a public API schema. Any schemas added to the `api/` folder will generate via the command-line update-schemas function.

### Update a schema
```
./run.py api update-schemas
```

### Please Check
- [x] Are tests passing?


api/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+import pydantic
+
+_logger = logging.getLogger(__name__)
+
+SCHEMAS_PATH = pathlib.Path(__file__).parent / "schemas"
+
+
+def load_public_schemas() -> List[Type[pydantic.BaseModel]]:
+    """Loads all pydantic Schemas for export.
```
mikelehen, line 15: "Loads" => "Finds"
"Schemas" => "model classes (i.e. that derive from pydantic.BasModel)" ?
Drop "for export" since that's an assumption about how this function will be used.

api/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+    If a class is prefixed with a '_' (i.e. _MySchema), it will not be exported
+    as a public schema.
+
+    Performs a bit of python magic to load all modules in the api/ folder and
+    find all subclasses of the base model.
```
mikelehen, line 21: "the base model" => "pydantic.BaseModel" (always good to be as explicit as possible)

api/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Returns: List of api schema specifications.
```
mikelehen, line 23: "api schema specifications" => "api model classes" ?

FWIW feel free to tell me if I don't understand python types / terminology here.

api/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    # Importing all python files in api/ to collect all schemas for public api.
+    # This is needed to load all subclasses of pydantic.BaseModel.
+    # If a module is not in the global modules, cls.__subclasses__
+    # will not be able to locate it.
```
mikelehen, line 28: I don't python well, but I think this comment could be clearer / more explicit, e.g.:

> Find all python files in api/ and load them as global modules. 
> This allows us to just enumerate pydantic.BaseModel.__subclasses__ to find all our model classes.

api/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    for path in root.rglob("*.py"):
+        relative = path.relative_to(root.parent)
+        module_name = str(relative).replace("\\", "/").replace("/", ".")[:-3]
+
+        spec = importlib.util.spec_from_file_location(module_name, str(path))
+        mod = importlib.util.module_from_spec(spec)
+        sys.modules[spec.name] = mod
+        spec.loader.exec_module(mod)
+
+    schemas = []
```
mikelehen, line 39: schemas => modelClasses or similar?

api/state_actuals_summary.py
```diff
@@ -0,0 +1,19 @@
+from typing import List
+import pydantic
+
+
+class _CountyCases(pydantic.BaseModel):
+    fips: str = pydantic.Field(..., description="FIPS state code + FIPS county code")
+    cases: int = pydantic.Field(..., description="Cumulative case count.")
+    deaths: int = pydantic.Field(..., description="Cumulative deaths count.")
+    date: str = pydantic.Field(..., description="Date of latest data")
```
mikelehen, line 9: FWIW- This should be an ISO 8601 date string, so we should put that in the description, and then make sure that it is.

ghop02:
These are great, thanks @mikelehen

api/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    find all subclasses of the base model.
+
+    Returns: List of api schema specifications.
+    """
+    # Importing all python files in api/ to collect all schemas for public api.
+    # This is needed to load all subclasses of pydantic.BaseModel.
+    # If a module is not in the global modules, cls.__subclasses__
+    # will not be able to locate it.
+    root = pathlib.Path(__file__).parent
+    for path in root.rglob("*.py"):
```
goldblatt, line 30: i'd probably just to like list in dir here? but idk this works?

.pylintrc
```diff
@@ -0,0 +1,5 @@
+[TYPECHECK]
+# Pydantic seems to fail the type check error because
+# it fails to find BaseModel, unclear what the issue
+# is as it pydantic.BaseModel definitely exists.
```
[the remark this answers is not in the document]
ghop02: https://github.com/samuelcolvin/pydantic/issues/992
I guess there is one. maybe I can try the cython approach.

ghop02:
> oh wait shouldn't this update the rquirements.txt?

I think i did!

test/api_schema_test.py
```diff
@@ -0,0 +1,14 @@
+import json
+import api
+
+
+def test_schemas_up_to_date():
```
fridiculous, line 5: What do you envision calling this method?

PR #191, opened 2020-04-10
consolidating fields even more

goldblatt:
Intermediate Step to get to using schemas. 

### Please Check
- [ ] Are tests passing?


libs/functions/generate_shapefiles.py
```diff
@@ -14,35 +14,10 @@ def _file_uri_to_path(uri: str) -> str:
...
-
-    fields = [field for field in df.columns if field not in blacklisted_fields]
-
     shp_writer.fields = shp_reader.fields  # Preserve fields that come from the census
 
-    for field_name in fields:
-        if field_name in non_integer_fields:
+    for field_name in SHAPEFILE_FIELDS:
+        if field_name in NON_INTEGER_FIELDS:
             shp_writer.field(field_name, "C", size=32)
```
ghop02, line 21: I realize this isn't in your change, but what does this `"C"` stand for?

goldblatt: probably char?

PR #192, opened 2020-04-10
Data Pipeline for top 100 counties

goldblatt:
This PR addresses issue https://github.com/covid-projections/covid-projections/issues/479

### Please Check
- [?] Will the *current* schema in [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.md) be updated with this PR?
- [ ] Are tests passing?


api/can_predictions.py
```diff
@@ -0,0 +1,26 @@
...
+aggregateDeaths
+"""
+class _CountyCases(pydantic.BaseModel):
+    fips: str = pydantic.Field(..., description="FIPS state code + FIPS county code")
+    cases: int = pydantic.Field(..., description="Cumulative case count.")
+    deaths: int = pydantic.Field(..., description="Cumulative deaths count.")
+    date: str = pydantic.Field(..., description="Date of latest data")
+
+
+class StateCaseSummary(pydantic.BaseModel):
```
ghop02, line 19: I guess I'm not sure how this schema here quite relates to the one in in the top counties pipeline. I feel like the columns should be what's in the docstring above

I would expect a step in the `top_counties_pipeline` to be conforming the data to this.

libs/pipelines/top_counties_pipeline.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+    return county_results
+
+def generate_api(intervention_result: TopCountiesPipelineResult): 
+    projection = intervention_result.projection_df
+
+    api_results = []
+    # TODO: I think pandas doesn't really vibe with iterating, but it's 3000 and the other option is applying 
+    # things in a weird way so idk this is anti pandas but w/e
```
ghop02, line 51: oh that works, or you can do `projection.to_dict(orient='records')` and it will return a list of dictionaries. 
that might be the same thing that's happening here? what type is `county_row`?

But I actually think iterating here is the right thing to do in general.

goldblatt:
format outputted like. this is a subset of the full data for the website? 
> {
  "data": [
    {
      "stateName": "New York",
      "countyName": "New York City",
      "fips": "36061",
      "lastUpdatedDate": "4/9/2020 23:02",
      "projections": {
        "aggregateDeaths": 2199,
        "hospitalBeds": {
          "peakShortfall": 24949,
          "peakDate": "2020-05-15T00:00:00",
          "shortageStartDate": "2020-04-10T00:00:00"
        }
      }
    },
]

ghop02:
Can you rebase against master + throw this in the cli folder?

deploy_dod_dataset.py
```diff
@@ -3,7 +3,7 @@
 import logging
 
 from libs.enums import Intervention
-from libs.pipelines import dod_pipeline
+from libs.pipelines import dod_pipeline, top_counties_pipeline
```
ghop02, line 6: best to throw these on separate lines (standard is one import per line), but you probably can actually remove this

libs/dataset_deployer.py
```diff
@@ -67,6 +67,11 @@ def upload_csv(key_name: str, csv: str, output_dir: str):
     _logger.info(f"Generated csv for {key_name}")
 
 
+def upload_json(key_name, json: str, output_dir: str):
```
ghop02, line 70: I don't really know if there's a ton of value in using the deployer, but hey it keeps it consistent

libs/functions/generate_api.py
```diff
@@ -0,0 +1,25 @@
...
+    if isinstance(panda_date_or_none, str):
+        return None
+    return panda_date_or_none.to_pydatetime()
+
+def generate_api_for_projection(projection): 
+    api_results = []
+    # TODO: I think pandas doesn't really vibe with iterating, but it's 3000 and the other option is applying 
+    # things in a weird way so idk this is anti pandas but w/e 
+    for index, county_row in projection.iterrows(): 
+        peak_date = _get_date_or_none(county_row[rc.PEAK_HOSPITALIZATIONS])
```
ghop02, line 16: When does this case happen?

requirements.txt
```diff
@@ -20,3 +20,4 @@ simplejson==3.17.0
 boto3==1.12.28
 seaborn==0.10.0
 click==7.1.1
+pydantic==1.0.0
```
ghop02, line 23: I think this should be 1.4 (weird naming on their part)

api/can_predictions.py
```diff
@@ -0,0 +1,28 @@
...
+import datetime
+
+"""
+Documented https://www.dropbox.com/scl/fi/o4bec2kz8dkcxdtabtqda/CAN-V1-API-Draft.paper?dl=0&rlkey=f7elx3zmo6tt5s7mj5nvap9a0
+
+"""
+
+class _HospitalBeds(pydantic.BaseModel):
+    peakShortfall: int = pydantic.Field(default=0, description="Shortfall of beds needed at the peek hospitalizaitons")
+    peakDate: Optional[datetime.datetime] = pydantic.Field(default='', description="Date of peak hospitalizations")
```
ghop02, line 12: does json have a `null` field? probably should use `None` instead here

libs/pipelines/top_counties_pipeline.py
```diff
@@ -0,0 +1,67 @@
...
+
+    counties_key_name = f"counties.{intervention.name}"
+    # note i think build_processed_dataset should porbably be renamed? 
+    counties_df = build_processed_dataset.get_usa_by_county_with_projection_df(
+        input_file, intervention.value
+    )
+    if run_validation:
+        validate_results.validate_counties_df(counties_key_name, counties_df)
+
+    county_results = TopCountiesPipelineProjectionResult(
```
ghop02, line 48: I don't 100% know how I feel about this suggestion, but I think it would be so so cool if this returned a `DataSource` subclassed object, that was like:

```
CountyProjections(DataSource):
    class Fields(object):
        ...<all of the fields in results_schema>
```

Then in the future, we could do validation to make sure when we load something it's reasonable

api/can_predictions.py
```diff
@@ -0,0 +1,28 @@
...
+    shortageStartDate: Optional[datetime.datetime] = pydantic.Field(default='', description="Date when hospitals overload")
+
+
+class _Projections(pydantic.BaseModel):
+    aggregateDeaths: int = pydantic.Field(...)
+    hospitalBeds: _HospitalBeds = pydantic.Field(...)
+
+class CANPredictionAPIRow(pydantic.BaseModel):
```
ghop02, line 20: Alternate name idea: `CountyProjectionSummary` (not sure how i feel about summary).

I think having `CAN` in the name is a bit unnecessary (I've probably used it elsewhere tho)

api/can_predictions.py
```diff
@@ -0,0 +1,28 @@
...
+    stateName: str = pydantic.Field(..., description="The state name")
+    countyName: str = pydantic.Field(..., description="The county name")
+    fips: str = pydantic.Field(..., description="Fips for State + County. Five character code")
+    lastUpdatedDate: str = pydantic.Field(..., description="Date of latest data")
+    projections: _Projections = pydantic.Field(...)
+
+class CANPredictionAPI(pydantic.BaseModel):
```
ghop02, line 27: We can get even more specific here - idea:
```
class Top100CountyOverloadSummary(pydantic.BaseModel):
    data: List[CANPredictionAPIRow] = pydantic.Field(..., max_rows=100)
```

libs/pipelines/top_counties_pipeline.py
```diff
@@ -0,0 +1,67 @@
...
+        counties_key_name, counties_df
+    )
+
+    return county_results
+
+def generate_api(projection_result: TopCountiesPipelineProjectionResult, sort_fields = [rc.MEAN_DEATHS],length = 100) -> TopCountiesPipelineResult: 
+    projection = projection_result.projection_df
+    sorted_limited = projection.sort_values(by=sort_fields, ascending=False).head(length)
+    return TopCountiesPipelineResult(projection_result.key, generate_api_for_projection(sorted_limited))
```
ghop02, line 57: can you do 
```
from libs.functions import generate_api
...
generate_api.generate_api_for_projection(sorted_limited)
```
very when reading to know where it came from

libs/pipelines/top_counties_pipeline.py
goldblatt: okay i will take out this comment

deploy_dod_dataset.py
goldblatt: cool yes done removed

libs/dataset_deployer.py
goldblatt: yeah mostly the consistency

libs/functions/generate_api.py
goldblatt: i added a comment for this it's basically our dates are null for this when there isn't a shortfall in beds

requirements.txt
goldblatt: done

api/can_predictions.py
goldblatt: removed that

api/can_predictions.py
goldblatt: i like keeping API in there too but i can use projection although thtere is non predicrted data

api/can_predictions.py
goldblatt: this was still WIP stuff it's all changed now

api/can_predictions.py
goldblatt: hmm this messes with that genrator file i wrote so gnna lve as is

goldblatt:
hmm the tests are failing in CI but not on my local branch. im going to commit this and then see if it still fails CI

Issue #201, opened 2020-04-12
Beds data has duplicate fips


ghop02:
Looks like a recent change (that I probably made) caused the bed data to break.

Would be great to add some better testing around this to make sure beds data properly loads.

davidstrauss:
I assume this is fixed now, but maybe we should open a new issue for adding the testing before closing this?

PR #202, opened 2020-04-12
Fixing duplicate county data in beds

ghop02:
This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/201

It also pulls down the covid-data-public repo so that we can write tests on the actual data sources we have.  I think in the ideal world I would rather have smaller unit test coverage, but this will at least give us an opportunity to make sure PRs aren't breaking existing behavior.

### Please Check
- [x] Are tests passing?

### DIFF
--- a/libs/datasets/dataset_utils.py
+++ b/libs/datasets/dataset_utils.py
@@ -198,7 +198,7 @@ def add_county_using_fips(data, fips_data):
     is_county = data.aggregate_level == AggregationLevel.COUNTY.value
     # Only want to add county names to county level data, so we'll slice out the county
     # data and combine it back at the end.
-    not_county = data[~is_county]
+    not_county_df = data[~is_county]
     data = data[is_county]
     fips_data = fips_data[fips_data.aggregate_level == AggregationLevel.COUNTY.value]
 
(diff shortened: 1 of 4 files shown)


libs/datasets/dataset_utils.py
```diff
@@ -195,6 +195,13 @@ def build_fips_data_frame():
 
 
 def add_county_using_fips(data, fips_data):
+    is_county = data.aggregate_level == AggregationLevel.COUNTY.value
+    # Only want to add county names to county level data, so we'll slice out the county
+    # data and combine it back at the end.
+    not_county = data[~is_county]
```
goldblatt, line 201: can we add _df to the end also it's wild to me that pandas works like this

libs/datasets/dataset_utils.py
```diff
@@ -195,6 +195,13 @@ def build_fips_data_frame():
...
+    data = data[is_county]
+    fips_data = fips_data[fips_data.aggregate_level == AggregationLevel.COUNTY.value]
```
goldblatt, line 203: i mean kinda wish we had a different dataset for state fips and county fips?

ghop02: Yeah, I do wonder if we should just be separating out county/state/country data in the data source itself

PR #203, opened 2020-04-12
[Model] Refactor MLE Fitter

erccarls:
WIP

### Please Check
- [ ] Are tests passing?


pyseir/load_data.py
```diff
@@ -303,15 +314,92 @@ def load_new_case_data_by_fips(fips, t0):
...
+        return None, None, None
+
+    times_new = (hospitalization_data['date'].dt.date - t0.date()).dt.days.values
+
+    if (hospitalization_data['current_hospitalized'] > 0).any():
+        return times_new, hospitalization_data['current_hospitalized'].values.clip(min=0), 'current'
+    elif (hospitalization_data['cumulative_hospitalized'] > 0).any():
+        return times_new, hospitalization_data['cumulative_hospitalized'].values.clip(min=0), 'cumulative'
+    else:
+        return None, None, None
```
zhang-xinyu, line 360: What is 'current'? Is it possible to pass a parameter to determine which category fo return?

pyseir/load_data.py
```diff
@@ -303,15 +314,92 @@ def load_new_case_data_by_fips(fips, t0):
...
+        Datetime to offset by.
+
+    Returns
+    -------
+    times: array(float) or NoneType
+        List of float days since t0 for the hospitalization data.
+    observed_hospitalizations: array(int) or NoneType
+        Array of new cases observed each day.
+    type: str
+        'cumulative' or 'current' or NoneType
```
zhang-xinyu, line 382: same question here...Wouldn't using string types make downstream analysis more vulnerable to error?

pyseir/load_data.py
```diff
@@ -303,15 +314,92 @@ def load_new_case_data_by_fips(fips, t0):
...
+    if len(hospitalization_data) == 0:
+        return None, None, None
+
+    times_new = (hospitalization_data['date'].dt.date - t0.date()).dt.days.values
 
+    if (hospitalization_data['current_hospitalized'] > 0).any():
+        return times_new, hospitalization_data['current_hospitalized'].values.clip(min=0), 'current'
+    elif (hospitalization_data['cumulative_hospitalized'] > 0).any():
+        return times_new, hospitalization_data['cumulative_hospitalized'].values.clip(min=0), 'cumulative'
+    else:
```
zhang-xinyu, line 398: Is there a reason that hospitalization data can be negative?

pyseir/models/suppression_policies.py
```diff
@@ -262,32 +294,22 @@ def generate_empirical_distancing_policy_by_state(t_list, state, future_suppress
...
-                                        future_suppression=future_suppression,
-                                        reference_start_date=reference_start_date)
 
-    weight = county_metadata[county_metadata.state == state][['fips', 'total_population']] \
-                                .set_index('fips').rename(columns={'total_population': 'weight'})
-    
-    results = dict()
+    # Aggregate the counties to the state level, weighted by population.
+    weight = county_metadata.loc[county_metadata.state == state, 'total_population'].values
+    weight = weight / weight.sum()
```
zhang-xinyu, line 305: Not sure if the order of county will remain the same as that of results, better use: 
```
weight = county_metadata.set_index('fips').loc[counties_fips].values
weight = weight / weight.sum()
```

pyseir/inference/model_fitter.py
```diff
@@ -0,0 +1,478 @@
...
+                 cases_to_deaths_err_factor=1,
+                 hospital_to_deaths_err_factor=1):
+
+        self.fips = fips
+        self.ref_date = ref_date
+        self.min_deaths = min_deaths
+        self.t_list = np.linspace(0, int(365 * n_years), int(365 * n_years) + 1)
+        self.cases_to_deaths_err_factor = cases_to_deaths_err_factor
+        self.hospital_to_deaths_err_factor = hospital_to_deaths_err_factor
+        self.t0_guess = 60
```
zhang-xinyu, line 61: Should t0_guess also be a parameter?

pyseir/inference/model_fitter.py
```diff
@@ -0,0 +1,478 @@
...
+            R0=3.4, limit_R0=[1, 8], error_R0=1.,
+            log10_I_initial=2, limit_log10_I_initial=[0, 5], error_log10_I_initial=.5,
+            t0=60, limit_t0=[0, 70], error_t0=1,
+            eps=.3, limit_eps=[0, 2], error_eps=.2,
+            t_break=20, limit_t_break=[0, 100], error_t_break=2,
+            test_fraction=.05, limit_test_fraction=[0.001, 1], error_test_fraction=.01,
+            hosp_fraction=1, limit_hosp_fraction=[0.001, 1], error_hosp_fraction=.1,
+            fix_hosp_fraction=self.hospitalizations is None, # Let's not fit this to start...
+            errordef=.5
+        )
```
zhang-xinyu, line 97: Better put these into a config file.

pyseir/inference/model_fitter.py
```diff
@@ -0,0 +1,478 @@
...
+        deaths_stdev[deaths_stdev == 0] = 1e10
+        if hosp_stdev is not None:
+            hosp_stdev[hosp_stdev == 0] = 1e10
+
+        return cases_stdev, hosp_stdev, deaths_stdev
+
+    def run_model(self, R0, eps, t_break, log10_I_initial):
+        """
+        Generate the model.
+
```
zhang-xinyu, line 190: missing parameters description

pyseir/inference/model_fitter.py
```diff
@@ -0,0 +1,478 @@
...
+        #     suppression_policy = \
+        #         suppression_policies.generate_empirical_distancing_policy(
+        #             fips=fips, future_suppression=eps, **suppression_policy_params)
+        # elif by == 'state':
+        #     # TODO: This takes > 200ms which is 10x the model run time. Can be optimized...
+        #     suppression_policy = \
+        #         suppression_policies.generate_empirical_distancing_policy_by_state(
+        #             state=state, future_suppression=eps, **suppression_policy_params)
+        suppression_policy = suppression_policies.generate_two_step_policy(self.t_list, eps, t_break)
+        self.SEIR_kwargs['E_initial'] = 1.2 * 10**log10_I_initial
```
zhang-xinyu, line 205: why is it 1.2? better put comment here ...

pyseir/models/suppression_policies.py
```diff
@@ -157,6 +157,31 @@ def get_interventions(start_date=datetime.now().date()):
     return interp1d(t_list, rho, fill_value='extrapolate')
 
 
+def generate_two_step_policy(t_list, eps, t_break):
```
zhang-xinyu, line 160: How this is used as opposed to `generate_empirical_distancing_policy`? Is the latter ever used for inference?

pyseir/load_data.py
erccarls: The problem is some states have current (will rename to occupancy) and some have cumulative. We don't know which until reading it.

erccarls: Added an enum

pyseir/load_data.py
erccarls: Added an enum

pyseir/load_data.py
erccarls: Errors

erccarls: Data is super messy. This doesn't occur for more than a few states though and is short lived. I will add a comment.

pyseir/models/suppression_policies.py
erccarls: Why not? it is a pandas dataframe so order is fixed. This method is much faster by 200ms (pandas is very slow for iteration) which when using MCMC or inference

erccarls: BTW, recommend using numpy as much as possible for things that get run in the inner optimization loops.

pyseir/inference/model_fitter.py
erccarls: I think they better here actually for two reasons: 
1. Some are programatic, like `fix_hosp_fraction`
2. These are physical constraints, and hence are not expected to change.

pyseir/models/suppression_policies.py
erccarls: Since we were having issues with that one, I switched us for right now to use the two step model which fits much better. I think we need to explore data quality issues on that side.

Issue #207, opened 2020-04-12
[Top 100 Counties] Deploy the counties to a public directory for the front end to consume


goldblatt:
Just get this lil script hooked into the deploy script

PR #208, opened 2020-04-13
Updating the run.sh to deploy first 100 counties.json

goldblatt:
addresses #207

This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/207

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/pipelines/top_counties_pipeline.py
+++ b/libs/pipelines/top_counties_pipeline.py
@@ -13,10 +13,10 @@ logger = logging.getLogger(__name__)
 PROD_BUCKET = "data.covidactnow.org"
 
 TopCountiesPipelineProjectionResult = namedtuple(
-    "TopCountiesPipelineProjectionResult", ["key", "projection_df",]
+    "TopCountiesPipelineProjectionResult", ["projection_df",]
 )
 
-TopCountiesPipelineResult = namedtuple("TopCountiesPipelineResult", ["key", "api"])
+TopCountiesPipelineResult = namedtuple("TopCountiesPipelineResult", ["api"])
 
 
 def run_projections(
@@ -43,7 +43,7 @@ def run_projections(
     if run_validation:
         validate_results.validate_counties_df(counties_key_name, counties_df)
 
-    county_results = TopCountiesPipelineProjectionResult(counties_key_name, counties_df)
+    county_results = TopCountiesPipelineProjectionResult(counties_df)
 
     return county_results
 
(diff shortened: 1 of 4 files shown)


libs/pipelines/top_counties_pipeline.py
```diff
@@ -13,10 +13,10 @@
 PROD_BUCKET = "data.covidactnow.org"
 
 TopCountiesPipelineProjectionResult = namedtuple(
-    "TopCountiesPipelineProjectionResult", ["key", "projection_df",]
+    "TopCountiesPipelineProjectionResult", ["projection_df",]
```
mikelehen, line 16: 1. Is the trailing `,` after "projection_df" intentional?
2. I'm not familiar with this code, but creating a tuple with one element feels non-tuple-y.  Can you verify this is intentional / what we want? 😄

fridiculous:
Can you update https://github.com/covid-projections/covid-data-model/blob/master/api/README.md#api to state that there's a 7th file being created and what the schema is of that 7th file?

run.sh
```diff
@@ -42,6 +42,8 @@ prepare () {
   COUNTY_SUMMARIES_DIR="${API_OUTPUT_DIR}/county_summaries";
   CASE_SUMMARIES_DIR="${API_OUTPUT_DIR}/case_summary"
 
+  CAN_PREDICTIONS_DIR="${API_OUTPUT_DIR}/can_predictions/";
```
mikelehen, line 45: Where did this name come from?  I don't see this first_100_counties API described in https://www.dropbox.com/scl/fi/o4bec2kz8dkcxdtabtqda/CAN-V1-API-Draft.paper?dl=0&eid=&oref=e&rlkey=f7elx3zmo6tt5s7mj5nvap9a0&t=278234600567106294963247#:t=278234600567106294963247 at all, so I'm not sure what I should expect it to look like or where I should expect it to be.

mikelehen: Note that all of the existing directories (county_summaries, case_summaries) are legacy names that the website currently relies on.  Going forward we should converge on the layout / naming in the API draft doc.

libs/pipelines/top_counties_pipeline.py
goldblatt: ya following the pattern of the other pipeline files. I'd rather be consistent here but i aggree it doesn't really make sense?

goldblatt: also the formatting is from black

run.sh
goldblatt: i don't really have a strong opinion on naming here, so happy to name it whatever. as for why this exists, there's #207 for it and the website wants it for the home page?

libs/pipelines/top_counties_pipeline.py
mikelehen: Okaydoke!

run.sh
mikelehen: I see.  If this is going to be legacy (and replaced with something defined in v1 API draft) then I don't care too much about the name either.  But FWIW I think everything we're producing are essentially "CAN predictions", right?  So maybe just output this file to the root `${API_OUTPUT_DIR}` or something?  🤷

goldblatt: cool i updated to the county directory? i think that makes sense?

Issue #206, opened 2020-04-12
[Top 100 Counties] Return Dates in same format as other front end code


goldblatt:
Currently dates are returned "peakDate": "2020-05-15T00:00:00" for example, but the website expects them to be formatted differently

goldblatt:
iso format like "peakDate": "2020-05-15T00:00:00", lastUpdated is some other format!

PR #213, opened 2020-04-13
formatting the updated date to be iso format

goldblatt:
This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/206

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/cli/run_top_counties_dataset.py
+++ b/cli/run_top_counties_dataset.py
@@ -10,9 +10,9 @@ PROD_BUCKET = "data.covidactnow.org"
 
 @click.command("deploy-top-counties")
 @click.option(
-    "--run_validation",
-    "-r",
-    default=True,
+    "--disable_validation",
+    "-dv",
+    is_flag=True,
     help="Run the validation on the deploy command",
 )
 @click.option(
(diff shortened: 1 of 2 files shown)


cli/run_top_counties_dataset.py
```diff
@@ -10,9 +10,9 @@
 
 @click.command("deploy-top-counties")
 @click.option(
-    "--run_validation",
-    "-r",
-    default=True,
+    "--disable_validation",
```
ghop02, line 13: most of the others are `--disable-validation` (click converts the name to th underscore in the function calls)

libs/functions/generate_api.py
```diff
@@ -7,14 +8,24 @@
 from libs.datasets import results_schema as rc
 
 
+def _format_date(input_date):
+    if not input_date: 
+        raise Exception("Can't format a date that doesn't exist")
+    if isinstance(input_date, str): 
+        # note if this is already in iso format it will be grumpy. maybe use dateutil
+        datetime_obj = datetime.strptime(input_date, '%m/%d/%Y %H:%M')
```
ghop02, line 16: Where does this format type come from? would be great to make everything iso earlier in the flow.

there's also this new function that might be helpful: https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat

cli/run_top_counties_dataset.py
goldblatt: coool done

libs/functions/generate_api.py
goldblatt: ya it's in the JHU dataset we can format to ISO there although I think that PR is in flight @ghop02 wanna fix it in your PR?

ghop02: Yeah sounds good!

PR #215, opened 2020-04-13
Model/add priors

erccarls:
Add strong priors on epsilon (by proxy R_effective)  to lower bound at 0.98.  This is needed to stabilize the fits in the days of early data.

We also 
1. remove territories until county metadata can be added for them.
2. Increase the weight of case data. This can likely be reduced later but is a critical leading indicator at the moment.
3. Make the transition period for the two step suppression policy 1 week instead of two as we suspected bias after reviewing states.


pyseir/inference/model_fitter.py
```diff
@@ -217,21 +232,24 @@ def run_model(self, R0, eps, t_break, log10_I_initial):
         #     suppression_policy = \
         #         suppression_policies.generate_empirical_distancing_policy_by_state(
         #             state=state, future_suppression=eps, **suppression_policy_params)
-        suppression_policy = suppression_policies.generate_two_step_policy(self.t_list, eps, t_break)
+        suppression_policy = suppression_policies.generate_two_step_policy(
+            self.t_list, eps, t_break)
 
         # Load up some number of initial exposed so the initial flow into infected is stable.
         steady_state_exposed_to_infected_ratio = 1.2
```
zhang-xinyu, line 239: Would better make this global varaible.

pyseir/inference/model_fitter.py
```diff
@@ -308,52 +339,105 @@ def _fit_seir(self, R0, t0, eps, t_break, test_fraction, hosp_fraction, log10_I_
...
+        prior, and invert this transform to get back to the epsilon Max
+        A-Posteriori (MAP) estimate.
+
+        Returns
+        -------
+        posterior_map_estimate: float
+            Max A-Posteriori (MAP) estimate for epsilon.
+        """
+        R_eff = R0 * eps
+        R_eff_stdev = R0 * eps_error
```
zhang-xinyu, line 356: This seems to miss the error of R0.

pyseir/inference/model_fitter.py
```diff
@@ -308,52 +339,105 @@ def _fit_seir(self, R0, t0, eps, t_break, test_fraction, hosp_fraction, log10_I_
...
+
+        x = np.linspace(0.00, 10, 1001)
+        delta_x = x[1] - x[0]
+
+        # This implements a hard lower limit of 0.98.
+        # TODO: As more data comes in, relax this.. Probably just use MCMC..
+        prior = gamma.pdf((x - 0.98) / .5, 1.3)
```
zhang-xinyu, line 363: If R eff change over time and we infer it by R0 and eps, why would we need a prior for it?

pyseir/inference/model_fitter.py
```diff
@@ -45,12 +47,13 @@ class ModelFitter:
...
         the influence of hospitalization data on the fit.
     """
+
     def __init__(self,
                  fips,
                  ref_date=datetime(year=2020, month=1, day=1),
                  min_deaths=2,
                  n_years=1,
-                 cases_to_deaths_err_factor=1,
+                 cases_to_deaths_err_factor=.333,
```
zhang-xinyu, line 56: How is this determined? Did you test different ratios and find best fit at this ratio?

pyseir/inference/model_fitter.py
erccarls: That is correct.  We need a better formalism to get the joint posteriors.

pyseir/inference/model_fitter.py
erccarls: This is essentially a prior on epsilon, but since we want to define the prior based on observations of R_eff from other sources we use this procedure.

pyseir/inference/model_fitter.py
erccarls: Yes. Later we will be able to reduce this, but for now we are seeing that death data is noisy for so many states (like 5 deaths) that we need more weight in the leading indicators like cases to effectively calculate R0.

zhang-xinyu: Cool! Thanks for the clarification!

pyseir/inference/model_fitter.py
zhang-xinyu: Ok, should be fine this is temporary solution, and agreed better formalism is needed.

pyseir/inference/model_fitter.py
zhang-xinyu: Got it.

pyseir/inference/model_fitter.py
zhang-xinyu: Would better to put add a single grid search at state level for the best ratio as we have more death data surfaced.

erccarls: I am not sure what the optimization objective would be. Goodness of fit doesn't make sense I don't think.

Issue #126, opened 2020-04-06
Write scheduled GitHub Action to run run.sh and deploy to S3.


mikelehen:
The goal is to create a scheduled GitHub Action that re-runs our models to generate new API artifacts, and publishes them on data.covidactnow.org ~once daily.

It will need to:
1. clone the covid-data-public repo (will git lfs be an issue?)
2. run `run.sh` (see https://github.com/covid-projections/covid-data-model/issues/123).
3. Publish to S3 according to URL format / version schema specified in https://docs.google.com/document/d/1VhQqMxtVYHQMIj9zFeb1DUQZ1aq1k4xIc_bYoTIqUb4/edit

Notes:
* There's some s3 python code [here](https://github.com/covid-projections/covid-data-model/blob/f1b790585fadf19d34c5176440ca895cb59dfaee/deploy_dod_dataset.py#L8) that might be useful. Or maybe there's a command-line tool that would be easier / better.
* We need to generate a snapshot-id to stick in the ultimate S3 destination.  This could be a precise timestamp or maybe github actions have a unique job ID we could use, or something.

mikelehen:
@goldblatt pointed out that we should use multi-part upload (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#multipartupload) since the shape files are large and can time out on slow internet connections.

fridiculous:
Also, please note there is a placeholder for this job here: https://github.com/covid-projections/covid-data-model/blob/master/.github/workflows/schedule.yml

kpinc:
I am not clear if this is the right place, but you may want to re-open this issue.  If you look at a state/county at covidactnow.org, the last-updated date is April 6.  This leads me to believe that the data hasn't been pushed since then, which is about when this task was started.

fridiculous:
yes @kpinc - thanks for noting.  the larger goal of model integration has yet to be streamlined to the site.

mikelehen:
This is done.  We're pushing new API snapshots to S3 twice a day, though it's still manual to get them onto the website (and that's slightly blocked by model integration work).

PR #216, opened 2020-04-13
Add workflow for labelling snapshots and tools/label-api.sh helper for invoking.

mikelehen:
Add Github Action Workflow for labeling snapshots.
* Triggers off of a repository dispatch with event type 'label-api-snapshot'.
* Requires 'label' and 'snapshot_id' in the payload.
* For now just copies the files, but we'd like a symlink or similar in the future.

Also added `tools/publish-api.sh` and `tools/label-api.sh` helpers for invoking
workflows manually, and added to README.md.

This PR helps address issue #126.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/.github/workflows/label_api_snapshot.yml
+++ b/.github/workflows/label_api_snapshot.yml
@@ -1,0 +1,44 @@
+# To debug, it's recommended you modify / use the version in the test-actions repo:
+# https://github.com/covid-projections/test-actions/blob/master/.github/workflows/label_api_snapshot.yml
+
+# Used to assign a label (like /snapshot/latest or /v0) to a published API
+# snapshot (like /snapshot/123).
+#
+name: Label API Snapshot.
+
+on:
+  # Hook to trigger a manual run.
+  # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
+  repository_dispatch:
+    types: label-api-snapshot
+
+jobs:
+  label-api-snapshot:
+    runs-on: ubuntu-latest
+
+    env:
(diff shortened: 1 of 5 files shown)


.github/workflows/label_api_snapshot.yml
```diff
@@ -0,0 +1,44 @@
...
+      uses: actions/checkout@v2
+    - name: Verify Label provided
+      if: ${{ !env.LABEL }}
+      run: 'echo "Missing client_payload parameter: label" ; exit 1'
+    - name: Verify Snapshot ID provided
+      if: ${{ !env.SNAPSHOT_ID }}
+      run: 'echo "Missing client_payload parameter: snapshot_id" ; exit 1'
+
+    # TODO: We want to replace this with a "symlink" of some kind, perhaps implemented at
+    # the CloudFront layer.
```
mikelehen, line 35: This is where we'd plug in a smarter symlink mechanism.

fridiculous:
Only comment is really about terminology - more of a question.

Would it make sense to change the the verbage from `publish-api` to `push-api`.  And similarly, once we are actually publishing with a label, since it's intended to happen after a manual qa step, should we do change `label-api` to `publish-api` where publish api comes with a default label?

mikelehen:
Renamed `publish-api.sh` to `push-api.sh`.

I left `label-api.sh` alone for now.  Perhaps once we've settled on what it means to publish an API (is it `v0` or `latest` or `stable` or whatever) and remove the label argument, I'm down with renaming it.

Issue #212, opened 2020-04-13
[API] Deploy the API data to the bucket via run.sh


goldblatt:
Something similar to https://github.com/covid-projections/covid-data-model/pull/208)

goldblatt:
this is probably just going to wind up solved with #211

igorkofman:
I think this is done!

Issue #211, opened 2020-04-13
[API] Make an api pipeline that reads in the data, processes it and converts it to the API json


goldblatt:
Make an API pipeline for the counties/states should follow something like (https://github.com/covid-projections/covid-data-model/blob/master/libs/pipelines/top_counties_pipeline.py)

PR #217, opened 2020-04-13
Generate API Files for counties/states & Upload to S3

goldblatt:
This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/211 & https://github.com/covid-projections/covid-data-model/issues/212

### Please Check
- [x] Are tests passing?

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -20,12 +20,14 @@ We currently provide 6 file types:
 │   │   ├── <state abreviation>.summary.json
 │   │   └── case_summary.version.json
 │   ├── county/
-|   |   ├── counties_top_100.json
 │   │   ├── <state abreviation>.<fips>.<intervention number>.json
-│   │   ├── <fips>.<intervention number>.json
 │   │   └── county.version.json
-│   ├── state/
-│   │   ├── <state abreviation>.<intervention number>.json
+│   ├── us/
+|   │   ├── counties/
+|   │   │   ├── <5 Digit FIPS>.<intervention number>.json
+|   │   │   └── counties_top_100.json
+|   │   └── states/
+|   │       └── <state abbreviation>.<intervention number>.json
 │   ├── county_summaries/
 │   │   ├── <state abreviation>.summary.json
 │   │   └── county_summary.version.json
@@ -172,9 +174,9 @@ For `county_summaries/<state abreviation>.summary.json` files see the [JSON Sche
 For `county/counties_top_100.json` fields see the [JSON Schema](schema/CANPredictionAPI.json)
 
 ### For State/County Calculated Results
-For fields like 
-- `county/<fips>.<intervention>.json` 
-- `state/<state abbreviation>.<intervention>.json` 
+For files like 
+- `us/counties/<fips>.<intervention>.json` 
+- `us/states/<state abbreviation>.<intervention>.json` 
 
 see the [JSON Schema](schema/CANPredictionAPIRow.json)
 
--- a/api/README.md
+++ b/api/README.md
@@ -22,7 +22,10 @@ We currently provide 6 file types:
 │   ├── county/
 |   |   ├── counties_top_100.json
 │   │   ├── <state abreviation>.<fips>.<intervention number>.json
+│   │   ├── <fips>.<intervention number>.json
 │   │   └── county.version.json
+│   ├── state/
+│   │   ├── <state abreviation>.<intervention number>.json
 │   ├── county_summaries/
 │   │   ├── <state abreviation>.summary.json
 │   │   └── county_summary.version.json
@@ -167,6 +170,13 @@ For `county_summaries/<state abreviation>.summary.json` files see the [JSON Sche
 
 ### Top Counties
 For `county/counties_top_100.json` fields see the [JSON Schema](schema/CANPredictionAPI.json)
+
+### For State/County Calculated Results
+For fields like 
+- `county/<fips>.<intervention>.json` 
+- `state/<state abbreviation>.<intervention>.json` 
+
+see the [JSON Schema](schema/CANPredictionAPIRow.json)
 
 ### Versions
 
--- a/libs/enums.py
+++ b/libs/enums.py
@@ -5,11 +5,6 @@ import enum
 UNKNOWN_FIPS = "99999"
 
 
-class AggregateLevel(enum.Enum):
-    STATE = 0
-    COUNTY = 1
-
-
 class Intervention(enum.Enum):
     NO_INTERVENTION = 0
     FLATTEN = 1
--- a/libs/enums.py
+++ b/libs/enums.py
@@ -4,6 +4,10 @@ import enum
 # TODO: This should maybe be unique per state.
 UNKNOWN_FIPS = "99999"
 
+class AggregateLevel(enum.Enum):
+    STATE = 0 
+    COUNTY = 1
+
 class Intervention(enum.Enum):
     NO_INTERVENTION = 0
     FLATTEN = 1
--- a/libs/enums.py
+++ b/libs/enums.py
@@ -4,24 +4,26 @@ import enum
 # TODO: This should maybe be unique per state.
 UNKNOWN_FIPS = "99999"
 
+
 class AggregateLevel(enum.Enum):
-    STATE = 0 
+    STATE = 0
     COUNTY = 1
+
 
 class Intervention(enum.Enum):
     NO_INTERVENTION = 0
     FLATTEN = 1
     FULL_CONTAINMENT = 2
     SOCIAL_DISTANCING = 3
-    CURRENT = 4 # look at what the state is and get the file for that
+    CURRENT = 4  # look at what the state is and get the file for that
 
     @classmethod
     def from_str(cls, label):
-        if label == "shelter_in_place": 
+        if label == "shelter_in_place":
             return cls.FLATTEN
-        elif label == "social_distancing": 
+        elif label == "social_distancing":
             return cls.SOCIAL_DISTANCING
-        elif label == "lockdown": 
+        elif label == "lockdown":
             return cls.FULL_CONTAINMENT
-        else: 
+        else:
             return cls.NO_INTERVENTION
--- a/libs/pipelines/api_pipeline.py
+++ b/libs/pipelines/api_pipeline.py
@@ -1,7 +1,8 @@
 from collections import namedtuple
 import logging
 
-from libs.enums import Intervention, AggregateLevel
+from libs.enums import Intervention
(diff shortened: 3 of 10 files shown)


libs/pipelines/api_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+APIPipelineAPIRow = namedtuple("APIPipelineAPIRow", ["key", "api"])
+
+APIPipelineResult = namedtuple("APIPipelineResult", ["api_tuples"])
+
+
+def run_projections(
+    input_file, aggregation_level, intervention: Intervention, run_validation=True
+) -> APIPipelineProjectionResult:
+    """Run the projections for the given intervention for states 
+    in order to genereate the api.
```
ghop02, line 29: typo in generate

libs/pipelines/api_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+        if run_validation:
+            validate_results.validate_counties_df(counties_key_name, counties_df)
+
+        county_results = APIPipelineProjectionResult(
+            intervention, AggregateLevel.COUNTY, counties_df
+        )
+
+        return county_results
+    else:
+        raise Exception("Non-valid aggreation level specified")
```
ghop02, line 67: I prefer `ValueError` errors here - just a bit more specific (also typo in aggregation)

libs/pipelines/api_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+logger = logging.getLogger(__name__)
+PROD_BUCKET = "data.covidactnow.org"
+
+APIPipelineProjectionResult = namedtuple(
+    "APIPipelineProjectionResult", ["intervention", "aggregate_level", "projection_df",]
+)
+
+APIPipelineAPIRow = namedtuple("APIPipelineAPIRow", ["key", "api"])
+
+APIPipelineResult = namedtuple("APIPipelineResult", ["api_tuples"])
```
ghop02, line 22: How do you feel about these?  I think these classes could be dataclasses to add some type inference.  don't have a great idea what api_tuples contains from just reading this

libs/enums.py
```diff
@@ -4,20 +4,26 @@
 # TODO: This should maybe be unique per state.
 UNKNOWN_FIPS = "99999"
 
+
+class AggregateLevel(enum.Enum):
+    STATE = 0
+    COUNTY = 1
```
ghop02, line 10: Hmm is there a reason that we should have both `AggregationLevel` (see dataset/dataset_utils.py) and `AggregateLevel`? It'd be great to combine those two.

run.sh
```diff
@@ -35,10 +35,12 @@ prepare () {
   # These directiories essentially define the structure of our API endpoints.
   # TODO: These should perhaps live in python, near the schemas (defined in api/)?
   STATES_DIR="${API_OUTPUT_DIR}/";
+  INPUT_BASE_DIR="${API_OUTPUT_DIR}/";
   # TODO: I think deploy_dod_dataset.py may currently have an implicit
   # requirement that the county model JSON is in a /county subdirectory of the
   # states?
   COUNTIES_DIR="${API_OUTPUT_DIR}/county";
+  STATES_OUTPUT_DIR="${API_OUTPUT_DIR}/state";
```
mikelehen, line 43: 1. Should this be `${API_OUTPUT_DIR}/us/states/` to match https://www.dropbox.com/scl/fi/o4bec2kz8dkcxdtabtqda/CAN-V1-API-Draft.paper?dl=0&eid=&oref=e&rlkey=f7elx3zmo6tt5s7mj5nvap9a0&t=278234600567106294963247#:t=278234600567106294963247 ?
2. STATES_DIR vs STATES_OUTPUT_DIR is confusing.  Maybe rename this to `V1_STATES_DIR`?
3. `COUNTIES_DIR` is the legacy directory. Perhaps we should create `V1_COUNTIES_DIR` that points at `us/counties` and move the top-100 counties list to there?

run.sh
```diff
@@ -35,10 +35,12 @@ prepare () {
   # These directiories essentially define the structure of our API endpoints.
   # TODO: These should perhaps live in python, near the schemas (defined in api/)?
   STATES_DIR="${API_OUTPUT_DIR}/";
+  INPUT_BASE_DIR="${API_OUTPUT_DIR}/";
```
mikelehen, line 38: We have API_OUTPUT_DIR, STATES_DIR, and INPUT_BASE_DIR that all point at the same thing, which seems confusing.

I actually screwed up below (line 57) and used the wrong one.

I vote we don't add another one and you just use API_OUTPUT_DIR or STATES_DIR.

run.sh
```diff
@@ -73,8 +75,16 @@ execute() {
   echo ">>> Generating ${API_OUTPUT_DIR}/version.json"
   generate_version_json
 
-  echo ">>> Generating Top 100 Counties json to ${COUNTIES_DIR}/first_100_counties.json"
-  ./run.py deploy-top-counties -i "${STATES_DIR}" -o "${COUNTIES_DIR}"
+  echo ">>> Generating Top 100 Counties json to ${COUNTIES_DIR}/counties_top_100.json"
+  mkdir -p "${COUNTIES_OUTPUT_DIR}"
```
mikelehen, line 79: COUNTIES_OUTPUT_DIR is undefined.

libs/pipelines/api_pipeline.py
goldblatt: ya i truly was at a loss for how to name these

run.sh
goldblatt: ya i really struggled with this file LOL let me see i think that the API docs had state and county not the plurals

libs/enums.py
goldblatt: ahh cool thought they were class vars.

libs/pipelines/api_pipeline.py
goldblatt: i think classes would bloat them out tbh?

goldblatt: normally it's like if you can't think of good names do they need to be named?

goldblatt: i renamed them a bit and i'm gonna ack this for the time being

run.sh
goldblatt: I guess overall i found it a bit confusing that the input and output directories were linked? i was trying to figure out a way to separate them

mikelehen: FWIW- This is why I wanted to generate the new schema JSON files directly from the pandas dataframes or whatever instead of going through this intermediate directory (which is both the output directory for the legacy schema and the input directory for this new stuff that you're adding).  I don't have a great suggestion for fixing the naming other than hopefully we get rid of the intermediate directory eventually, so we maybe don't need to worry about it too much.

goldblatt: i changed it around a lilttle bit in my pr lmk what you think

libs/pipelines/api_pipeline.py
```diff
@@ -17,27 +18,35 @@
     "APIPipelineProjectionResult", ["intervention", "aggregate_level", "projection_df",]
 )
 
-APIPipelineAPIRow = namedtuple("APIPipelineAPIRow", ["key", "api"])
+APIGenrationRow = namedtuple("APIGenrationRow", ["key", "api"])
```
ghop02, line 21: typo :)

goldblatt: ahh yes my spelling is v bad >.<

run.sh
```diff
@@ -34,11 +34,15 @@ prepare () {
 
   # These directiories essentially define the structure of our API endpoints.
   # TODO: These should perhaps live in python, near the schemas (defined in api/)?
-  STATES_DIR="${API_OUTPUT_DIR}/";
+
+  INPUT_BASE_DIR="${API_OUTPUT_DIR}/";
+
+  API_OUTPUT_COUNTIES = "${API_OUTPUT_DIR}/counties"
+  API_OUTPUT_STATES = "${API_OUTPUT_DIR}/states"
```
mikelehen, line 41: So to match the v1 API spec, this should still have a `us/` in the path, no? https://www.dropbox.com/scl/fi/o4bec2kz8dkcxdtabtqda/CAN-V1-API-Draft.paper

run.sh
```diff
@@ -50,13 +54,13 @@ execute() {
   # Go to repo root (where run.sh lives).
   cd "$(dirname "$0")"
 
-  echo ">>> Generating state models to ${STATES_DIR}"
+  echo ">>> Generating state models to ${API_OUTPUT_DIR}"
   # TODO(#148): We need to clean up the output of these scripts!
   ./run.py model state -o "${API_OUTPUT_DIR}" > /dev/null
 
-  echo ">>> Generating county models to ${COUNTIES_DIR}"
+  echo ">>> Generating county models to ${API_OUTPUT_DIR}"
```
mikelehen, line 61: should be `${API_OUTPUT_DIR}/county` I think?

api/README.md
```diff
@@ -22,7 +22,10 @@ We currently provide 6 file types:
 │   ├── county/
 |   |   ├── counties_top_100.json
 │   │   ├── <state abreviation>.<fips>.<intervention number>.json
+│   │   ├── <fips>.<intervention number>.json
```
mikelehen, line 25: Can you update these changes to match the new directory structure (should be `us |- counties |- <fips>.<intervention>.json` or similar)?

api/README.md
```diff
@@ -22,7 +22,10 @@ We currently provide 6 file types:
...
 │   │   └── county.version.json
+│   ├── state/
+│   │   ├── <state abreviation>.<intervention number>.json
```
mikelehen, line 28: update please?

api/README.md
```diff
@@ -168,6 +171,13 @@ For `county_summaries/<state abreviation>.summary.json` files see the [JSON Sche
 ### Top Counties
 For `county/counties_top_100.json` fields see the [JSON Schema](schema/CANPredictionAPI.json)
 
+### For State/County Calculated Results
+For fields like
```
mikelehen, line 175: Should "fields" be "files" here?

api/README.md
```diff
@@ -168,6 +171,13 @@ For `county_summaries/<state abreviation>.summary.json` files see the [JSON Sche
 ### Top Counties
 For `county/counties_top_100.json` fields see the [JSON Schema](schema/CANPredictionAPI.json)
 
+### For State/County Calculated Results
+For fields like 
+- `county/<fips>.<intervention>.json` 
+- `state/<state abbreviation>.<intervention>.json`
```
mikelehen, line 177: Update to match the new directories please. 😄

PR #219, opened 2020-04-14
API declarations

igorkofman:

### Please Check
- [ ] Are tests passing?


api/can_api_definition.py
```diff
@@ -0,0 +1,82 @@
...
+
+"""
+Documented https://www.dropbox.com/scl/fi/o4bec2kz8dkcxdtabtqda/CAN-V1-API-Draft.paper?dl=0&rlkey=f7elx3zmo6tt5s7mj5nvap9a0
+
+"""
+
+class _ResourceUsageProjection(pydantic.BaseModel):
+    peakShortfall: int = pydantic.Field(
+        ..., description="Shortfall of resource needed at the peek utilization"
+    )
```
goldblatt, line 13: this needs to be optional as it might be none

api/can_api_definition.py
```diff
@@ -0,0 +1,82 @@
...
+    peakDate: datetime.date = pydantic.Field(
+        ..., description="Date of peak resource utilization"
+    )
```
goldblatt, line 16: this also needs to be optional

api/can_api_definition.py
```diff
@@ -0,0 +1,82 @@
...
+    shortageStartDate: Optional[datetime.date] = pydantic.Field(
+        ..., description="Date when resource shortage begins"
+    )
+
+class _Projections(pydantic.BaseModel):
```
goldblatt, line 21: also think specififying the projection window would be helpful

libs/functions/generate_api.py
```diff
@@ -22,20 +22,27 @@ def generate_api_for_projection(projection):
...
-        county_result = CANPredictionAPIRow(
+        county_result = CovidActNowCountySummary(
+            lat=county_row[rc.LATITUDE],
+            long=county_row[rc.LONGITUDE],
+            actuals=None,
             stateName=county_row[rc.STATE],
             countyName=county_row[rc.COUNTY],
             fips=county_row[rc.FIPS],
-            lastUpdatedDate=county_row[rc.LAST_UPDATED],
+            lastUpdatedDate=_get_date_or_none(county_row[rc.LAST_UPDATED]),
```
goldblatt, line 44: can this be empty?

api/can_api_definition.py
igorkofman: alternatively it could be 0?

api/can_api_definition.py
igorkofman: agreed

libs/functions/generate_api.py
igorkofman: it shouldn't be

api/can_api_definition.py
goldblatt: i think we pass in a <null> for this so we could either make it 0 elsewhere or make it none

PR #224, opened 2020-04-14
Sensitivity Analysis

rjchacko:
This PR has a new module called `outcomes` with a class `OutcomesSampler` that runs many rollouts of the SEIR model at a random sample of points in a parameter space. It computes summary statistics for those rollouts and accumulates them into a DataFrame along with the parameters that were sampled. A Gaussian process is fit to the samples to get a function that can be easily evaluated on different slices of the parameter space.

There's a new `notebooks` folder with a notebook with a first pass at doing a sensitivity analysis and a `models` folder to persist samples and models as experiments are being run. 

Questions:

1. What should the parameter space be? Right now I'm varying `R0`, `delta`, and `suppression_policy`.
2. What are some outcomes I should look at?

### Please Check
- [x] Are tests passing?


erccarls:
@rjchacko I think this looks great and will provide a useful tool as we e.g. add new suppression release policies, we can start to study the delta in expected cumulative deaths, hospitalizations, and immunity under different variations of policy.

If you could run through and add some brief docstrings that would be super helpful to another that might pick this up in a week or two. 

If we can also get a basic econ model locked in and integrated, it seems like this would be the perfect tool to obtain economic and health outcome responses to policy variations, and produce a set of tradeoff curves.

rjchacko:
@erccarls okay added some docstrings and did some assorted clean up and tests are passing. One other improvement I could make is to refactor `OutcomeModels` to implement the sklearn `Estimator` API, but I figure that could go into another PR.

PR #226, opened 2020-04-15
Model/add priors

erccarls:

### Please Check
- [ ] Are tests passing?


pyseir/cli.py
```diff
@@ -106,8 +109,9 @@ def _run_all(state=None, run_mode='default', generate_reports=True, output_inter
         cache_all_data()
 
     if state:
-        if not states_only:
-            _impute_start_dates(state.title())
+        # Deprecate temporarily since not needed.
```
ghop02: Why is this deprecated? Would be helpful for people like me with a bit more context

pyseir/cli.py
```diff
@@ -165,18 +170,18 @@ def generate_state_report(state):
...
+@click.option('--run-mode', default=DEFAULT_RUN_MODE, type=str, help='State to generate files for. If no state is given, all states are computed.')
 @click.option('--states-only', default=False, is_flag=True, type=bool, help='Only model states')
 def map_outputs(state, output_interval_days, run_mode, states_only):
     _map_outputs(state, output_interval_days=int(output_interval_days), run_mode=run_mode, states_only=states_only)
 
 
 @entry_point.command()
 @click.option('--state', default=None, help='State to generate files for. If no state is given, all states are computed.')
-@click.option('--run-mode', default='default',type=str, help='State to generate files for. If no state is given, all states are computed.')
+@click.option('--run-mode', default=DEFAULT_RUN_MODE,type=str, help='State to generate files for. If no state is given, all states are computed.')
```
ghop02, line 197: Are these an enum? anyways, to make this more explicit you can add a: 
```
type=click.Choices([run_mode.value for run_mode en RunModeEnum])
```

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -34,13 +34,13 @@ def __init__(self, state, output_interval_days=4, run_mode='can-before',
...
-        self.state_output_dir = os.path.join(output_dir, 'web_ui', 'state')
+        # output_dir = output_dir or OUTPUT_DIR
+        # self.county_output_dir = os.path.join(output_dir, 'web_ui', 'county')
+        # self.state_output_dir = os.path.join(output_dir, 'web_ui', 'state')
         self.state_abbreviation = us.states.lookup(state).abbr
 
-        os.makedirs(self.county_output_dir, exist_ok=True)
-        os.makedirs(self.state_output_dir, exist_ok=True)
+        # os.makedirs(self.county_output_dir, exist_ok=True)
+        # os.makedirs(self.state_output_dir, exist_ok=True)
```
ghop02, line 43: why are these commented out?

pyseir/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    state_obj = us.states.lookup(fips[:2])
+    if len(fips) == 5:
+        agg_level = AggregationLevel.COUNTY
+        county = load_data.load_county_metadata_by_fips(fips)['county']
+    else:
+        agg_level = AggregationLevel.STATE
+
+    artifact = RunArtifact(artifact)
+
+    if artifact is RunArtifact.MLE_FIT_REPORT:
```
ghop02, line 52: Is there a reason you did `is` here rather than `==`? would prefer `==` since is has some weird gotchas

pyseir/models/suppression_policies.py
```diff
@@ -120,11 +120,11 @@ def get_interventions(start_date=datetime.now().date()):
...
                 rho.append(1)
             elif (actual_date - today).days <= 30:
-                rho.append(1.1 / R0)
+                rho.append(0.97 / R0)
             elif (actual_date - today).days <= 60:
-                rho.append(1.0 / R0)
+                rho.append(0.97 / R0)
             elif (actual_date - today).days <= 90:
-                rho.append(0.8 / R0)
+                rho.append(0.97 / R0)
```
ghop02, line 127: Seems like this should be pulled out into a variable?

pyseir/ensembles/ensemble_runner.py
```diff
@@ -235,7 +228,7 @@ def init_run_mode(self):
                 self.override_params['gamma'] = 1   # 100% of Exposed go to the infected bucket.
 
                 # 1.2 is a ~ steady state for the exposed bucket initialization.
```
ghop02, line 230: 1.2 -> 0.6

pyseir/models/suppression_policies.py
erccarls: We could pull it out into config, but it is soon to be deprecated and they are just set by hand more or less.. So I dunno that it is worth it.

pyseir/utils.py
erccarls: Enum compares are supposed to be `is` according to python 3 spec.
https://stackoverflow.com/questions/25858497/should-enum-instances-be-compared-by-identity-or-equality

pyseir/deployment/webui_data_adaptor_v1.py
erccarls: removed in last push thanks.

pyseir/cli.py
erccarls: Thank you!

pyseir/utils.py
ghop02: 🙀 This is good to know and makes sense!

davidstrauss:
Rebuilt for merging to master in #238

Issue #210, opened 2020-04-13
[API] Extende generate_api.py to include new API fields


goldblatt:
The generate API takes the dataframe and turns it into pydantic fields. those fields then can be converted to json for the api. 

https://github.com/covid-projections/covid-data-model/commit/3b79b5ffe12d1fa55ccd95d5870e7ad13d40c0bf#diff-8afcc007867311c4fe2a6de49e14a334

PR #228, opened 2020-04-15
First pass at Adding the timeseries API

goldblatt:
This PR addresses issue #210 

### Please Check
- [] Will the *current* schema in [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.md) be updated with this PR?
- [?] Are tests passing?

### DIFF
--- a/libs/datasets/can_model_output_schema.py
+++ b/libs/datasets/can_model_output_schema.py
@@ -3,7 +3,7 @@ Ouput of the
 """
 DATE = "date"
 TOTAL = "total"
-SUSCEPTIBLE = "susceptible"
+TOTAL_SUSCEPTIBLE = "susceptible"
 EXPOSED = "exposed"
 INFECTED = "infected"
 INFECTED_A = "infected_a"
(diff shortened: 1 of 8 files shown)


libs/datasets/can_model_output_schema.py
```diff
@@ -1,29 +1,59 @@
...
+    DEAD,
+    BEDS,
     "i",
     "j",
     "k",
     "l",
-    "population",
+    POPULATION,
     "m",
     "n",
```
goldblatt, line 55: Generally confused about what these are? someone pointed me to a few places and i looked at the original file, but where lines 18-32 came from

libs/functions/generate_api.py
```diff
@@ -4,11 +4,18 @@
...
+    CANPredictionTimeseriesRow,
     _Projections,
     _ResourceUsageProjection,
 )
 from libs.datasets import results_schema as rc
 from libs.constants import NULL_VALUE
+from libs.functions import get_can_projection
+from libs.datasets.dataset_utils import AggregationLevel
+from libs.us_state_abbrev import US_STATE_ABBREV
+from libs.datasets import can_model_output_schema as can_scema
```
ghop02, line 18: can_scema -> can_schema

libs/datasets/can_model_output_schema.py
```diff
@@ -1,29 +1,59 @@
 """
 Ouput of the 
 """
+DATE = "date"
+TOTAL = "total"
+SUSCEPTIBLE = "susceptible"
+EXPOSED = "exposed"
+INFECTED = "infected"
+INFECTED_A = "infected_a"
```
ghop02, line 9: would be cool to name this closer to the comments below

libs/datasets/can_model_output_schema.py
igorkofman: they are just unused values from old spreadsheet

PR #240, opened 2020-04-15
adds a lil function to create the directory if it doesn't exist. Leav…

goldblatt:
…es a bit of directories, but i think they get deleted hopefully by the action cleanup.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/ensembles/ensemble_runner.py
+++ b/pyseir/ensembles/ensemble_runner.py
@@ -24,6 +24,14 @@ from libs.datasets import JHUDataset
 
 _logger = logging.getLogger(__name__)
 
+def _get_or_make_directory(*path): 
+    current_path_pieces = []
+    for path_piece in path: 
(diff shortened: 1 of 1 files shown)


pyseir/ensembles/ensemble_runner.py
```diff
@@ -24,6 +24,14 @@
 
 _logger = logging.getLogger(__name__)
 
+def _get_or_make_directory(*path): 
+    current_path_pieces = []
+    for path_piece in path: 
+        current_path = os.path.join(*current_path_pieces, path_piece)
+        if not os.path.exists(current_path): 
+            os.mkdir(current_path)
```
ghop02, line 32: I think you can just call `os.mkdirs(full_path, exists=True)` instead of needing to build this logic yourself?

ghop02: ah, it's `exist_ok`: https://docs.python.org/3/library/os.html#os.makedirs

Issue #230, opened 2020-04-15
Standardize Code Formatting


ghop02:
Now that most of the large epics are in master, i'd love to get an auto-formatter setup. 

Steps that have to happen:
 - [ ] run black on repo `make fmt` or `black .`
 - [ ] Add instructions for how to easily add black formatting as a pre-commit hook
 - [ ] Add linting to make sure that committed code is properly formatted: [Github action linter](https://github.com/marketplace/actions/black-code-formatter)

BrettBoval:
black now runs as a pre-commit hook

PR #242, opened 2020-04-16
Running code formatting

ghop02:
This just runs black on all the code in `libs/`. 

Addresses part of #230


fridiculous:
probably worth redoing after all outstanding branches get merged?  simply to avoid the conflicts?

PR #243, opened 2020-04-16
Model/optimize pyseir integration

erccarls:
Age structure dynamics add substantial matrix calculations to the SEIR model which produced a 50x runtime hit per simulation.  The additional variable tracking required also introduces expensive jacobian computation for the standard ODEINT solver. Switching to a Runge-Kutta method provides much faster integration.  Low level optimizations have also been made to interpolation, array insertion, and array splitting.

Deep optimization of age structured dynamics.  Takes runtime from about 500ms per sim -> 35ms per sim with no loss to accuracy for 18 age bins.

### DIFF
--- a/pyseir/models/seir_model_age.py
+++ b/pyseir/models/seir_model_age.py
@@ -1,5 +1,5 @@
 import numpy as np
-from scipy.integrate import odeint
+from scipy.integrate import odeint, solve_ivp
 import matplotlib.pyplot as plt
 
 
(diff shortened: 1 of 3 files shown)


erccarls:
no. This is a very smooth differential equation.

pyseir/models/seir_model_age.py
```diff
@@ -621,16 +626,23 @@ def run(self):
...
+                                       method='RK23', rtol=1e-3, atol=1e-3).y
 
-        S_fracs_within_age_group = S / S.sum(axis=0)
+        S, E, A, I, HGen, HICU, HICUVent = np.split(result_time_series[:-7], 7)
+        R, D, D_no_hgen, D_no_icu, HAdmissions_general, HAdmissions_ICU, TotalAllInfections = result_time_series[-7:]
 
-        if self.suppression_policy is not None:
-            Rt = self.calculate_Rt(S_fracs_within_age_group, self.suppression_policy(self.t_list))
+        if self.approximate_R0:
+            Rt = self.suppression_policy(self.t_list)
```
zhang-xinyu, line 639: Should this be: 
```
R0 = self.calculate_R0()
Rt = R0 * self.suppression_policy(self.t_list)
```
?

PR #248, opened 2020-04-16
Remove hospitalization data filter and add cumulative total infections

erccarls:
Remove hospitalization data filter and add cumulative total infections

### Please Check
- [ ] Are tests passing?


erccarls:
@ghop02 This was Igor's request. It was previously empty.

PR #267, opened 2020-04-18
Workflow will alert slack when we label something

goldblatt:

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/.github/workflows/label_api_snapshot.yml
+++ b/.github/workflows/label_api_snapshot.yml
@@ -42,3 +42,11 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         SOURCE_DIR: 's3://${{env.AWS_S3_BUCKET}}/snapshot/${{env.SNAPSHOT_ID}}/'
         DEST_DIR: '${{env.LABEL}}/'
+    
+    - name: Slack notification
+      env:
+        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+        SLACK_CHANNEL: api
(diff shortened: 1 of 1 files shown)


.github/workflows/label_api_snapshot.yml
```diff
@@ -42,3 +42,11 @@ jobs:
...
         SOURCE_DIR: 's3://${{env.AWS_S3_BUCKET}}/snapshot/${{env.SNAPSHOT_ID}}/'
         DEST_DIR: '${{env.LABEL}}/'
+    
+    - name: Slack notification
+      env:
+        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+        SLACK_CHANNEL: api
+      uses: Ilshidur/action-slack@fb92a78
+      with:
+        args: '{{env.SNAPSHOT_ID}} has been labelled {{env.LABEL}}'
```
ghop02, line 52: would be good to always run this `if: always()` and alert on the status (in case it fails).

this is the Github action i've used, if it's helpful: https://github.com/8398a7/action-slack
```
steps:
  - uses: 8398a7/action-slack@v3
    with:
      status: ${{ job.status }}
      fields: repo,message,commit,author,action,eventName,ref,workflow # selectable (default: repo,message)
    env:
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # optional
      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
    if: always() # Pick up events even if the job fails or is canceled.
```

goldblatt: ohh i took some of the stuff like job.status from here, but i aesthetically liked this one (which like if we need the other one can switch LOL)

PR #271, opened 2020-04-19
Model/add rt inference

erccarls:
This PR adds a class for state and county level inference of R_t for case, mortality, and hospitalization data based on an extension of Kevin Systrom's analysis.  A forthcoming PR will add integration between this and inference.

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 import numpy as np
 import pandas as pd
 from scipy import stats as sps
+from scipy import signal
 from matplotlib import pyplot as plt
 import us
 from pyseir import load_data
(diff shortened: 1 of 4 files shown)


pyseir/inference/infer_rt.py
```diff
@@ -0,0 +1,460 @@
...
+        ).get_average_seir_parameters()
+
+        # Serial period = Incubation + 0.5 * Infections
+        self.serial_period = 1 / self.default_parameters['sigma'] + 0.5 * 1 /   self.default_parameters['delta']
+
+        # If we only receive current hospitalizations, we need to account for
+        # the outflow to reconstruct new admissions.
+        if self.hospitalization_data_type is load_data.HospitalizationDataType.CURRENT_HOSPITALIZATIONS:
+            los_general = self.default_parameters['hospitalization_length_of_stay_general']
+            los_icu = self.default_parameters['hospitalization_length_of_stay_general']
```
rjchacko, line 109: Should this be using `hospitalization_length_of_stay_icu` instead?

```suggestion
            los_icu = self.default_parameters['hospitalization_length_of_stay_icu']
```

pyseir/inference/infer_rt.py
```diff
@@ -0,0 +1,460 @@
...
+            hosp_rate_general = self.default_parameters['hospitalization_rate_general']
+            hosp_rate_icu = self.default_parameters['hospitalization_rate_icu']
+            icu_rate = hosp_rate_icu / hosp_rate_general
+            flow_out_of_hosp = self.hospitalizations[:-1] * ((1 - icu_rate) / los_general + icu_rate / los_icu)
+            # We are attempting to reconstruct the cumulative hospitalizations.
+            self.hospitalizations = np.diff(self.hospitalizations) + flow_out_of_hosp
```
rjchacko, line 115: Should the sign be flipped on `flow_out_of_hospital`?
```suggestion
            self.hospitalizations = np.diff(self.hospitalizations) - flow_out_of_hosp
```

pyseir/inference/infer_rt.py
```diff
@@ -0,0 +1,460 @@
...
+
+        for timeseries_type in available_timeseries:
+
+            df = pd.DataFrame()
+            dates, times, posteriors = self.get_posteriors(timeseries_type)
+            if posteriors is not None:
+                df[f'Rt_MAP__{timeseries_type.value}'] = posteriors.idxmax()
+                for ci in self.confidence_intervals:
+                    ci_low, ci_high = self.highest_density_interval(posteriors, ci=ci)
+                    df[f'Rt_ci{int(100 * (1 - ci))}__{timeseries_type.value}'] = ci_low
```
zhang-xinyu, line 346: Maybe just change to `Rt_ci{int(100 * ci)}_low`/`Rt_ci{int(100 * ci)}_high`? Lower bound of 95% interval should be Rt_ci2.5 instead of Rt_ci5 right?

pyseir/inference/infer_rt.py
```diff
@@ -0,0 +1,460 @@
...
+                    df[f'Rt_ci{int(100 * ci)}__{timeseries_type.value}'] = ci_high
```
zhang-xinyu, line 347: Same question here.

pyseir/inference/infer_rt.py
erccarls: No, because we are trying to reconstruct the new cumulative cases, so we have to add back the cases that left first, and then compute the derivative.

pyseir/inference/infer_rt.py
```diff
@@ -0,0 +1,460 @@
...
+            Output integers since the reference date.
+        posteriors: pd.DataFrame
+            Posterior estimiates for each timestamp with non-zero data.
+        """
+        dates, times, timeseries = self.apply_gaussian_smoothing(timeseries_type)
+        if len(timeseries) == 0:
+            return None, None, None
+
+        # (1) Calculate Lambda
+        lam = timeseries[:-1].values * np.exp((self.r_list[:, None] - 1) / self.serial_period)
```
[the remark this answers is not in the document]
erccarls: Lambda is the poisson likelihood over R_t for a given increase from t-1 to t cases. comment added.

pyseir/inference/infer_rt.py
erccarls: I changed to 90% CIs and accepted your changes above.

Issue #268, opened 2020-04-18
Create Batch versions of API data


ghop02:
Right now, the API reports per state and per county data, separated out into individual files.
`/latest/us/states/` and `/latest/us/counties/` . 
We don't yet have this data aggregated into batch versions of this data:

 - /latest/us/states(.intervention).json
 - /latest/us/states(.intervention).timeseries.json
 - /latest/us/counties(.intervention).json
 - /latest/us/counties(.intervention).timeseries.json

This will mostly be a rollup of what we currently have.

ghop02:
couple questions - do we want to surface just the raw data anwywhere without a specific intervention?

Also a quick proposed schema change:
```diff
- /latest/us/states(.intervention).json
- /latest/us/states(.intervention).timeseries.json
+ /latest/us/states.summary(.intervention).json
+ /latest/us/states.timeseries(.intervention).json
+ /latest/us/states(.intervention).json
```
the idea being that `states(.intervention).json` has all timeseries and summary data

igorkofman:
1. Define "raw data"?
2. Let's stick to current schema to start

ghop02:
Oh, sorry, `actuals` would've been the better term.

And sounds good.

PR #273, opened 2020-04-20
Adding bulk files

ghop02:
Adds 4 new files:
```
 - us/states(.intervention).{json,csv}
 - us/states(.intervention).timeseries.json
 - us/states(.intervention).{json,csv}
 - us/counties(.intervention).timeseries.json
```

This PR addresses issue https://github.com/covid-projections/covid-data-model/issues/268

### Please Check
- [x] Are tests passing?

### DIFF
--- a/cli/run_states_api.py
+++ b/cli/run_states_api.py
@@ -45,8 +45,11 @@ def deploy_states_api(disable_validation, input_dir, output, summary_output):
         states_results_api = api_pipeline.generate_api(states_result, input_dir)
         api_pipeline.deploy_results(states_results_api, output)
 
-        states_summary = api_pipeline.build_states_summary(states_results_api, intervention)
+        states_summary = api_pipeline.build_states_summary(
+            states_results_api, intervention
+        )
         states_timeseries = api_pipeline.build_states_timeseries(states_results_api, intervention)
-        api_pipeline.deploy_results([states_summary, states_timeseries], summary_output)
+        api_pipeline.deploy_results([states_summary], summary_output, write_csv=True)
+        api_pipeline.deploy_results([states_timeseries], summary_output)
 
         logger.info("finished states job")
--- a/cli/run_states_api.py
+++ b/cli/run_states_api.py
@@ -27,7 +27,12 @@ PROD_BUCKET = "data.covidactnow.org"
     default="results/output/states",
     help="Output directory for artifacts",
 )
-def deploy_states_api(disable_validation, input_dir, output):
+@click.option(
+    "--summary-output",
+    default="results/output",
+    help="Output directory for state summaries.",
+)
+def deploy_states_api(disable_validation, input_dir, output, summary_output):
     """The entry function for invocation"""
 
     for intervention in list(Intervention):
@@ -39,5 +44,9 @@ def deploy_states_api(disable_validation, input_dir, output):
         )
         states_results_api = api_pipeline.generate_api(states_result, input_dir)
         api_pipeline.deploy_results(states_results_api, output)
+
+        states_summary = api_pipeline.build_states_summary(states_results_api, intervention)
+        states_timeseries = api_pipeline.build_states_timeseries(states_results_api, intervention)
+        api_pipeline.deploy_results([states_summary, states_timeseries], summary_output)
 
         logger.info("finished states job")
--- a/libs/dataset_deployer.py
+++ b/libs/dataset_deployer.py
@@ -1,4 +1,7 @@
+from typing import List
 import os
+import pathlib
+import csv
 import io
 import logging
 import boto3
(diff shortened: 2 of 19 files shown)


ghop02:
The only difference between this and the API spec is that the `.timeseries` files do have the summaries in them.

This is mostly to mirror the `{fips,state}.timeseries` files. The schema is the same, the data is just a list.

libs/dataset_deployer.py
```diff
@@ -67,9 +70,54 @@ def upload_csv(key_name: str, csv: str, output_dir: str):
...
+        data: data to flatten
+        level_separator: separator to use when combining keys from nested dictionary.
+    """
+    flattened = {}
+    for key, value in data.items():
+        if not isinstance(value, dict):
+            flattened[key] = value
+            continue
+
+        value = flatten_dict(value)
```
goldblatt, line 86: i think this should be nested_value

libs/functions/generate_api.py
```diff
@@ -121,7 +121,7 @@ def generate_api_for_state_timeseries(projection_row, intervention, input_dir):
     )
 
 
-def generate_api_for_county_timeseries(projection_row, intervention, input_dir):
+def generate_county_timeseries(projection_row, intervention, input_dir):
```
goldblatt, line 124: wanna give this a type as well

libs/pipelines/api_pipeline.py
```diff
@@ -14,6 +23,19 @@
...
+        """
+        Args:
+            file_stem: Stem of output filename.
+            data: Data
+
+        """
+        self.file_stem = file_stem
+        self.data = data
+
+
```
goldblatt, line 38: I wonder if we want to include the intervention here as well

libs/pipelines/api_pipeline.py
```diff
@@ -80,60 +102,118 @@ def run_projections(
...
-
-def deploy_results(result: APIGeneration, output: str):
+    return results
+
+
+def build_states_summary(state_data: List[APIOutput], intervention) -> APIOutput:
+    state_summaries = [
+        output.data
+        for output in state_data
+        # Don't love using type here, but CovidActNowStateTimeseries inherits
```
goldblatt, line 152: is the issue here that we return a list now instead of a tuple? like can't we just keep it a tuple?

libs/pipelines/api_pipeline.py
```diff
@@ -80,60 +102,118 @@ def run_projections(
...
+    counties_data: List[APIOutput], intervention
+) -> APIOutput:
+    county_summaries = [
+        output.data
+        for output in counties_data
+        if isinstance(output.data, CovidActNowCountyTimeseries)
+    ]
+    county_api_data = CovidActNowCountiesTimeseries(data=county_summaries)
+    key = f"counties.{intervention.name}.timeseries"
+    return APIOutput(key, county_api_data)
```
goldblatt, line 197: these all seem relatively duplicated, I think there's probably something we can do with like passing in classes or something

run.sh
```diff
@@ -47,6 +47,7 @@ prepare () {
   # TODO: These should perhaps live in python, near the schemas (defined in api/)?
   API_OUTPUT_COUNTIES="${API_OUTPUT_DIR}/us/counties"
   API_OUTPUT_STATES="${API_OUTPUT_DIR}/us/states"
+  API_OUTPUT_US="${API_OUTPUT_DIR}/us"
```
goldblatt, line 50: can we call this summart output?

libs/dataset_deployer.py
```diff
@@ -67,9 +70,54 @@ def upload_csv(key_name: str, csv: str, output_dir: str):
     _logger.info(f"Generated csv for {key_name}")
 
 
-def upload_json(key_name, json: str, output_dir: str): 
+def flatten_dict(data: dict, level_separator: str = '.') -> dict:
```
goldblatt, line 73: is there no way this dictionary can have lists in it already?

cli/run_states_api.py
```diff
@@ -40,4 +45,11 @@ def deploy_states_api(disable_validation, input_dir, output):
         states_results_api = api_pipeline.generate_api(states_result, input_dir)
         api_pipeline.deploy_results(states_results_api, output)
```
goldblatt, line 46: I think i would change the name here to be like state_result_list to be clear

libs/pipelines/api_pipeline.py
```diff
@@ -80,60 +102,118 @@ def run_projections(
...
         )
     else:
         raise ValueError("Aggregate Level not supported by api generation")
     key_prefix = _get_api_prefix(projection_result.aggregation_level, row)
     generated_key = f"{key_prefix}.{projection_result.intervention.name}.timeseries"
-    return APIGenerationRow(generated_key, generated_data)
+    return APIOutput(generated_key, generated_data)
 
 
 def generate_api(
```
goldblatt, line 131: I think we could refactor this to just return two lists then we wouldn't have to do that type checking stuff?

ghop02: Yeah I was just thinking that, if it returned summaries and summaries with timestamp life would be much easier

libs/dataset_deployer.py
ghop02: So I think it can have lists... as of now the list will just be one column. Can raise an error if we think that's a problem.

goldblatt: cool i think that's is okay (I mean i think some slightly complicated things can happen like a list of dictionaries), but maybe we deal with that if/when it happens

ghop02:
Great, I'm going to merge this as is, and then separate those out first thing when i get a chance!

PR #274, opened 2020-04-20
Parallelize Run Model 3x

fridiculous:
`execute_model` now runs in parallel in 3x.

Original benchmark was 50 min for execute_model
Now execute_model runs in 20 min.

20x, 10x and 5x were also tried, however, the reliability of the runs was unsuccessful.  Overall, CPU utilization improved, however, there is still a lot on table.  Perfectly efficient CPU would result <10 min per run.

### DIFF
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -18,10 +18,10 @@ on:
 jobs:
   execute-model:
     runs-on: self-hosted
-    strategy:
-      max-parallel: 25
-      matrix:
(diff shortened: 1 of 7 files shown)


mikelehen:
Sorry to be a grump, but can we chat before moving forward with this?  It adds extra complexity (harder to understand and harder to debug issues) and fragility (states/territories are hardcoded in an additional place).  

If the speed-up was incredible, then it would likely be worth it.  But there's not that much meaningful difference between 20min and 50min.  Both are basically "kick it off and come back after lunch", especially when you add in the time for the rest of the process (API, deploy to S3).

In general, I'm pretty loathe to be doing parallelization at multiple layers concurrently, especially when we're still not utilizing the machine optimally.  It feels like maybe instead we need to simplify things until the parallelization becomes straightforward.

PR #279, opened 2020-04-21
20x Api Speed.  Remove Unneeded Computation, Refactor & Parallelize

fridiculous:
New results. 1 min, 50 seconds locally
Previous result:  >> 40 min locally, (25+ min on Github actions)

Change made:
* Cache intervention.json network request happening from covid-public 
* remove unused boto session that's done over and over again
* Use known date format for faster date conversions
* refactor iterrows to use pandas native map functions
* use pandas apply to parallelize per county manipulations

## The Result
```bash
run.sh ../covid-data-public/ ../data/api-results-118/ execute_api  440.69s user 26.94s system 423% cpu 1:50.30 total
```

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/dataset_deployer.py
+++ b/libs/dataset_deployer.py
@@ -1,41 +1,21 @@
 import os
 import io
 import logging
-import boto3
 
 _logger = logging.getLogger(__name__)
 
 
 class DatasetDeployer(object):
-    """Common deploy operations for persisting files to s3 or local folder.
+    """Common deploy operations for persisting files to a local folder.
 
-    Deploys to S3 if the BUCKET_NAME environment variable is set.
     """
 
     def __init__(self, key="filename.csv", body="a random data", output_dir="."):
-        self.s3 = boto3.client("s3")
-        # Supplied by ENV on AWS
-        # BUCKET_NAME format is s3://{BUCKET_NAME}
-        self.bucket_name = os.environ.get("BUCKET_NAME")
         self.key = key
         self.body = body
         self.output_dir = output_dir
 
-    def _persist_to_s3(self):
-        """Persists specific data onto an s3 bucket.
-        This method assumes versioned is handled on the bucket itself.
-        """
-        _logger.info(f"persisting {self.key} to s3")
-
-        response = self.s3.put_object(
-            Bucket=self.bucket_name, Key=self.key, Body=self.body, ACL="public-read"
-        )
-        return response
-
     def _persist_to_local(self):
-        """Persists specific data onto an s3 bucket.
-        This method assumes versioned is handled on the bucket itself.
-        """
         _logger.info(f"persisting {self.key} to local")
 
         with open(os.path.join(self.output_dir, self.key), "wb") as f:
@@ -46,14 +26,8 @@ class DatasetDeployer(object):
                 self.body.encode("UTF-8") if isinstance(self.body, str) else self.body
             )
 
-        pass
-
     def persist(self):
-        if self.bucket_name:
-            self._persist_to_s3()
-        else:
-            self._persist_to_local()
-        return
+        self._persist_to_local()
 
 
 def upload_csv(key_name: str, csv: str, output_dir: str):
@@ -67,7 +41,7 @@ def upload_csv(key_name: str, csv: str, output_dir: str):
     _logger.info(f"Generated csv for {key_name}")
 
 
-def upload_json(key_name, json: str, output_dir: str): 
+def upload_json(key_name, json: str, output_dir: str):
     DatasetDeployer(f"{key_name}.json", json, output_dir).persist()
     _logger.info(f"Generated json for {key_name}")
 
--- a/libs/functions/calculate_projections.py
+++ b/libs/functions/calculate_projections.py
@@ -16,6 +16,10 @@ from libs.datasets.projections_schema import (
 from libs.enums import Intervention
 from libs.constants import NULL_VALUE
 
+from pandarallel import pandarallel
+pandarallel.initialize(progress_bar=True)
+
+
 def _calc_short_fall(x):
     return abs(x.beds - x.all_hospitalized) if x.all_hospitalized > x.beds else 0
 
@@ -99,7 +103,6 @@ def _calculate_projection_data(state, fips, file_path):
         record['Beds at Peak Hospitilization Date'] = beds_at_peak_hospitalization_date
         record['Population'] = population
 
-        return pd.Series(record)
     return pd.Series(record)
 
 def _get_intervention_type(intervention_type, state, state_interventions_df):
@@ -133,25 +136,24 @@ def get_file_path(input_dir, state, fips, intervention_type):
     file_name = f"{state}.{fips}.{intervention_type}.json"
     return os.path.join(input_dir, file_name)
 
-from tqdm import tqdm
-tqdm.pandas()
-
 
 def get_county_projections_df(input_dir, initial_intervention_type, state_interventions_df):
     """
     for each state in our data look at the results we generated via run.py
     to create the projections
+
+    #columns=CALCULATED_PROJECTION_HEADERS_COUNTIES)
     """
     fips_pd = FIPSPopulation.local().data  # to get the state, county & fips
 
     fdf = fips_pd[['state' ,'fips']]
-    fdf.loc[:,'intervention_type'] = fdf.state.apply(
+    fdf.loc[:,'intervention_type'] = fdf.state.parallel_apply(
         lambda x: _get_intervention_type(initial_intervention_type, x, state_interventions_df)
         )
-    fdf.loc[:, 'path'] = fdf.apply(
+    fdf.loc[:, 'path'] = fdf.parallel_apply(
         lambda x: get_file_path(input_dir, x.state, x.fips, x.intervention_type),
         axis=1).values
-    ndf = fdf.progress_apply(
+    ndf = fdf.parallel_apply(
         lambda x:_calculate_projection_data(x.state, x.fips, x.path), axis=1)
 
     missing = ndf.isnull().sum()['State']
@@ -159,6 +161,4 @@ def get_county_projections_df(input_dir, initial_intervention_type, state_interv
     if (missing > 2000):
         raise Exception(f"Missing a majority of counties from input_dir: {input_dir}")
     print(f"Models missing for {missing} counties")
-
-    #ndf = pd.DataFrame(results, columns=CALCULATED_PROJECTION_HEADERS_COUNTIES)
     return ndf
--- a/libs/functions/calculate_projections.py
+++ b/libs/functions/calculate_projections.py
@@ -40,7 +40,7 @@ def _read_json_as_df(path):
         exclude=CAN_MODEL_OUTPUT_SCHEMA_EXCLUDED_COLUMNS,
     )
 
-    df["date"] = pd.to_datetime(df.date)
+    df["date"] = pd.to_datetime(df.date, format='%m/%d/%y')
     df["all_hospitalized"] = df["all_hospitalized"].astype("int")
     df["beds"] = df["beds"].astype("int")
     df["dead"] = df["dead"].astype("int")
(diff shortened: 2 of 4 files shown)


.github/workflows/deploy_api.yml
```diff
@@ -9,7 +9,7 @@ on:
   schedule:
    - cron: '30 0,12 * * *'
 
-  # push:
+  push:
```
ghop02, line 12: Was this not enabled before for a particular reason?

.github/workflows/deploy_api.yml
```diff
@@ -189,7 +189,7 @@ jobs:
       # To pin to an old data set or model code, put the branch/tag/commit here:
       COVID_DATA_PUBLIC_REF: 'master'
       # !!! Change this to your BRANCH if you want to test it
-      COVID_DATA_MODEL_REF: 'master'
+      COVID_DATA_MODEL_REF: 'fridiculous/improve-speed-of-gen-api'
```
ghop02, line 192: need to clean these up

libs/dataset_deployer.py
```diff
@@ -4,37 +4,20 @@
...
 import io
 import logging
-import boto3
 
 _logger = logging.getLogger(__name__)
 
 
 class DatasetDeployer(object):
-    """Common deploy operations for persisting files to s3 or local folder.
+    """Common deploy operations for persisting files to a local folder.
```
ghop02, line 12: I think there might be a couple other references to s3 somewhere (maybe in some of the cli commands).  can you look for those and clean them up?

libs/functions/calculate_projections.py
```diff
@@ -16,6 +16,10 @@
 from libs.enums import Intervention
 from libs.constants import NULL_VALUE
 
+from pandarallel import pandarallel
+pandarallel.initialize(progress_bar=True)
```
ghop02, line 20: I would prefer not to do this initialization at the module level.  would it be possible to add this in to probably `run.py`? We don't yet have a common place to initialize all of these things - may be worth creating a function that does that initialization (would setup sentry, logging and now this)

libs/functions/calculate_projections.py
```diff
@@ -108,46 +119,57 @@ def get_state_projections_df(input_dir, initial_intervention_type, state_interve
...
     to create the projections
+
+    columns=CALCULATED_PROJECTION_HEADERS_STATES
     """
 
-    # save results in a list of lists, converted to df later
-    results = []
-    for state in list(US_STATE_ABBREV.values()):
-        intervention_type = _get_intervention_type(initial_intervention_type, state, state_interventions_df)
+    sdf = pd.DataFrame(US_STATE_ABBREV.values(), columns=['state'])
```
ghop02, line 126: can you rename this `state_df`?

libs/functions/calculate_projections.py
```diff
@@ -108,46 +119,57 @@ def get_state_projections_df(input_dir, initial_intervention_type, state_interve
...
+        )
+    sdf.loc[:, 'path'] = sdf.parallel_apply(
+        lambda x: get_file_path(input_dir, x.state, x.intervention_type, fips=None),
+        axis=1).values
+    ndf = sdf.parallel_apply(
+        lambda x:_calculate_projection_data(x.state, x.path, fips=None), axis=1)
+
+    num_processed_states = ndf.notnull().sum()['State']
+
+    if (num_processed_states < 51):
```
ghop02, line 138: no parens needed here. I'm also worried that this will make it harder to test if we're just running one state.  Can we just do this in the existing validation we do (so that we're able to skip it?)

libs/functions/calculate_projections.py
```diff
@@ -80,21 +87,25 @@ def _calculate_projection_data(file_path):
...
-            beds_at_peak_hospitalization_date,
-            population
-        ]
-    return None
+
+        record['State'] = state
+        if fips:
+            record['FIPS'] = fips
+
+        record['16-day_Hospitalization_Prediction'] = hosp_16_days
```
ghop02, line 95: Are these keys constants somewhere? we should use those.

libs/functions/calculate_projections.py
```diff
@@ -108,46 +119,57 @@ def get_state_projections_df(input_dir, initial_intervention_type, state_interve
...
-    for state in list(US_STATE_ABBREV.values()):
-        intervention_type = _get_intervention_type(initial_intervention_type, state, state_interventions_df)
+    sdf = pd.DataFrame(US_STATE_ABBREV.values(), columns=['state'])
+    sdf.loc[:,'intervention_type'] = sdf.state.parallel_apply(
+        lambda x: _get_intervention_type(initial_intervention_type, x, state_interventions_df)
+        )
+    sdf.loc[:, 'path'] = sdf.parallel_apply(
+        lambda x: get_file_path(input_dir, x.state, x.intervention_type, fips=None),
+        axis=1).values
+    ndf = sdf.parallel_apply(
```
ghop02, line 133: What does ndf stand for?

libs/functions/calculate_projections.py
```diff
@@ -108,46 +119,57 @@ def get_state_projections_df(input_dir, initial_intervention_type, state_interve
...
     to create the projections
+
+    #columns=CALCULATED_PROJECTION_HEADERS_COUNTIES)
     """
     fips_pd = FIPSPopulation.local().data  # to get the state, county & fips
 
-    # save results in a list of lists, converted to df later
-    results = []
+    fdf = fips_pd[['state' ,'fips']]
+    fdf.loc[:,'intervention_type'] = fdf.state.parallel_apply(
```
ghop02, line 161: This doesn't really need to be parallel does it? it's just an enum right?

libs/functions/calculate_projections.py
```diff
@@ -108,46 +119,57 @@ def get_state_projections_df(input_dir, initial_intervention_type, state_interve
...
+        lambda x: _get_intervention_type(initial_intervention_type, x, state_interventions_df)
+        )
+    fdf.loc[:, 'path'] = fdf.parallel_apply(
```
ghop02, line 164: This also probably doesn't need to be parallel, there's no intensive operation there

libs/functions/calculate_projections.py
```diff
@@ -80,21 +87,25 @@ def _calculate_projection_data(file_path):
...
+        record['Mean Hospitalizations'] = mean_hospitalizations
+        record['Mean Deaths'] = mean_deaths
+        record['Peak Hospitalizations On'] = peak_hospitalizations_date
+        record['Peak Deaths On'] = peak_deaths_date
+        record['Hospital Shortfall Date'] = hospitals_shortfall_date
+        record['Peak Hospitlizations Shortfall'] = peak_hospitalizations_short_falls
+        record['Beds at Peak Hospitilization Date'] = beds_at_peak_hospitalization_date
+        record['Population'] = population
+
+    return pd.Series(record)
```
ghop02, line 108: It feels weird that this is returning a series.  I do love these having keys, but doesn't feel like it needs to be a series

libs/functions/calculate_projections.py
```diff
@@ -108,46 +119,57 @@ def get_state_projections_df(input_dir, initial_intervention_type, state_interve
...
-        intervention_type = _get_intervention_type(initial_intervention_type, state, state_interventions_df)
+    sdf = pd.DataFrame(US_STATE_ABBREV.values(), columns=['state'])
+    sdf.loc[:,'intervention_type'] = sdf.state.parallel_apply(
+        lambda x: _get_intervention_type(initial_intervention_type, x, state_interventions_df)
+        )
+    sdf.loc[:, 'path'] = sdf.parallel_apply(
+        lambda x: get_file_path(input_dir, x.state, x.intervention_type, fips=None),
+        axis=1).values
+    ndf = sdf.parallel_apply(
+        lambda x:_calculate_projection_data(x.state, x.path, fips=None), axis=1)
```
ghop02, line 134: I'm a bit hesitant at this refactor - I like that the intervention type and path are separated out from the calculate projection data part (so that we can in fact parallelize those operations), but it seems unnecessary to use pandas.
```python
interventions_and_paths = []
for state in US_STATE_ABBREV.values():
    path = get_path()
    intervention = get_intervention_type()

pool = Pool(10 or w/e)
projection_data = pool.map(_calculate_projection_data, interventions_and_path)
missing = [result for result in projection_data if result is None]
results = [result for result in projection_data if result is not None]
return pd.DataFrame(results, columns=CALCULATED_PROJECTION_HEADERS_STATES)
```

ghop02: just a thought there ^

.github/workflows/deploy_api.yml
fridiculous: it's enabled just for testing.  I take it off before.

We can't have an hour long job go off on every push.

.github/workflows/deploy_api.yml
fridiculous: yup, just for testing

libs/functions/calculate_projections.py
fridiculous: Makes sense.  Should we do that in a separate module level init branch?

libs/functions/calculate_projections.py
fridiculous: it was the previous df name in counties.  I can call it `new_df`

libs/functions/calculate_projections.py
fridiculous: I'll remove it

libs/functions/calculate_projections.py
fridiculous: They are in Columns

libs/functions/calculate_projections.py
fridiculous: Pandas recomends returning series when build dataframes as a result of an apply

libs/functions/calculate_projections.py
fridiculous: Yeah, I understand.  I started with the same code shown above, but went away form multiprocessing, and even tried joblib, but for a variety of technical, complexity issues, and practices, i think it's best to use `apply` with parallelization.


1. My understanding is that using "for loops" is a not a best practice in pandas.  And removing for loops improved performance when going from iterrows => apply.

2. In addition, with this type of refactor, if something needs parallelization, we can just simply change it from apply to parallel_apply.

3. Wrt parallelizatio nitself, started with multiprocessing.pool.  Ran into issues pickling the data, (just using tuples and lists). As a result, I had to write specific functions that just handled how data was being returned to specific rows.   After that, we have to reaggregate the rows, and handle rows that are null differently from rows that are full.  So now we have mappers that return multiple types -> anti-pattern.

There's just a lot of code to do the basic stuff that `parellel_apply` does.

I also benchmarked swifter pandarallel and even looked at dask (but didn't try it).

Once I settled with pandaparallel, it just worked, and i removed a ton of code.

fridiculous:
> couple comments - I think I'd rather not use pandarallel (though it does look cool!) just because it adds a bit more complexity that we can easily get around with built in multiprocessing.
> 
> but apart from comments on the `_calculate_projection_data` df, lgtm

Yea, not sure why it looks like it adds more complexity.  Where is the new complexity?

libs/functions/calculate_projections.py
ghop02: can we prefer to throw this in `run.py` for now

libs/functions/calculate_projections.py
```diff
@@ -108,46 +119,56 @@ def get_state_projections_df(input_dir, initial_intervention_type, state_interve
...
+
+    columns=CALCULATED_PROJECTION_HEADERS_STATES
     """
 
-    # save results in a list of lists, converted to df later
-    results = []
-    for state in list(US_STATE_ABBREV.values()):
-        intervention_type = _get_intervention_type(initial_intervention_type, state, state_interventions_df)
+    states_df = pd.DataFrame(US_STATE_ABBREV.values(), columns=['state'])
+    states_df.loc[:,'intervention_type'] = states_df.state.apply(
```
ghop02, line 127: Can you run `black` on this file?

fridiculous: done sir

PR #280, opened 2020-04-21
Minor Bugfixes

erccarls:
Patch whitelist generation, linting error (which should not be showing up..) and skipping of bad inference results.


erccarls:
No one is awake, but this one is straightforward and fixes broken master, so I am going ahead with merge.

PR #284, opened 2020-04-21
Pass Rt through API

igorkofman:

### Please Check
- [ ] Are tests passing?


libs/datasets/can_model_output_schema.py
```diff
@@ -32,13 +34,13 @@
     DEAD,
     BEDS,
     CUMULATIVE_INFECTED,
-    "j",
-    "k",
+    R_T,
+    R_T_STDEV,
```
ghop02, line 38: running out of letters! :)

libs/functions/calculate_projections.py
```diff
@@ -80,6 +82,8 @@ def _calculate_projection_data(file_path):
         peak_hospitalizations_short_falls = df.iloc[df.all_hospitalized.idxmax()].short_fall
         peak_deaths_date = df.iloc[df.new_deaths.idxmax()].date
         population = df.iloc[0].population
+        r_t = df.iloc[10].r_t
```
ghop02, line 85: why 10?

libs/datasets/projections_schema.py
```diff
@@ -31,6 +33,8 @@
     "Peak Hospitlizations Shortfall",
     "Beds at Peak Hospitilization Date",
     "Population"
+    "Rt"
+    "Rt_stdev"
```
goldblatt, line 37: I think these need commas :)

igorkofman: ty!

PR #287, opened 2020-04-22
Expose ICU / Vent Compartments and Capacity Information

erccarls:
This PR addresses two issues:

1. Exposing vent, icu capacity, and vent utilization.
2. Adding column names to clarify.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/deployment/webui_data_adaptor_v1.py
+++ b/pyseir/deployment/webui_data_adaptor_v1.py
@@ -121,7 +121,11 @@ class WebUIDataAdaptorV1:
         policies = [key for key in pyseir_outputs.keys() if key.startswith('suppression_policy')]
 
         all_hospitalized_today = None
-        fit_results = load_inference_result(fips)
+        try:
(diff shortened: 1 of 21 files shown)


pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -123,28 +124,31 @@ def map_fips(self, fips):
...
+                        and fips not in self.df_whitelist.fips.values):
                 continue
 
             output_for_policy = pyseir_outputs[suppression_policy]
             output_model = pd.DataFrame()
 
+            # Hospitalizations need to be rescaled by the inferred factor to
+            # match observations for display purposes.
             if suppression_policy == 'suppression_policy__inferred' and fit_results:
-                if len(fips) == 5 and fips not in self.df_whitelist.fips.values:
```
erccarls, line 142: This is moved to the above location to include counties that don't have fit results.

PR #291, opened 2020-04-22
Update fit_results lookup by fips

erccarls:
This PR addresses 2 issues

1. incorrect county R_t shipping. Previously, state lookup codes were being used incorrectly resulting in the "first" county's R_t being used for subsequent fips in a state.
2. Whitelist filtering was previously being checked after also checking for valid fit results, which if not existing, went down an alternate code path.

Acceptance tests performed:
- Verified that correct county R_t are being shipped and differ across the state.
- Verified that the number of output/web_ui/county/*.2.json == 291 == whitelist length == number of MLE reports.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/deployment/webui_data_adaptor_v1.py
+++ b/pyseir/deployment/webui_data_adaptor_v1.py
@@ -123,13 +123,18 @@ class WebUIDataAdaptorV1:
         all_hospitalized_today = None
         try:
             fit_results = load_inference_result(fips)
-        except ValueError:
+        except (KeyError, ValueError):
             fit_results = None
-            logging.error(f'Fit result not found for {fips}: Skipping inference elements')
+            logging.warning(f'Fit result not found for {fips}: Skipping inference elements')
 
         for i_policy, suppression_policy in enumerate(policies):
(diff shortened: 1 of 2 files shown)


pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -123,13 +123,18 @@ def map_fips(self, fips):
...
+        except (KeyError, ValueError):
             fit_results = None
-            logging.error(f'Fit result not found for {fips}: Skipping inference elements')
+            logging.warning(f'Fit result not found for {fips}: Skipping inference elements')
 
         for i_policy, suppression_policy in enumerate(policies):
             if suppression_policy == 'suppression_policy__full_containment':  # No longer shipping this.
                 continue
+            if suppression_policy == 'suppression_policy__inferred' \
+                    and len(fips) == 5 \
```
igorkofman, line 134: what is this for?

mikelehen: I think it's so we don't try to apply the whitelist to states (which have fips codes of length 2)

mikelehen:
Merging so this gets included in tonight's snapshot...

Issue #281, opened 2020-04-21
Add test rate results to state historical summaries


ghop02:
idea for how this should go:
* Add a field here for testing rate: https://github.com/covid-projections/covid-data-model/blob/master/api/can_api_definition.py#L50

* modify `get_usa_by_states_df` to add in state testing results by day [code](https://github.com/covid-projections/covid-data-model/blob/master/libs/build_processed_dataset.py#L145-L152):
  1. initialize `CovidTrackingDataSource.local()`
  2. Aggregate the latest results, taking the most recent testing data from each day
  3. join those results with the larger projections data frame.

* We then pass rows from that dataset into [here](https://github.com/covid-projections/covid-data-model/blob/master/libs/pipelines/api_pipeline.py#L104-L106).  it should then be pretty straightforward to add that into the output api [here](https://github.com/covid-projections/covid-data-model/blob/master/libs/functions/generate_api.py#L149-L161).

I'm not 100% sold on having to do it this way, but it would certainly be the easiest for right now. 

for adding in the state timeseries data, this may be the easiest place to add it in:
https://github.com/covid-projections/covid-data-model/blob/master/libs/functions/generate_api.py#L100-L105.

There are a few more levels of indirection than I'd like right now, but I think it'd be best to simplify as we go.

PR #292, opened 2020-04-22
Add Positive and Negative Tests to v1 API

fridiculous:
This PR addresses issue #281 
Works with actuals, and time series.

### Please Check
- [x] Are tests passing?


libs/build_processed_dataset.py
```diff
@@ -46,6 +51,36 @@ def _get_abbrev_df():
...
+
+    return ctd_df
+
+
+def get_testing_timeseries_by_state(state):
+    test_df = _get_testing_df()
+    # use a string for dates
+    test_df['date'] = test_df.date.apply(lambda x: x.strftime("%m/%d/%y"))
+    test_df['positive'] = test_df['positive'].astype(int)
+    test_df['negative'] = test_df['negative'].astype(int)
```
ghop02, line 76: Might be good to move this into the _get_testing_df() function!

api/schemas/CANPredictionTimeseriesRow.json
```diff
@@ -37,6 +37,16 @@
       "title": "Cumulativeinfected",
       "description": "Number of cumulative infections",
       "type": "integer"
+    },
+    "cumulativePosTested": {
```
mikelehen, line 41: nit: "Tested" => "Tests" maybe? since:
1. I think we need a noun in here. :-) I guess "Tested" _can_ be a noun, but 1) it's ambiguous, and 2) Since people can be tested multiple times, it's probably clearer to use "Tests"
2. "positive tested" and "negative tested" are just unusual phrasing.

Also, I'm rarely a fan of abbreviations unless they're ubiquitous.  So maybe go ahead and do `cumulativePositiveTests` and `cumulativeNegativeTests`?  Our JSON files will always be compressed (gzip) on the wire, so we don't need to work too hard to save bytes. :-)

api/can_api_definition.py
```diff
@@ -45,6 +45,12 @@ class _Actuals(pydantic.BaseModel):
     cumulativeConfirmedCases: int = pydantic.Field(
         ..., description="Number of confirmed cases so far"
     )
+    cumulativePosTested: Optional[int] = pydantic.Field(
```
ghop02, line 48: I don't think there's a reason to abbreviate here - doesn't look like we do anywhere else

libs/build_processed_dataset.py
```diff
@@ -46,6 +51,36 @@ def _get_abbrev_df():
     )
 
 
+def _get_testing_df():
+    # TODO: read this from a dataset class
+    abbrev_df = _get_abbrev_df()
+
+    ctd_df = CovidTrackingDataSource.local().data
+    ctd_df = ctd_df[CovidTrackingDataSource.TEST_FIELDS]
+    ctd_df = ctd_df.merge(
```
ghop02, line 60: What's the intent of this?

libs/build_processed_dataset.py
```diff
@@ -46,6 +51,36 @@ def _get_abbrev_df():
...
+def get_testing_timeseries_by_state(state):
+    test_df = _get_testing_df()
+    # use a string for dates
+    test_df['date'] = test_df.date.apply(lambda x: x.strftime("%m/%d/%y"))
+    test_df['positive'] = test_df['positive'].astype(int)
+    test_df['negative'] = test_df['negative'].astype(int)
+
+    state_test_df = test_df[test_df[CovidTrackingDataSource.Fields.STATE] == state]
+    # just select state
+    return state_test_df[['positive', 'negative', 'date']]
```
ghop02, line 80: Would love to refer to these by the class field names instead of the strings here

libs/build_processed_dataset.py
```diff
@@ -163,27 +202,50 @@ def get_usa_by_states_df(input_dir, intervention_type):
...
+            left_on="abbreviation",
+            right_on="State",
+            how="left",
+        ).drop(["abbreviation", "State"], axis=1)
     )
 
-    states_remapped = states_abbrev.rename(columns=OUTPUT_COLUMN_REMAP_TO_RESULT_DATA)
+    STATE_COLS_REMAP = OUTPUT_COLUMN_REMAP_TO_RESULT_DATA
+    STATE_COLS_REMAP["positive"] = CUMULATIVE_POSITIVE_TESTS
+    STATE_COLS_REMAP["negative"] = CUMULATIVE_NEGATIVE_TESTS
```
ghop02, line 235: Why is this done here?

libs/build_processed_dataset.py
```diff
@@ -163,27 +202,50 @@ def get_usa_by_states_df(input_dir, intervention_type):
...
+
+    states_remapped = states_abbrev.rename(columns=STATE_COLS_REMAP)
 
     states_final = pd.DataFrame(states_remapped, columns=RESULT_DATA_COLUMNS_STATES)
+    states_final[CUMULATIVE_POSITIVE_TESTS] = states_final[CUMULATIVE_POSITIVE_TESTS].fillna(0)
```
ghop02, line 240: hmm don't think we fill in zeros elsewhere? Not sure where we have landed on this.

libs/datasets/sources/covid_tracking_source.py
```diff
@@ -86,13 +96,38 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
...
+
         # Covid Tracking source has the state level fips, however none of the other
         # data sources have state level fips, and the generic code may implicitly assume
         # it doesn't.  I would like to add a state level fips (maybe for example a state fips code
         # of 45 being 45000), but it's not there, so in the meantime we're setting fips to null so
         # as not to confuse downstream data.
         data[cls.Fields.FIPS] = None
 
+        # must stay true: positive + negative  ==  total
+        assert (data[cls.Fields.POSITIVE_TESTS].fillna(0) +\
```
ghop02, line 119: can you run black on this? would love to not sneak in `\`'s

libs/build_processed_dataset.py
goldblatt: if we fill with zeros do we need the field to optional?

libs/functions/generate_api.py
```diff
@@ -95,20 +99,31 @@ def _generate_timeseries_row(json_data_row):
         ICUBedCapacity=None,
         cumulativeDeaths=json_data_row[can_schema.DEAD],
         cumulativeInfected=json_data_row[can_schema.CUMULATIVE_INFECTED],
+        cumulativePosTested=int(json_data_row['positive']),
+        cumulativeNegTested=int(json_data_row['negative']),
```
goldblatt, line 103: can we use the vars here instead of positive and negative?

libs/functions/generate_api.py
```diff
@@ -95,20 +99,31 @@ def _generate_timeseries_row(json_data_row):
...
     state_abbrev = US_STATE_ABBREV[projection_row[rc.STATE]]
     fips = projection_row[rc.FIPS]
-    can_dataseries = get_can_projection.get_can_raw_data(
+    raw_dataseries = get_can_projection.get_can_raw_data(
         input_dir, state_abbrev, fips, AggregationLevel.STATE, intervention
     )
+
+    # join in state testing data
+    test_df = get_testing_timeseries_by_state(projection_row[rc.STATE])
+    new_df = pd.DataFrame(raw_dataseries).merge(test_df, left_on='date', right_on='date', how='left').fillna(0)
```
goldblatt, line 115: ohh my gut it saying that pandas here is not the energy i want in this file

api/can_api_definition.py
fridiculous: I can go either way

libs/build_processed_dataset.py
fridiculous: column names need to be re-mapped

libs/build_processed_dataset.py
igorkofman: Yes, pls add a comment!

libs/build_processed_dataset.py
```diff
@@ -46,6 +51,42 @@ def _get_abbrev_df():
...
+        right_on="abbreviation",
+        how="inner",
+        suffixes=["_dropcol", ""],
+    ).drop("state_dropcol", axis=1)
+
+    return ctd_df
+
+
+def get_testing_timeseries_by_state(state):
+    test_df = _get_testing_df()
```
igorkofman, line 84: testing instead of "test"?

libs/build_processed_dataset.py
```diff
@@ -46,6 +51,42 @@ def _get_abbrev_df():
...
+    state_test_df = test_df[test_df[CovidTrackingDataSource.Fields.STATE] == state]
+    # just select state
```
igorkofman, line 86: is this a comment for the previous line?

libs/build_processed_dataset.py
```diff
@@ -145,6 +189,15 @@ def get_usa_by_states_df(input_dir, intervention_type):
     projections_df = get_state_projections_df(
         input_dir, intervention_type, interventions_df
     )
+    test_df = _get_testing_df()
```
igorkofman, line 192: comment would be awesome!

libs/datasets/projections_schema.py
```diff
@@ -1,4 +1,5 @@
 OUTPUT_COLUMN_REMAP_TO_RESULT_DATA = {
+    'state': 'Province/State',
     'state_x': 'Province/State',
```
igorkofman, line 3: what's the deal here?

libs/datasets/sources/covid_tracking_source.py
```diff
@@ -67,6 +67,22 @@ class Fields(object):
         TimeseriesDataset.Fields.AGGREGATE_LEVEL: Fields.AGGREGATE_LEVEL,
     }
 
+    TESTS_ONLY_FIELDS = [
+        Fields.DATE,
+        Fields.POSITIVE_TESTS,
+        Fields.NEGATIVE_TESTS,
+    ]
+
+    TEST_FIELDS = [
```
igorkofman, line 76: TESTS?

PR #297, opened 2020-04-22
Changing county summaries to iterate over model output files to pickup fips

ghop02:
Fixes https://github.com/covid-projections/covid-projections/issues/534

### DIFF
--- a/libs/pipelines/api_pipeline.py
+++ b/libs/pipelines/api_pipeline.py
@@ -1,7 +1,11 @@
 from typing import List
+import pathlib
 from collections import namedtuple
+from collections import defaultdict
 import logging
+
 import pydantic
+from api.can_api_definition import CountyFipsSummary
 from api.can_api_definition import CovidActNowCountySummary
 from api.can_api_definition import CovidActNowCountiesSummary
 from api.can_api_definition import CovidActNowCountiesTimeseries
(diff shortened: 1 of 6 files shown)


libs/pipelines/api_pipeline.py
```diff
@@ -179,6 +183,37 @@ def build_counties_timeseries(
...
+    """
+    found_fips_by_state = defaultdict(set)
+    found_fips = set()
+    for path in pathlib.Path(input_dir).iterdir():
+        if not str(path).endswith('.json'):
+            continue
+
+        state, fips, intervention, _ = path.name.split('.')
+        found_fips_by_state[state].add(fips)
+        found_fips.add(fips)
```
goldblatt, line 202: this feels a lil ruff don't we have access to this as a var somewhere?

libs/pipelines/api_pipeline.py
```diff
@@ -179,6 +183,37 @@ def build_counties_timeseries(
...
+
+    Returns: List of API Output objects.
+    """
+    found_fips_by_state = defaultdict(set)
+    found_fips = set()
+    for path in pathlib.Path(input_dir).iterdir():
+        if not str(path).endswith('.json'):
+            continue
+
+        state, fips, intervention, _ = path.name.split('.')
```
goldblatt, line 200: also does python have the like match stuff (I'm thinking in javascript right now), but like a regex match?

libs/pipelines/api_pipeline.py
ghop02: I could load all of the data files, but don't need anything inside them so this was the easiest way to get this dat

libs/pipelines/api_pipeline.py
ghop02: yeah it does - could do it later but this works well enough for right now

PR #300, opened 2020-04-24
Fixing a couple small errors in api output

ghop02:
Fixes a couple small errors in the api output.

 * Ran into a validation error where pandas nan values were sneaking in.
 * Also the change to using __root__ broke the timeseries output it looks like - this fixes that and screams louder if it breaks.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/pipelines/api_pipeline.py
+++ b/libs/pipelines/api_pipeline.py
@@ -227,16 +227,16 @@ def deploy_results(results: List[APIOutput], output: str, write_csv=False):
         output_path.mkdir(parents=True, exist_ok=True)
 
     for api_row in results:
-        dataset_deployer.upload_json(api_row.file_stem, api_row.data.json(), output)
+        dataset_deployer.upload_json(
+            api_row.file_stem, api_row.data.json(), output
+        )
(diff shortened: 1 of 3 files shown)


libs/pipelines/api_pipeline.py
```diff
@@ -227,16 +227,16 @@ def deploy_results(results: List[APIOutput], output: str, write_csv=False):
...
-                if not isinstance(data.get('data'), list):
-                    # Most of the API schemas have the lists under the `'data'` key.
-                    logger.warning(f"Missing data field with list of data.")
-                    continue
+                if not isinstance(data.get('__root__'), list):
+                    raise ValueError("Cannot find list data for csv export.")
                 else:
-                    data = data['data']
+                    data = data['__root__']
             dataset_deployer.write_nested_csv(data, api_row.file_stem, output)
```
goldblatt, line 240: ack sorry i missed this one >.<

ghop02: no you're good I also approved it knowing that i wrote the code like this

PR #304, opened 2020-04-24
Filling in fips and standardizing state name

ghop02:
Standardizes the way we use states:
 - "state" refers to the two letter abbreviation
 - "state_full_name" refers to the full name.

### Please Check
- [ ] Are tests passing?


libs/build_processed_dataset.py
```diff
@@ -111,7 +86,7 @@ def _get_usa_by_county_df():
     # USA only
     us_df = remapped_df[(remapped_df["Country/Region"] == "US")]
     jhu_column_names = [
-        "Province/State",
```
fridiculous, line 114: : )

PR #306, opened 2020-04-24
Fix #258: Support Inference Output for counties

### LINKED ISSUE #258: Support Inference Output for counties
Context: https://github.com/covid-projections/covid-data-model/pull/260.
We should remove `if intervention in Intervention.county_supported_interventions():` once we support counties intervention

mikelehen:
This enables generating OBSERVED_INTERVENTION files in the API for inference projections.

This required tweaking our validation in several places, since we expect many
counties not to have inference results.

This PR addresses issue #258.

### Please Check
- [n/a] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/master/api/schemas) updated?
- [n/a] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [X] Are tests passing?

### DIFF
--- a/cli/run_counties_api.py
+++ b/cli/run_counties_api.py
@@ -43,26 +43,32 @@ def deploy_counties_api(disable_validation, input_dir, output, summary_output):
             raise NotADirectoryError(directory)
 
     for intervention in list(Intervention):
-        # TODO(issues/#258): remove check once counties support inferrence
-        if intervention in Intervention.county_supported_interventions():
-            county_result = api_pipeline.run_projections(
-                input_dir,
-                AggregationLevel.COUNTY,
-                intervention,
-                run_validation=not disable_validation,
-            )
-            county_summaries, county_timeseries = api_pipeline.generate_api(
-                county_result, input_dir
-            )
-            api_pipeline.deploy_results([*county_summaries, *county_timeseries], output)
+        county_result = api_pipeline.run_projections(
+            input_dir,
+            AggregationLevel.COUNTY,
+            intervention,
+            run_validation=not disable_validation,
+        )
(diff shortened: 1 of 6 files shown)


cli/run_counties_api.py
```diff
@@ -43,26 +43,32 @@ def deploy_counties_api(disable_validation, input_dir, output, summary_output):
             raise NotADirectoryError(directory)
 
     for intervention in list(Intervention):
-        # TODO(issues/#258): remove check once counties support inferrence
-        if intervention in Intervention.county_supported_interventions():
```
mikelehen, line 47: None of this code block changed.  I just removed this `if`.

libs/validate_results.py
```diff
@@ -71,10 +72,15 @@ def validate_counties_df(key, counties_df):
     _raise_error_if_incorrect_headers(key, RESULT_DATA_COLUMNS_COUNTIES, counties_df)
 
     # assert data from each of the states
+    expected_missing = EXPECTED_MISSING_STATES.union(
+        EXPECTED_MISSING_STATES_FROM_COUNTES
+    )
+    if key == "counties.OBSERVED_INTERVENTION":
```
ghop02, line 78: I'm not 100% sure where this is called from, but it would be better to pass in the actual enum here.

After looking, it looks like it would be fairly straightforward to pass it in.  Then you could just do:
```if intervention is Intevention.OBSERVED_INTERVENTION:```
Future proofs us for if (when) we end up changing the intervention name

mikelehen: Good suggestion. Done, thanks!

PR #307, opened 2020-04-24
Add icu ventilator usage + fix cumulative tests

ghop02:
Adds ICU/Vent capacity/utilization numbers to state and county timeseries.

Also fixes a bug i introduced where it was pulling incorrect values for cumulative testing data + add a validation to make sure it never happens again.

### Please Check
- [x] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [ ] Are tests passing?

### DIFF
--- a/api/can_api_definition.py
+++ b/api/can_api_definition.py
@@ -110,9 +110,17 @@ class CANPredictionTimeseriesRow(pydantic.BaseModel):
         ...,
         description="Number of ICU beds projected to be in-use or that were actually in use (if in the past)",
     )
-    ICUBedCapacity: Optional[int] = pydantic.Field(
+    ICUBedCapacity: int = pydantic.Field(
         ...,
         description="Number of ICU beds projected to be in-use or actually in use (if in the past)",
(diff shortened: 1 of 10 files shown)


api/can_api_definition.py
```diff
@@ -149,6 +158,27 @@ class PredictionTimeseriesRowWithHeader(CANPredictionTimeseriesRow):
 class CovidActNowStateTimeseries(CovidActNowStateSummary):
     timeseries: List[CANPredictionTimeseriesRow] = pydantic.Field(...)
 
+    # pylint: disable=no-self-argument
+    @pydantic.validator('timeseries')
+    def check_timeseries_have_cumulative_test_data(cls, rows, values):
```
goldblatt, line 163: is this a cls method?

ghop02: yeah all of the validators are classmethods

davidstrauss:
Follow-up for API doc changes here: #312

PR #319, opened 2020-04-26
Refactor Pyseir for Performance

fridiculous:
Gets execute_model to finish <20 min.  Last 8 minutes are just generating output

Can take a list of states.  runs states only in parallel, then runs all counties in parallel

### Please Check
- [x] Are tests passing?


.github/workflows/deploy_api.yml
```diff
@@ -9,15 +9,15 @@ on:
...
+  push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
     types: publish-api
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  COVID_DATA_MODEL_REF: 'fridiculous/parallelize-pyseir-within'
```
erccarls, line 20: reminder to change back.

pyseir/cli.py
```diff
@@ -137,51 +137,138 @@ def _map_outputs(
...
 
 
-def _run_all(
-    state=None,
+def _state_only_pipeline(
+    state,
+    states_only=False,
     run_mode=DEFAULT_RUN_MODE,
-    generate_reports=False,
+    generate_reports=True,
```
erccarls, line 144: This can be false by default.

pyseir/cli.py
```diff
@@ -137,51 +137,138 @@ def _map_outputs(
...
+    # do everything for just states in paralell
+    states_only_func = partial(
+        _state_only_pipeline,
+        run_mode=run_mode,
+        generate_reports=generate_reports,
+        output_interval_days=output_interval_days,
+        states_only=True,
+        output_dir=output_dir,
+    )
+    p = Pool()
```
erccarls, line 201: Can just use a single pool and p.join() between steps until the very end where a single `p.close()` will work.

pyseir/cli.py
```diff
@@ -137,51 +137,138 @@ def _map_outputs(
...
+    #calculate culate inference
+    # parallelize async
+    p = Pool()
+    for state in states:
+        p.apply_async(_infer_rt, state)
+    p.close()
+    p.join()
+
+    #calculate ensemble
+    print(f"executing model for {len(all_county_fips)} counties")
```
erccarls, line 215: Switch to logging instead of print.

pyseir/cli.py
```diff
@@ -137,51 +137,138 @@ def _map_outputs(
...
+    p = Pool()
+    ensemble_func = partial(_run_county, ensemble_kwargs=dict(run_mode=run_mode, generate_report=generate_reports))
+    p.map(ensemble_func, all_county_fips)
+    p.close()
+    p.join()
+
+    #output it all
+    output_interval_days = int(output_interval_days)
+    _cache_global_datasets()
+
```
erccarls, line 232: There is one more step in here which is to generate the pyseir merged output.  I'd like to remove the "webui output mapper at some point since this is just a weird intermediate data model that exists for no reason now.

pyseir/cli.py
```diff
@@ -137,51 +137,138 @@ def _map_outputs(
...
+    output_dir=None,
+    skip_whitelist=False,
+):
     if state:
         # Deprecate temporarily since not needed. Our full model fits have
         # superseded these for now. But we may return to a context where this
         # method is used to measure localized Reff.
         # if not states_only:
         #     _impute_start_dates(state)
-        _infer_rt(state, states_only=states_only)
```
erccarls, line 165: Might suggest we keep both run method until we have verified everything and added new docs.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -283,7 +287,26 @@ def generate_state(self, states_only=False):
             p = Pool()
             p.map(self.map_fips, all_fips)
             p.close()
+            p.join()
 
+    def build_own_fips(self, include_fips):
+        state_fips = us.states.lookup(self.state).fips
+
+        df = load_data.load_county_metadata()
+        all_fips = df[df['state'].str.lower() == self.state.lower()].fips
```
erccarls, line 296: We should just move this list construction to `pyseir/utils.py` so it doesn't have to be repeated.

run.sh
```diff
@@ -61,7 +61,8 @@ execute_model() {
   cd "$(dirname "$0")"
 
   echo ">>> Generating state and county models to ${API_OUTPUT_DIR}"
-  pyseir run-all --run-mode=can-before-hospitalization-new-params --output-dir="${API_OUTPUT_DIR}"
+  # TODO(#148): We need to clean up the output of these scripts!
+  pyseir build-all --run-mode=can-before-hospitalization-new-params --output-dir="${API_OUTPUT_DIR}" > /dev/null
```
erccarls, line 65: We should drop run-mode here since there is already a default elsewhere.

pyseir/inference/model_fitter.py
```diff
@@ -687,31 +714,16 @@ def run_state(state, states_only=False):
...
         all_fips = df_whitelist[df_whitelist['state'].str.lower() == state_obj.name.lower()].fips.values
 
         if len(all_fips) > 0:
             p = Pool()
-            fitters = p.map(ModelFitter.run_for_fips, all_fips)
+            p.map(_execute_model_for_fips, all_fips)
             p.close()
-
-            county_output_file = get_run_artifact_path(all_fips[0], RunArtifact.MLE_FIT_RESULT)
-            pd.DataFrame([fit.fit_results for fit in fitters if fit]).to_json(county_output_file)
```
fridiculous, line 711: here's the issue

pyseir/cli.py
```diff
@@ -137,31 +139,138 @@ def _map_outputs(
...
+            output_interval_days=output_interval_days,
+            run_mode=run_mode,
+            jhu_dataset=nyt_dataset,
+            cds_dataset=cds_dataset,
+            output_dir=output_dir,
+        )
+        web_ui_mapper.generate_state(all_fips=all_county_fips.keys())
+        # mapper_list += build_function_with_fips
+        # web_ui_mapper.build_own_fips(all_county_fips)
+        # web_ui_mapper.execute_own_fips_async(p)
```
ghop02, line 249: remove these comments?

pyseir/inference/fit_results.py
```diff
@@ -41,7 +68,9 @@ def load_inference_result(fips):
         Dictionary of fit result information.
     """
     output_file = get_run_artifact_path(fips, RunArtifact.MLE_FIT_RESULT)
-    df = pd.read_json(output_file, dtype={'fips': 'str'})
+    json_file = open(output_file, 'r')
+    d = simplejson.load(json_file, object_hook=lambda x: convert_fips_to_str(x))
```
ghop02, line 72: Why the switch to this?

pyseir/inference/infer_rt.py
```diff
@@ -513,6 +518,28 @@ def run_state(state, states_only=False):
         rt_inferences = all_fips.map(lambda x: RtInferenceEngine.run_for_fips(x)).tolist()
 
         for fips, rt_inference in zip(all_fips, rt_inferences):
+            # if fips=="16001":
+            #     import ipdb; ipdb.set_trace()
```
ghop02, line 522: remove this comment

pyseir/inference/model_fitter.py
```diff
@@ -680,6 +680,40 @@ def run_for_fips(cls, fips, n_retries=3):
         return model_fitter
 
 
+def _execute_model_for_fips(fips):
```
ghop02, line 683: When are there not fips for this? I would rather handle the missing fips in the calling function if possible

pyseir/inference/model_fitter.py
```diff
@@ -680,6 +680,40 @@ def run_for_fips(cls, fips, n_retries=3):
...
+    if fips:
+        model_fitter = ModelFitter.run_for_fips(fips)
+        return model_fitter
+    return None
+
+def _persist_results_per_state(state_df):
+        county_output_file = get_run_artifact_path(state_df.fips[0], RunArtifact.MLE_FIT_RESULT)
```
ghop02, line 690: indentation

pyseir/inference/model_fitter.py
```diff
@@ -680,6 +680,40 @@ def run_for_fips(cls, fips, n_retries=3):
...
+    """
+    state_obj = us.states.lookup(state)
+    logging.info(f'Get fips list for state {state_obj.name}')
+
+    df_whitelist = load_data.load_whitelist()
+    df_whitelist = df_whitelist[df_whitelist['inference_ok'] == True]
+
+    all_fips = df_whitelist[df_whitelist['state'].str.lower() == state_obj.name.lower()].fips.tolist()
+
+    # county_fips_per_state = dict(zip(state_county_fips, [state]*len(state_county_fips)))
```
ghop02, line 712: remove comment

pyseir/inference/model_fitter.py
```diff
@@ -680,6 +680,40 @@ def run_for_fips(cls, fips, n_retries=3):
...
+            data.to_json(buf)
+
+        for fips, county_series in state_df.iterrows():
+            with open(get_run_artifact_path(fips, RunArtifact.MLE_FIT_MODEL), 'wb') as f:
+                pickle.dump(county_series.mle_model, f)
+
+
+def build_county_list(state):
+    """
+    Build the and return the fips list
```
ghop02, line 702: I think it might be helpful to put a little more context about the whitelist here.  I don't really know a ton about the whitelist and probably wouldn't expect it to be summoned here given the name

pyseir/inference/fit_results.py
```diff
@@ -26,6 +27,32 @@ def load_t0(fips):
...
+    Returns:
+        record {dict} -- {'fips': {'01': '101'},
+                    'R0': {'01': 3.4387109255},
+                    ... }
+
+    """
+    d = {}
+    for col, vdict in record.items():
+        if col=='fips':
+            d.update({col: {k:str(v) for k,v in vdict.items()}})
```
ghop02, line 49: Any reason you're using update here?
`d[col] = {k: str(v) for k, v for vdict.items()}` seems a bit more straightforward

pyseir/cli.py
fridiculous: don't we use webui thou?

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -227,7 +235,26 @@ def generate_state(self, states_only=False):
             p = Pool()
             p.map(self.map_fips, all_fips)
             p.close()
+            p.join()
+
+    def build_own_fips(self, include_fips):
```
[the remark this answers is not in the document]
fridiculous: removing.  there was a fair amount of bandaid solutioning around getting around the web_ui class

pyseir/inference/fit_results.py
fridiculous: lots of i/o, some performance benefit

pyseir/inference/infer_rt.py
```diff
@@ -80,21 +80,25 @@ def __init__(self,
...
+                else:
+                    self.display_name = self.state
+
+                # TODO Swap for new data source.
+                self.times, self.observed_new_cases, self.observed_new_deaths = \
+                    load_data.load_new_case_data_by_fips(self.fips, t0=self.ref_date)
+                self.hospital_times, self.hospitalizations, self.hospitalization_data_type = \
+                    load_data.load_hospitalization_data(self.fips, t0=self.ref_date)
+            except Exception:
+                logging.warn(f'Failed to init inference for {self.display_name}')
```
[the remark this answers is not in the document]
fridiculous: This happens a lot, and it's not clear why.  there probably more performance improvements simply because fips is being passed through this unnecessarily

pyseir/inference/model_fitter.py
fridiculous: There is some weird fips cascading logic where sometimes some counties don't show up.  similiar to the other case.

pyseir/inference/model_fitter.py
fridiculous: Actually, i'm not exactly clear on how the whitelist is working either, but noted cc @erccarls

pyseir/inference/infer_rt.py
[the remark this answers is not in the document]
ghop02: I guess I'm not sure that catching the exception here is the right place for it - we'll be failing silently (I think `return None` is a no-op) and I'm not sure what actually causes the failure (all of this code looks relatively benign).  

My only thought is that if it's spawning a ton of processes and everyone's trying to run `load_data.load_county_metadata()` for the first time maybe it's an issue? 

I guess I would much rather catch the exception in the calling code because I don't deterministically know what's going to happen here.

pyseir/load_data.py
```diff
@@ -296,6 +298,22 @@ def load_county_metadata_by_fips(fips):
     return county_metadata_merged
 
 
+
+def get_all_fips_codes_for_a_state(state: str):
+        """Returns a list of fips codes for a state
```
ghop02, line 303: This is indented a bit too far

pyseir/inference/infer_rt.py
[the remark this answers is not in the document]
fridiculous: it's happening because the wrong fips are sent during the county fips generation, either cause the data isn't sufficient, or the fips whitelisted at earlier step but not filtered at a subsequent step.

pyseir/cli.py
```diff
@@ -137,31 +139,138 @@ def _map_outputs(
...
-    state=None,
+def _state_only_pipeline(
+    state,
+    states_only=True,
     run_mode=DEFAULT_RUN_MODE,
     generate_reports=False,
     output_interval_days=1,
+    output_dir=None,
+):
+    states_only = True
```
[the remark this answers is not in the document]
erccarls: seconded!

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -93,6 +93,13 @@ def map_fips(self, fips):
         else:
             population = self.population_data.get_state_level('USA', state=self.state_abbreviation)
 
+        # logging.info(f'Mapping output to WebUI for {self.state}, {fips}')
+        # pyseir_outputs = load_data.load_ensemble_results(fips)
```
erccarls, line 97: Isn't this needed?

pyseir/inference/fit_results.py
erccarls: Does this truncate leading zeros if it is being read in as an int?  These are really tiny files so I would argue the complexity of this is not worth it.

pyseir/inference/infer_rt.py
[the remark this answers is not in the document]
erccarls: I also don't like this switch. It could be indicating problems like the leading zeros one.  Can we undo and just solve the underlying problem?

pyseir/inference/infer_rt.py
```diff
@@ -502,17 +507,39 @@ def run_state(state, states_only=False):
     state_obj = us.states.lookup(state)
     df = RtInferenceEngine.run_for_fips(state_obj.fips)
     output_path = get_run_artifact_path(state_obj.fips, RunArtifact.RT_INFERENCE_RESULT)
-    df.to_json(output_path)
+    with open(output_path, 'w') as buf:
```
erccarls, line 510: Why this? Pandas already does this for you.

pyseir/inference/infer_rt.py
```diff
@@ -502,17 +507,39 @@ def run_state(state, states_only=False):
...
-        all_fips = df[df['state'].str.lower() == state_obj.name.lower()].fips
+        all_fips = load_data.get_all_fips_codes_for_a_state(state)
 
         # Something in here doesn't like multiprocessing...
         rt_inferences = all_fips.map(lambda x: RtInferenceEngine.run_for_fips(x)).tolist()
 
         for fips, rt_inference in zip(all_fips, rt_inferences):
             county_output_file = get_run_artifact_path(fips, RunArtifact.RT_INFERENCE_RESULT)
             if rt_inference is not None:
-                rt_inference.to_json(county_output_file)
```
erccarls, line 518: Same comment. No benefit to doing this.

pyseir/inference/infer_rt.py
```diff
@@ -502,17 +507,39 @@ def run_state(state, states_only=False):
...
+    states_only: bool
+        If True only run the state level.
+    """
+    if not fips:
+        return None
+
+    df = RtInferenceEngine.run_for_fips(fips)
+    county_output_file = get_run_artifact_path(fips, RunArtifact.RT_INFERENCE_RESULT)
+    if df is not None and not df.empty:
+        with open(county_output_file, 'w') as buf:
```
erccarls, line 544: Same here.

pyseir/inference/model_fitter.py
erccarls: Are you sure this isn't because of the leading zeros issue?

pyseir/inference/model_fitter.py
```diff
@@ -680,6 +680,38 @@ def run_for_fips(cls, fips, n_retries=3):
...
+def _execute_model_for_fips(fips):
+    if fips:
+        model_fitter = ModelFitter.run_for_fips(fips)
+        return model_fitter
+    return None
+
+def _persist_results_per_state(state_df):
+    county_output_file = get_run_artifact_path(state_df.fips[0], RunArtifact.MLE_FIT_RESULT)
+    data = state_df.drop(['state', 'mle_model'], axis=1)
+    with open(county_output_file, 'w') as buf:
```
erccarls, line 692: just use `to_json(path)`

pyseir/inference/model_fitter.py
erccarls: Whitelist is generated in `inference/whitelist_generator.py`. It is a straightforward cut on a few things like min total cases, min total deaths and how many "new XXX" datapoints exist above a threshold.

The reason it is called here is because we do not want to run inference over those counties with very little data.

pyseir/inference/model_fitter.py
```diff
@@ -700,13 +732,18 @@ def run_state(state, states_only=False):
     df_whitelist = df_whitelist[df_whitelist['inference_ok'] == True]
 
     output_path = get_run_artifact_path(state_obj.fips, RunArtifact.MLE_FIT_RESULT)
-    pd.DataFrame(model_fitter.fit_results, index=[state_obj.fips]).to_json(output_path)
+    data = pd.DataFrame(model_fitter.fit_results, index=[state_obj.fips])
+    with open(output_path, 'w') as buf:
+        data.to_json(buf)
```
erccarls, line 737: Was ok as is.

pyseir/inference/model_fitter.py
erccarls: What do you mean by issue?

pyseir/inference/whitelist_generator.py
```diff
@@ -5,7 +5,7 @@
 from datetime import datetime
 from pyseir.utils import get_run_artifact_path, RunArtifact
 from pandarallel import pandarallel
-pandarallel.initialize(progress_bar=True)
+pandarallel.initialize(progress_bar=False)
```
erccarls, line 8: Read from env? lie `PYSEIR_VERBOSITY`?  I like the progress bars on local.

pyseir/inference/whitelist_generator.py
```diff
@@ -68,7 +68,8 @@ def generate_whitelist(self):
         output_path = get_run_artifact_path(
             fips='06', # Dummy fips since not used here...
             artifact=RunArtifact.WHITELIST_RESULT)
-        df_whitelist.to_json(output_path)
```
erccarls, line 71: Revert

pyseir/load_data.py
```diff
@@ -296,6 +298,22 @@ def load_county_metadata_by_fips(fips):
     return county_metadata_merged
 
 
+
+def get_all_fips_codes_for_a_state(state: str):
```
erccarls, line 302: This might be better in `utils.py`

erccarls: And you could also add an `@lru_cache()` decorator.

pyseir/inference/whitelist_generator.py
fridiculous: good to know

pyseir/load_data.py
fridiculous: the main reason i kept it in load_data, is because it's already colocated with load_county_metadata.  I figured some of these helper methods may make sense to re-arrange in a different pr.

pyseir/inference/model_fitter.py
fridiculous: it was a note to myself when i was working through the logic

pyseir/inference/model_fitter.py
fridiculous: yes

pyseir/inference/fit_results.py
fridiculous: ok, ill revert this

pyseir/inference/infer_rt.py
fridiculous: Slight i/o improvement.  The less pandas has to infer, the better

pyseir/deployment/webui_data_adaptor_v1.py
fridiculous: it's not clear to me if it is tbh

ghop02: Resolution on this?

pyseir/inference/infer_rt.py
```diff
@@ -511,8 +516,7 @@ def run_state(state, states_only=False):
 
     # Run the counties.
     if not states_only:
-        df = load_data.load_county_metadata()
-        all_fips = df[df['state'].str.lower() == state_obj.name.lower()].fips
+        all_fips = load_data.get_all_fips_codes_for_a_state(state)
```
ghop02, line 519: 😍

pyseir/inference/infer_rt.py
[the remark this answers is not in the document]
ghop02: Still would like to not catch the exception here.  At the very least, can we wrap where this exception is being raised in a try catch and add sentry reporting?

pyseir/inference/model_fitter.py
```diff
@@ -697,6 +697,37 @@ def run_for_fips(cls, fips, n_retries=3, with_age_structure=False):
         return model_fitter
 
 
+def _execute_model_for_fips(fips):
+    if fips:
+        model_fitter = ModelFitter.run_for_fips(fips)
+        return model_fitter
+    return None
```
ghop02, line 704: Would be good to raise an alert when this happens, I'd like for this failure to not be silent

pyseir/load_data.py
```diff
@@ -301,6 +303,21 @@ def load_county_metadata_by_fips(fips):
...
+    """Returns a list of fips codes for a state
+
+    Arguments:
+        state {str} -- the full state name
+
+    Returns:
+        fips [list] -- a list of fips codes for a state
+    """
+    df = load_county_metadata()
```
ghop02, line 316: Can we use `libs.datasets.combined_datasets.build_us_latest_with_all_fields()` for this so that the fips codes match the rest of the input data we have?

pyseir/cli.py
```diff
@@ -137,31 +139,137 @@ def _map_outputs(
...
+    p.map(ensemble_func, all_county_fips.keys())
+
+    #output it all
+    output_interval_days = int(output_interval_days)
+    _cache_global_datasets()
+
+    root.info(f"outputing web results for states and {len(all_county_fips)} counties")
+    # does not parallelize well, because web_ui mapper doesn't serialize efficiently
+    # TODO: Remove intermediate artifacts and paralellize artifacts creation better
+    # Approximately 40% of the processing time is taken on this step
```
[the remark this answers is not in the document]
fridiculous: beleive it or not, i tried it, but i had to revert back because doubles weren't supported.

pyseir/inference/model_fitter.py
fridiculous: adding a warning

pyseir/load_data.py
fridiculous: could we do this in a separate pr?

pyseir/inference/infer_rt.py
fridiculous: addressed

pyseir/load_data.py
ghop02: sounds good!

pyseir/cli.py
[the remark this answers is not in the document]
ghop02: poke

PR #325, opened 2020-04-27
Model/update future scenarios to inference base

erccarls:
Finally, a PR that removes more code than it adds.. 

This PR addresses a bunch of things and greatly simplifies our overall model pipeline. 

- Swap all interventions to start with the inferred model and then update the future intervention.
- Future interventions ramp up to a final value over 2 weeks rather than immediate starts. 
- Increase available inferred counties from 300 -> 1500, by relaxing whitelist and having a tiered inference which fixes more parameters to state values as data decreases. This results in <5% bad fit rate.
- Better initial guesses for inference
- Reduce the error of the zero-inflated poisson process. This adds great stability.
- Increases errors on anomalously low case data. We were already doing this for deaths.

Because of the swap to inference we get to delete backfills and hospitalization tunings and consolidate them in a single place. We can also remove several code paths to deal with the different procedures for initial conditions generation (since these are now all derived from the fit).


pyseir/ensembles/ensemble_runner.py
```diff
@@ -175,88 +136,14 @@ def init_run_mode(self):
         """
         self.suppression_policies = dict()
 
-        if self.run_mode is RunMode.CAN_BEFORE_HOSPITALIZATION:
```
erccarls, line 178: These old scenarios are no longer needed.

pyseir/inference/model_fitter.py
```diff
@@ -78,31 +79,9 @@ class ModelFitter:
     )
 
     PARAM_SETS = {
-        ('KY', 'KS', 'ID', 'NE', 'VA', 'RI'):  dict(
```
erccarls, line 81: Now that we have more data, these fit guesses and limits can be relaxed

pyseir/inference/whitelist_generator.py
```diff
@@ -26,10 +27,10 @@ class WhitelistGenerator:
     """
     def __init__(
             self,
-            total_cases=100,
-            total_deaths=10,
-            nonzero_case_datapoints=20,
-            nonzero_death_datapoints=2):
+            total_cases=20,
```
erccarls, line 30: Relaxing whitelist significantly.

pyseir/load_data.py
```diff
@@ -108,23 +108,6 @@ def cache_county_case_data():
     county_case_data.to_pickle(os.path.join(DATA_DIR, 'covid_case_timeseries.pkl'))
 
 
-def cache_hospital_beds():
```
erccarls, line 111: No longer used after swapping to CAN data sources.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -50,58 +50,6 @@ def __init__(self, state, output_interval_days=4, run_mode='can-before',
         self.df_whitelist = load_data.load_whitelist()
         self.df_whitelist = self.df_whitelist[self.df_whitelist['inference_ok'] == True]
 
-    def backfill_output_model_fips(self, fips, t0, final_beds, output_model):
```
erccarls, line 53: Backfills no longer needed since we are using models

pyseir/inference/model_fitter.py
```diff
@@ -266,9 +280,10 @@ def calculate_observation_errors(self):
         else:
             hosp_stdev = None
 
-        # Zero inflated poisson Avoid floating point errors..
-        cases_stdev[cases_stdev == 0] = 1e10
-        deaths_stdev[deaths_stdev == 0] = 1e10
+        # Zero inflated poisson Avoid floating point errors. This is set to a
+        # value that still provides some constraints toward zero.
```
macklin, line 284: Minor: I don't understand the `Avoid floating point errors` statement, but I don't have a ton of context, and it is by no means a blocker

pyseir/inference/model_fitter.py
```diff
@@ -182,6 +152,46 @@ def __init__(self,
...
+
+        if len(self.fips) == 5:
+            t0_guess = list(self.times)[np.argwhere(np.cumsum(self.observed_new_cases) >= 3)[0][0]]
+
+            state_fit_result = load_inference_result(fips=self.state_obj.fips)
+
+            self.fit_params['t0'] = t0_guess # state_fit_result['t0']
+
+            total_cases = np.sum(self.observed_new_cases)
+            self.fit_params['log10_I_initial'] = np.log10(3 / self.fit_params['test_fraction'])
```
zhang-xinyu, line 179: Instead of 3, should this be cumulative observed new cases at t0_guess?

pyseir/load_data.py
```diff
@@ -563,10 +553,37 @@ def cache_all_data():
...
+    """
+    Return the value of compartment at a specified date.
+
+    Parameters
+    ----------
+    fips: str
+        State or County fips.
+    compartment: str
+        Name of the compartment to retrieve.
+    date: datetime
```
macklin, line 570: Add description

pyseir/models/suppression_policies.py
```diff
@@ -22,6 +22,31 @@
...
+    """
+    Returns the future suppression level for a given R0 and a "future scenario".
+
+    Parameters
+    ----------
+    R0:float
+        Reproduction number
+    scenario: str
+        'no_intervention', 'flatten_the_curve', 'social_distancing'.
+    Returns
```
macklin, line 35: Minor: Add empty line before `Returns`

pyseir/inference/model_fitter.py
```diff
@@ -266,9 +280,10 @@ def calculate_observation_errors(self):
...
 
-        # Zero inflated poisson Avoid floating point errors..
-        cases_stdev[cases_stdev == 0] = 1e10
-        deaths_stdev[deaths_stdev == 0] = 1e10
+        # Zero inflated poisson Avoid floating point errors. This is set to a
+        # value that still provides some constraints toward zero.
+        cases_stdev[cases_stdev == 0] = 1e2
+        deaths_stdev[deaths_stdev == 0] = 1e2
         if hosp_stdev is not None:
             hosp_stdev[hosp_stdev == 0] = 1e10
```
zhang-xinyu, line 288: Does hosp_stdev also need to be updated?

pyseir/models/suppression_policies.py
```diff
@@ -22,6 +22,31 @@
...
+def get_future_supression_from_r0(R0, scenario):
+    """
+    Returns the future suppression level for a given R0 and a "future scenario".
+
+    Parameters
+    ----------
+    R0:float
+        Reproduction number
+    scenario: str
+        'no_intervention', 'flatten_the_curve', 'social_distancing'.
```
macklin, line 34: Minor: If you have an enum for this, you could use it here

pyseir/models/suppression_policies.py
```diff
@@ -22,6 +22,31 @@
...
+    -------
+    epsilon: float
+        Suppression percentage.
+    """
+    if scenario == 'no_intervention':
+        return 1
+    elif scenario == 'flatten_the_curve':
+        return 0.97 / R0
+    elif scenario == 'social_distancing':
+        return 1.7 / R0
```
macklin, line 45: Optional: Explain where these constants came from

pyseir/models/suppression_policies.py
```diff
@@ -170,16 +195,30 @@ def generate_two_step_policy(t_list, eps, t_break, transition_time=14):
         Suppression level after t_break
     t_break: float
         Time since simulation start to place a break.
+    transition_time: float
+        Length of time to transition between epsilon states.
+    t_break_final: int
```
macklin, line 200:
```suggestion
    t_break_final: int or NoneType
```

pyseir/models/suppression_policies.py
```diff
@@ -170,16 +195,30 @@ def generate_two_step_policy(t_list, eps, t_break, transition_time=14):
...
+        Time since simulation start to place a break to a final level.
+    eps_final: float
```
macklin, line 202:
```suggestion
    eps_final: float or NoneType
```

pyseir/models/suppression_policies.py
```diff
@@ -170,16 +195,30 @@ def generate_two_step_policy(t_list, eps, t_break, transition_time=14):
...
+        Suppression level after t_break_final.
```
macklin, line 203: Optional: explain behavior when `eps_final=None`

pyseir/load_data.py
zhang-xinyu, line 570: missing description here

pyseir/models/suppression_policies.py
zhang-xinyu, line 34: would be nice to add enum for this.

pyseir/models/suppression_policies.py
zhang-xinyu, line 45: It is unclear to me how value for social distancing was chosen ?

pyseir/models/suppression_policies.py
```diff
@@ -170,16 +195,30 @@ def generate_two_step_policy(t_list, eps, t_break, transition_time=14):
...
-        fill_value='extrapolate')
+    if eps_final is None:
+        return interp1d(
+            x=[0, t_break, t_break + transition_time, 100000],
+            y=[1, 1, eps, eps],
+            fill_value='extrapolate'
+        )
+    else:
+        return interp1d(
+            x=[0, t_break, t_break + transition_time, t_break_final, t_break_final + transition_time, 100000],
```
zhang-xinyu, line 218: Maybe confirm t_break_final > t_break + transition_time here

pyseir/inference/model_fitter.py
erccarls: old comment. cleaned up.

pyseir/models/suppression_policies.py
erccarls: Yeah. Expected to change, so never bothered, but we should once we decide on reopening nomenclature.

pyseir/models/suppression_policies.py
erccarls: Someone else's brain... We have refs somewhere, but these are then hand tuned for product purposes.

pyseir/inference/model_fitter.py
erccarls: I'll add a comment. The problem is that this is e.g. 20x observed cases, which is fairly late in the simulation, so I found it much better to fit using these parameters after a lot of trial and error.

pyseir/inference/model_fitter.py
erccarls: 👍

pyseir/models/suppression_policies.py
erccarls: agree. they are about to change so will do in a broader cleanup.

pyseir/models/suppression_policies.py
erccarls: Same :)... They were largely from some early studies max found in Wuhan.  We will be updating once we have some better reopening scenarios.  There are refs in the modeling doc.

PR #328, opened 2020-04-27
Fix hospitalization rate in webui output

erccarls:
After parameter updates, this backfill parameter has not been updated, but impacts those locations which do not have hospitalization data.

### DIFF
--- a/pyseir/deployment/webui_data_adaptor_v1.py
+++ b/pyseir/deployment/webui_data_adaptor_v1.py
@@ -71,7 +71,7 @@ class WebUIDataAdaptorV1:
             Backfill dataframe.
         """
         backfill_to_date = date(2020, 3, 3)   # @TODO: Parameterize
-        hospitalization_rate = 0.04           # @TODO: Parameterize
+        hospitalization_rate = 0.02           # @TODO: Parameterize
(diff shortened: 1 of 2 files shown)


pyseir/models/suppression_policies.py
```diff
@@ -153,7 +153,7 @@ def get_interventions(start_date=datetime.now().date()):
         else:
             raise ValueError(f'Invalid scenario {scenario}')
 
-    return lambda x: np.interp(x, t_list, rho)
+    return interp1d(t_list, rho, fill_value='extrapolate')
```
erccarls, line 156: Also reverting these since lambdas don't pickle when using multiprocessing.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -71,7 +71,7 @@ def backfill_output_model_fips(self, fips, t0, final_beds, output_model):
             Backfill dataframe.
         """
         backfill_to_date = date(2020, 3, 3)   # @TODO: Parameterize
-        hospitalization_rate = 0.04           # @TODO: Parameterize
+        hospitalization_rate = 0.02           # @TODO: Parameterize
```
erccarls, line 74: Not parametrizing because this is going away today or tomorrow when we swap to inference models.

PR #352, opened 2020-04-30
Add nevada actuals to our covid tracking data

ghop02:
Adds in a data source for numbers we're manually entering from nevada and then updates values in covid tracking with those nevada numbers.

### DIFF
--- a/libs/datasets/custom_aggregations.py
+++ b/libs/datasets/custom_aggregations.py
@@ -39,11 +39,11 @@ def calculate_combined_new_york_counties(
     if are_boroughs_zero:
         grouped = non_ny_county.groupby(group).sum()
         for column in grouped.columns:
-            assert sum(grouped[column]) == 0, f"{column} is unexpectedly zero."
+            assert sum(grouped[column]) == 0, f"{column} is unexpectedly not zero."
     else:
         grouped = non_ny_county.groupby(group).sum()
         for column in grouped.columns:
-            assert sum(grouped[column]) != 0, f"{column} is unexpectedly not zero."
+            assert sum(grouped[column]) != 0, f"{column} is unexpectedly zero."
 
     aggregated = new_york_county.groupby(group).sum().reset_index()
     aggregated["fips"] = NEW_YORK_COUNTY_FIPS
@@ -59,9 +59,15 @@ def update_with_combined_new_york_counties(
     """Updates data replacing all new york county data with one number.
 
     """
-    if not sum(data.aggregate_level == AggregationLevel.COUNTY.value):
+    is_county = data.aggregate_level == AggregationLevel.COUNTY.value
+    if not sum(is_county):
         # No county level data, skipping county aggregation.
         return data
+
+    if not len(data[(data.state == 'NY') & is_county]):
+        # No NY data, don't apply aggregation.
+        return data
+
     data = data.set_index(["aggregate_level"])
     county = data.loc[AggregationLevel.COUNTY.value].reset_index()
     county = calculate_combined_new_york_counties(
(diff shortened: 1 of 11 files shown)


ghop02:
Need to merge this first: https://github.com/covid-projections/covid-data-public/pull/37

libs/datasets/sources/covid_tracking_source.py
```diff
@@ -100,7 +101,7 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
         data[cls.Fields.AGGREGATE_LEVEL] = AggregationLevel.STATE.value
         # Date checked is the time that the data is actually updated.
         # assigning the date field as the date floor of that day.
-        data[cls.Fields.DATE] = data[cls.Fields.DATE_CHECKED].dt.floor("D")
+        data[cls.Fields.DATE] = data[cls.Fields.DATE_CHECKED].dt.tz_localize(None).dt.floor("D")
```
fridiculous, line 104: why is it important to localize?

libs/datasets/custom_aggregations.py
```diff
@@ -39,11 +39,11 @@ def calculate_combined_new_york_counties(
     if are_boroughs_zero:
         grouped = non_ny_county.groupby(group).sum()
         for column in grouped.columns:
-            assert sum(grouped[column]) == 0, f"{column} is unexpectedly zero."
+            assert sum(grouped[column]) == 0, f"{column} is unexpectedly not zero."
     else:
         grouped = non_ny_county.groupby(group).sum()
         for column in grouped.columns:
-            assert sum(grouped[column]) != 0, f"{column} is unexpectedly not zero."
+            assert sum(grouped[column]) != 0, f"{column} is unexpectedly zero."
```
davidstrauss, line 46: For assertions, I try to avoid specifying the condition in the message so there isn't risk of desync between the message and *actual* condition. (Hardly a blocker, though, and fixing the existing message is still a nice fix.)

libs/datasets/sources/covid_tracking_source.py
davidstrauss, line 104: `tz_localize(None)` seems to just strip the time in a way that removes timezone so that the time is presumed local (for any definition of local).

What do you think about using `.normalize()` (possibly `.normalize().date()`)  instead of stripping localization and doing a `floor("D")`?

libs/datasets/sources/covid_tracking_source.py
```diff
@@ -131,6 +132,62 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
...
+
+    @classmethod
+    def _load_nevada_override_data(cls):
+        from libs.datasets import NevadaHospitalAssociationData
+
+        data = NevadaHospitalAssociationData.local().timeseries(fill_na=False).data
+        columns_to_include = []
+        for timeseries_column, ct_column in cls.TIMESERIES_FIELD_MAP.items():
+            if timeseries_column in data.columns:
+                columns_to_include.append(ct_column)
```
davidstrauss, line 150: Does this mean that NHA data will only replace Covid Tracking data for the dates we have NHA data for? Will that cause any inconsistency day-to-day?

libs/datasets/sources/covid_tracking_source.py
ghop02: Ah because we don't use timezones in other dates, so when comparing indexes between different data sources, it doesn't work properly if the timezones are in different formats

ghop02: I like the idea of it, but we actually have to have this be non timezoned times (none of the other data sources have time zones, so this having a timezone actually breaks comparison with dates from other data sources in some operations).

libs/datasets/sources/covid_tracking_source.py
ghop02: Correct.  

For right now the model seems to be using the latest values, so this is actually okay as long as covid tracking doesn't start reporting nv data.  Probably worth an issue to make sure this doesn't bite us in the future though.

libs/datasets/custom_aggregations.py
ghop02: agreed - I generally prefer to not do assertions actually, then you can better more contextual python error handling, but probably will just leave this be for now!

PR #353, opened 2020-04-30
Adding weighted average of bed utilization rates for nyc borough

ghop02:
This is a bit hacky - would eventually like to put some logic to pass in which columns can do weighted averages, but too much complexity for right now.

Just computes the weighted average for the occupancy rates in bed data

### DIFF
--- a/libs/datasets/beds.py
+++ b/libs/datasets/beds.py
@@ -92,12 +92,7 @@ class BedsDataset(object):
         columns_to_consider = [cls.Fields.STAFFED_BEDS, cls.Fields.LICENSED_BEDS]
         data[cls.Fields.MAX_BED_COUNT] = data[columns_to_consider].max(axis=1)
 
-        # When grouping nyc data, we don't want to count the generated field
-        # as a value to sum.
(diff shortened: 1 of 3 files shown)


libs/datasets/beds.py
```diff
@@ -114,6 +109,47 @@ def from_source(cls, source: "DataSource", fill_missing_state=True):
...
+        if not len(nyc_data):
+            return data
+        group = cls.STATE_GROUP_KEY + [cls.Fields.GENERATED]
+        weighted_all_bed_occupancy = None
+
+        if cls.Fields.ALL_BED_TYPICAL_OCCUPANCY_RATE in data.columns:
+            licensed_beds = nyc_data[cls.Fields.LICENSED_BEDS]
+            occupancy_rates = nyc_data[cls.Fields.ALL_BED_TYPICAL_OCCUPANCY_RATE]
+            weighted_all_bed_occupancy = (
+                (licensed_beds * occupancy_rates).sum() / licensed_beds.sum()
```
fridiculous, line 126: yea!

PR #355, opened 2020-04-30
Align rt values

erccarls:
This PR adds a time-lag alignment feature when computing R_t, which brings down our previous estimate of R_t down by a bit for some states, and quite a lot for others where death was a significantly lagging indicator relative to cases). Likely this means that cases in some localities are being processed and reported much more efficiently. 

The alignment procedure finds the time shift that maximizes correlation of the time series curvature, i.e. we fit the shape of the curve.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -383,6 +383,8 @@ class RtInferenceEngine:
                         series_a=df_all[f'Rt_MAP__{TimeseriesType.NEW_CASES.value}'].iloc[-last_idx:],
                         series_b=df_all[f'Rt_MAP__{timeseries_type.value}'].iloc[-last_idx:]
                     )
+                    df_all[f'lag_days__{timeseries_type.value}'] = shift_in_days
+
                     # Shift all the columns.
                     for col in df_all.columns:
                         if timeseries_type.value in col:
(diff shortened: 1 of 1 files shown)


pyseir/inference/infer_rt.py
```diff
@@ -441,15 +460,20 @@ def align_time_series(series_a, series_b):
...
+        valid_shifts = []
         xcor = []
         for i in shifts:
-            series_a = np.diff(series_a)
-            series_b = np.diff(series_b.shift(i))
-            valid = (~np.isnan(series_a) & ~np.isnan(series_b))
-            xcor.append(signal.correlate(series_a[valid], series_b[valid]).mean())
-
-        return shifts[np.argmax(xcor)]
+            _series_a = np.diff(series_a)
```
zhang-xinyu, line 467: Seems this can be calculated just once before the loop.

pyseir/inference/infer_rt.py
```diff
@@ -441,15 +460,20 @@ def align_time_series(series_a, series_b):
         shift: int
             A shift period applied to series b that aligns to series a
         """
-        shifts = range(-20, 20)
+        shifts = range(-30, 5)
```
zhang-xinyu, line 463: Maybe increase the upper limit to increase recall.

PR #358, opened 2020-04-30
Pyseir/demographic mapper

zhang-xinyu:

### Please Check
- [ ] Are tests passing?


pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,592 @@
...
+        MLE epi parameters and associated errors. The parameter used is
+        t0_date.
+    measures: str or list(str)
+        Names of covid measures, should be interpretable by CovidMeasure.
+    measure_units: str or list(str)
+        Units of covid measures, should be interpretable by CovidMeasureUnit.
+    target_age_distribution: callable
+        PDF of age of target population.
+    risk_modifier_by_age: callable
+        Returns risk rations by age group that modifies the risk of
```
erccarls, line 137:
```suggestion
        Returns risk ratios by age group that modifies the risk of
```

erccarls: I'm not sure I really understand what the inputs and outputs of this distribution are.

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,592 @@
...
+            target_age_distribution = lambda x: np.array([1] * len(self.parameters['age_groups']))
+        self.target_age_distribution = target_age_distribution
+        self.risk_modifier_by_age = risk_modifier_by_age
+
+        # get parameters required to calculate covid measures
+        self.hospitalization_rates, self.mortality_rates = self._generate_hospitalization_mortality_rates()
+        self.prevalence = self._age_specific_prevalence()
+
+        self.mapped_predictions = None
+
```
erccarls, line 171: nit: Extra line here

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,592 @@
...
+        return prevalence
+
+    def _reconstruct_mortality_rates_trajectory(self):
+        """
+        Reconstruct trajectory of inferred per-capita mortality rates through
+        time from MLE model parameters and compartments.
+
+        Returns
+        -------
+        mortality_rate_general: np.array
```
erccarls, line 194: Since this has led to confusion, why don't we call this `mortality_inflow_rate`

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,592 @@
...
+           new hospitalization / (asymptomatic + symptomatic infections)
+
+        Parameters
+        ----------
+        measure_unit: CovidMeasureUnit
+            Unit of covid measure
+
+        Returns
+        -------
+          : np.array
```
erccarls, line 292: Would be nice to label these outputs since there are multiple fo them,

pyseir/deployment/demographic_mapper.py
zhang-xinyu: The input is an array of age and output is the pdf based on target age distribution. I have updated the description and the variable's name.

pyseir/deployment/demographic_mapper.py
zhang-xinyu: Good idea!

pyseir/deployment/demographic_mapper.py
zhang-xinyu: Agreed, now all output variables are put in a single dictionary.

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+        """
+        Calculate age specific hospitalization rates among infected
+        population (IHR) for a given measure unit by types of
+        hospitalization.
+
+        When unit is per capita, calculates the probability that an infected
+        person (asymptomatic or symptomatic) ultimately gets admitted to
+        hospital:
+            probability of being symptomatic * rate of hospitalized / (rate of
+            hospitalized + rate of recovery without hospitalization)
```
macklin, line 336: Apologies if I'm missing something, but I don't see this being computed in the code

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+
+        elif measure is CovidMeasure.HR_ICU:
+            # prevalence is used to count the probability that a person is
+            # infected
+            return (self.IHR['HICU'][measure_unit.value]
+                  + self.IHR['HVent'][measure_unit.value]) * self.prevalence
+
+        else:
+            logging.warning(f'covid_measure {measure.value} is not relevant to hospitalization rate')
+            return None
```
macklin, line 537: I would raise a `ValueError` here, but that's up to you

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+        -------
+        IFR: np.array
+            Array of time series of age-specific infection fatality rate.
+        """
+
+        IFR = 0
+        for key in self.IFR:
+            IFR += self.IFR[key][measure_unit.value]
+
+        return IFR
```
macklin, line 558: I think you can do:
```suggestion
        return sum(self.IFR[key][measure_unit.value] for key in self.IFR)
```

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+    def generate_predictions(self):
+        """
+        Generates predictions of covid measures using the MLE model.
+
+        Returns
+        -------
+        predictions: dict
+            Contains:
+            - compartments:
+              - <compartment>: time series of population at a specific
```
macklin, line 570:
```suggestion
              - <compartment>: time series of population at specific
```

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+                               infection states (susceptible, infected,
+                               hospitalized, etc.) by demographic group
+                               simulated by MLE model. Each time series is
+                               recorded as pd.DataFrame, with dates of
+                               prediction as index.
+              name of compartment include: S - susceptible, E - exposed,
```
macklin, line 576:
```suggestion
              name of compartments include: S - susceptible, E - exposed,
```

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+    fit_results: dict
+        MLE epi parameters and associated errors. The parameter used is
+        t0_date.
+    measures: str or list(str)
+        Names of covid measures, should be interpretable by CovidMeasure.
+    measure_units: str or list(str)
+        Units of covid measures, should be interpretable by CovidMeasureUnit.
+    target_age_distribution_pdf: callable
+        Function that takes an array of age as input returns PDF of the age
+        based on the age distribution of target population.
```
macklin, line 161: This description could be fleshed out a little more. For example, what age ranges need to be handled, and at what granularity? Are there certain age bins that are expected?

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+                                with dates of prediction as index.
+        """
+        # calculate weights
+        age_bin_centers = [np.mean(tup) for tup in self.parameters['age_groups']]
+        weights = self.target_age_distribution_pdf(age_bin_centers)
+        if (weights != 1).sum() == 0:
+            logging.warning('no target age distribution is given, measure is aggregated assuming age '
+                            'distrubtion at given FIPS code')
+
+        weights /= weights.sum()
```
macklin, line 674: If `self.target_age_distribution_pdf` is returning a `pdf`, shouldn't this have already been done inside that method?

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+        demographic_group_size_ratio = weights / (self.parameters['N'] / self.parameters['N'].sum())
+
+        mapped_predictions = defaultdict(dict)
+
+        for c in predictions['compartments']:
+            mapped_predictions['compartments'][c] = predictions['compartments'][c].dot(demographic_group_size_ratio)
```
macklin, line 680: Sorry if I'm missing context, but why are you using `dot` rather than `*` (elementwise multiplication)? What's the reason to return a scalar to store in `mapped_predictions['compartments'][c]`

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,729 @@
...
+import pandas as pd
+import numpy as np
+from collections import defaultdict
+from enum import Enum
+from datetime import datetime, timedelta
+
+
+class CovidMeasure(Enum):
+    """
+    - IHR_infected: probability an infected person is hospitalized,
```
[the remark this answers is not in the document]
zhang-xinyu: Good catch! This should be just IHR.

pyseir/deployment/demographic_mapper.py
zhang-xinyu: Ah the conditions should be switched here. Thanks for pointing this out!

pyseir/deployment/demographic_mapper.py
zhang-xinyu: Good idea!

pyseir/deployment/demographic_mapper.py
zhang-xinyu: This would be necessary when the age groups of the mle model are not same as those that are used to generate the pdf (or more accurately pmf) function. In many cases, sum of pdf would be < 1 without the normalization. Another reason of doing this is when target age distribution pdf is not given and weights is set to array of ones (although such case should be quite rare...).

pyseir/deployment/demographic_mapper.py
zhang-xinyu: The dot is to calculate average of compartment size weighted by group size ratio. Potentially we could also store the adjusted compartment size by age group (elementwise multiplication), but not sure if that would be necessary since we will just calculate the index based on the aggregated measure.

environment.yaml
```diff
@@ -37,3 +37,4 @@ dependencies:
   - requests==2.23.0
   - simplejson==3.17.0
   - adtk
+  - dill
```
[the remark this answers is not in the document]
zhang-xinyu: Not aware of this, thanks!

pyseir/deployment/demographic_mapper.py
macklin: :+1: makes sense

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,719 @@
...
+            total_rate_out_of_I = self.parameters['delta']
+            for key in hospital_inflow_rates:
+                total_rate_out_of_I += hospital_inflow_rates[key]
+
+            HR = {}
+            for key in hospital_inflow_rates:
+                HR[key] = np.tile(hospital_inflow_rates[key] / total_rate_out_of_I,
+                                  (len(self.parameters['t_list']), 1)).T
+            for key in HR:
+                IHR[key][measure_unit.value] = HR[key] * fraction_of_symptomatic
```
macklin, line 379: Add an `else` statement that raises a `ValueError` (stating that `measure_unit` isn't implemented)

pyseir/deployment/demographic_mapper.py
```diff
@@ -0,0 +1,719 @@
...
+
+        mortality_probs = {}
+        for key in hospital_recovery_inflow_rates:
+            mortality_probs[key] = \
+                mortality_inflow_rates[key] / (mortality_inflow_rates[key] + hospital_recovery_inflow_rates[key])
+
+        IFR = defaultdict(dict)
+
+        for key in mortality_probs:
+            IFR[key][measure_unit.value] = mortality_probs[key] * IHR[key][measure_unit.value]
```
macklin, line 432: Could you add an `if measure_unit is CovidMeasureUnit.PER_CAPITA` statement in the appropriate place here to make things more clear?

Issue #361, opened 2020-05-01
Regression: Aggregate API endpoints contain unexpected __root__ key at top-level.


mikelehen:
See https://data.covidactnow.org/snapshot/208/us/counties.OBSERVED_INTERVENTION.timeseries.json

There's a top-level `__root__` key that should not be there.

For comparison: https://data.covidactnow.org/latest/us/counties.OBSERVED_INTERVENTION.timeseries.json

mikelehen:
Thanks @ghop02 !

PR #362, opened 2020-05-01
Fix root and a couple small fixes

ghop02:
This PR addresses issue #361

### Please Check
- [x] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [x] Are tests passing?


libs/pipelines/api_pipeline.py
```diff
@@ -218,6 +217,34 @@ def build_county_summary_from_model_output(input_dir) -> List[APIOutput]:
     return results
 
 
+def remove_root(obj: dict):
+    """Removes __root__ and replaces with __root__ value.
+
+    When pydantic models with the key __root__ are used to represent
+    json lists are serialized using `pydantic_model.dict()` the "__root__"
+    key is included. This will remove that __root__ key and replace with a list.
```
mikelehen, line 225: Thank you for the comments, but I cannot parse the first sentence here. 😬 Mind rephrasing?

libs/pipelines/api_pipeline.py
```diff
@@ -218,6 +217,34 @@ def build_county_summary_from_model_output(input_dir) -> List[APIOutput]:
     return results
 
 
+def remove_root(obj: dict):
```
mikelehen, line 220: What would you think of remove_root_wrapper ?

libs/pipelines/api_pipeline.py
```diff
@@ -218,6 +217,34 @@ def build_county_summary_from_model_output(input_dir) -> List[APIOutput]:
...
+    # Objects with __root__ should have it as the only key.
+    if len(obj) == 1 and '__root__' in obj:
+        return obj['__root__']
+
+    results = {}
+    for key, value in obj.items():
+        if isinstance(value, dict):
+            value = remove_root(value)
+
+        results[key] = value
```
mikelehen, line 243: Do we really need to do this recursively?  Maybe I don't understand how we're using __root__.  I assumed it was only an issue for root arrays.

mikelehen:
Heh, sorry. I was too slow.  Feel free to ignore.

libs/pipelines/api_pipeline.py
ghop02: heh yeah will do

libs/pipelines/api_pipeline.py
ghop02: we do because root can be on any `pydantic.BaseModel` object.

so if you had:
```
class FirstList(pydantic.BaseModel):
    __root__: List = pydantic.Field(...)


class SecondList(pydantic.BaseModel):
    __root__: List[FirstList] = pydantic.Field(...)

```

calling dict on an instance of `SecondList` would return:
```
{"__root__": [{"__root__": []}])
```
what we want is: `[[]]`

mikelehen: I see. 👍 

Thanks for the explanation.

Issue #363, opened 2020-05-01
Snapshot 211 timeseries data has extra 2020-04-28 entries for Nevada.


mikelehen:
> $ curl -s https://data.covidactnow.org/snapshot/211/us/states/NV.OBSERVED_INTERVENTION.timeseries.json | python -mjson.tool | grep 2020-04-28 | wc -l
>
> 18

This causes weirdness like this on the website:
![image](https://user-images.githubusercontent.com/206364/80780757-e4af5200-8b24-11ea-8b7d-692e61dd37f3.png)

I don't know what's going on.

PR #365, opened 2020-05-01
Fix testing timeseries data pulling in county data as well

ghop02:
This PR addresses issue #363

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/api/can_api_definition.py
+++ b/api/can_api_definition.py
@@ -188,9 +188,32 @@ class CovidActNowStateTimeseries(CovidActNowStateSummary):
 
         return rows
 
+    @pydantic.validator('timeseries')
+    def check_timeseries_one_row_per_date(cls, rows, values):
+        dates_in_row = len(set(row.date for row in rows))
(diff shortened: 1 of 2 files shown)


api/can_api_definition.py
```diff
@@ -188,10 +188,33 @@ def check_timeseries_have_cumulative_test_data(cls, rows, values):
...
+            )
+
+        return rows
+
 
 class CovidActNowCountyTimeseries(CovidActNowCountySummary):
     timeseries: List[CANPredictionTimeseriesRow] = pydantic.Field(...)
 
+    @pydantic.validator('timeseries')
+    def check_timeseries_one_row_per_date(cls, rows, values):
```
igorkofman, line 208: does this have to be duplicated?

ghop02: I think since the county and state are separate pydantic models we need to.  can probably define the logic in some similar spot, but not sure there's a single point that works super easy

PR #368, opened 2020-05-01
Make whitelist slightly more stringent and tweak fitter.

erccarls:
Make the whitelist a little more conservative from now, going from 20 case low threshold to 50 case, which is 50% of counties -> 25% of counties due to the steep recall curve.  We can try to push this back down in a few days when we move to R_t suppression models.

Also some minor fit tweaks and plotting bug fix.

image

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/inference/model_fitter.py
+++ b/pyseir/inference/model_fitter.py
@@ -67,7 +67,7 @@ class ModelFitter:
 
     DEFAULT_FIT_PARAMS = dict(
         R0=3.4, limit_R0=[2, 4.5], error_R0=.05,
-        log10_I_initial=1, limit_log10_I_initial=[.333, 3],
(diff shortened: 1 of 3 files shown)


pyseir/inference/model_fitter.py
```diff
@@ -170,7 +170,7 @@ def set_inference_parameters(self):
 
         if len(self.fips) == 5:
             # Setting
-            idx_enough_cases = np.argwhere(np.cumsum(self.observed_new_cases) >= 3)[0][0]
+            idx_enough_cases = np.argwhere(np.cumsum(self.observed_new_cases) >= 2)[0][0]
```
ghop02, line 173: Would love to pull this `2` out as a constant, just to give an opportunity to comment if possible

PR #369, opened 2020-05-01
Add ICU Normalization + Param Update.

erccarls:
Add in ICU data and rescaling. Update hosp acute LOS to reduce the occupancy ratio for ICU / total.  The results below show how we improve the absolute tuning, but the variation in ICU / general occupancy is also reduced to zero for states where this is available.

These plots are for CA only.

Before. Hosp general too low, ICUs look right in absolute terms.
--------
image
image

After. Hosp and ICUs look right in absolute and relative terms.
---------
image
image


erccarls:
Re: inference, we should be fitting the full time series and then trust those results. That is what pyseir is doing currently, but these just rescale to today's values. For better or worse, the public seems to like seeing consistency with the daily numbers. 

Upcoming PR will add some of the ICU pieces to inference (since it already has overall hosps).

PR #370, opened 2020-05-02
Expand interventions that have inferred results to lower the threshold

ghop02:
basically because all of the results are now using infernece there are less counties that are making it through.  this change counts all interventions as using the inference algorithm so it will lower the minimum acceptable counties in the output

### DIFF
--- a/libs/enums.py
+++ b/libs/enums.py
@@ -13,6 +13,15 @@ class Intervention(enum.Enum):
     OBSERVED_INTERVENTION = 2 # given the previous pattern, how do we predict going forward
 
     @classmethod
+    def inferred_interventions(cls):
+        return [
+            Intervention.NO_INTERVENTION,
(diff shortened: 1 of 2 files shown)


libs/enums.py
```diff
@@ -12,6 +12,15 @@ class Intervention(enum.Enum):
     # We are using enum 2 for consistency with the website
     OBSERVED_INTERVENTION = 2 # given the previous pattern, how do we predict going forward
 
+    @classmethod
+    def inferred_interventions(cls):
+        return [
+            Intervention.NO_INTERVENTION,
+            Intervention.STRONG_INTERVENTION,
+            Intervention.WEAK_INTERVENTION,
+            Intervention.SELECTED_INTERVENTION,
```
mikelehen, line 21: I think you're missing OBSERVED_INTERVENTION ?

PR #372, opened 2020-05-02
Adding multi datasource functionality

ghop02:
This lets us build a dataset that includes columns from different data sources.

It enables a really clean api, where you don't need to know what data source to use to get the data you need:
```
from libs.datasets.aggregate_actual_timeseries import AggregateActualTimeseries
timeseries = AggregateActualTimeseries.initialize()
timeseries.get_subset(AggregationLevel.STATE, after='2020-04-12', state='NV').data
```
image

```
from libs.datasets.aggregate_actual_timeseries import CombinedLatestValues
latest = CombinedLatestValues.initialize()
latest.get_data_for_state('NV')

{'aggregate_level': 'state',
 'country': 'USA',
 'state': 'NV',
 'fips': '32',
 'staffed_beds': 7702.0,
 'licensed_beds': 9585.0,
 'all_beds_occupancy_rate': 0.68,
 'icu_occupancy_rate': 0.79,
 'icu_beds': 844.0,
 'population': 3080156.0,
 'deaths': 243.0,
 'cases': 5053.0,
 'current_icu': 158.0,
 'current_hospitalized': 364.0,
 'positive_tests': 4998,
 'negative_tests': 36988}
```

This will make surfacing and pulling the data from various data sources much easier - as it will be a single source of truth for all of the data we're using.

Wanted to throw this up to get some feedback.

Couple things I would like to improve:
 * I don't love the names `MetadataDataset`.  It's supposed to be all the data that is _not_ timeseries - so like population, or latest case count.
 * Class names for these Combined/Aggregate classes.  Any ideas or suggestions?

Anyways, i have more thoughts, but wanted to get this out there for people to see!


libs/datasets/dataset_utils.py
```diff
@@ -279,3 +280,97 @@ def summarize(data, aggregate_level, groupby):
...
+    existing_df = existing_df.set_index(index_fields)
+
+    # Sort indices so that we have chunks of equal length in the
+    # correct order so that we can splice in values.
+    existing_df = existing_df.sort_index()
+    new_data = new_data.sort_index()
+
+    # Build series that point to rows that match in each data frame.
+    existing_df_in_new_data = existing_df.index.isin(new_data.index)
+    new_data_in_existing_df = new_data.index.isin(existing_df.index)
```
fridiculous, line 348: the section above feels like it really should be done via a join, no?

ghop02: I found the join tricky with only updating a few columns. would be down to do it with a join if it's possible.

ghop02:
I would like to add better validation around duplicates and add a few more comments, but this should be in a good spot to review

I'd like to check that the results are the same between master and the api outputs here also

api/can_api_definition.py
```diff
@@ -97,6 +110,12 @@ class CovidActNowCountySummary(CovidActNowAreaSummary):
     countyName: str = pydantic.Field(..., description="The county name")
 
 
+class CANActualsTimeseriesRow(_Actuals):
+    date: datetime.date = pydantic.Field(
```
fridiculous, line 114: can we formatting to the date?

libs/datasets/combined_datasets.py
```diff
@@ -0,0 +1,173 @@
...
+from libs.datasets.sources.nha_hospitalization import NevadaHospitalAssociationData
+from libs.datasets.sources.cds_dataset import CDSDataset
+from libs.datasets.sources.covid_tracking_source import CovidTrackingDataSource
+from libs.datasets.sources.covid_care_map import CovidCareMapBeds
+from libs.datasets.sources.fips_population import FIPSPopulation
+from libs.datasets import CommonFields
+from libs.datasets import dataset_filter
+from libs import us_state_abbrev
+
+FeatureDataSourceMap = NewType(
```
fridiculous, line 19: As a whole, this is a big step in the right direction.  As you've clearly outlined here, with this data structure, we can see the original sources for a given field.

For the purposes of visibility, it's probably best to note that we've discussed that one way to improve this would be to move from showcasing datasets dependencies to showing casing a dependency graph of transformations.

ghop02: Yeah that's a great idea -  I'll add a comment around that

libs/datasets/timeseries.py
```diff
@@ -133,8 +128,32 @@ def get_subset(
 
         return self.__class__(data)
 
+    def get_data_for_fips(self, fips) -> List[dict]:
```
fridiculous, line 131: as a rule of thumb, can we replace `data` with the type of data being returned?

```suggestion
    def get_records_with_fips(self, fips) -> List[dict]:
```

libs/datasets/timeseries.py
```diff
@@ -133,8 +128,32 @@ def get_subset(
...
+            fips: FIPS code.
+
+        Returns: List of dictionary records with NA values replaced to be None
+        """
+
+        pd_data = self.get_subset(None, fips=fips).data
+        pd_data = pd_data.where(pd.notnull(pd_data), None)
+        return pd_data.to_dict(orient="records")
+
+    def get_data_for_state(self, state) -> List[dict]:
```
fridiculous, line 144: It's just that the term `data` tends obfiscate the underlying object type.

```suggestion
    def get_records_in_state(self, state) -> List[dict]:
```

libs/functions/generate_api.py
```diff
@@ -84,73 +86,50 @@ def _generate_api_for_projections(projection_row):
...
-        cumulativePositiveTests=None,
-        cumulativeNegativeTests=None,
-        hospitalBeds={
-            "capacity": projection_row[rc.PEAK_BED_CAPACITY],
-            "currentUsage": None,
-            "typicalUsageRate": county_beds_data.get(BedsDataset.Fields.ALL_BED_TYPICAL_OCCUPANCY_RATE),
-        },
-        ICUBeds=icu_beds,
-    )
+def _generate_actuals_timeseries(actuals_timeseries_dataset, intervention):
```
fridiculous, line 124: i'm ok with this, since it's how it was done before, but i'd probably like move away from an iter-row based data transformation approach.

libs/us_state_abbrev.py
```diff
@@ -77,6 +77,7 @@
     "Colorado": "08",
     "Connecticut": "09",
     "Delaware": "10",
+    "District of Columbia": "11",
```
fridiculous, line 80: TIL

libs/datasets/latest_values_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @property
+    def county_data(self) -> pd.DataFrame:
+        """Returns a new BedsDataset containing only county data."""
+        is_county = (
+            self.data[self.Fields.AGGREGATE_LEVEL] == AggregationLevel.COUNTY.value
+        )
+        return self.data[is_county]
+
+    def get_data_for_state(self, state) -> dict:
```
fridiculous, line 175:
```suggestion
    def get_record_for_state(self, state) -> dict:
```

libs/datasets/latest_values_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        # we map NaNs to none here so that they can be generated via the API easier
+        data = self.state_data.where(pd.notnull(self.state_data), None)
+        row = data[data[self.Fields.STATE] == state]
+        if not len(row):
+            return {}
+
+        return row.iloc[0].to_dict()
+
+    def get_data_for_fips(self, fips) -> dict:
```
fridiculous, line 191:
```suggestion
    def get_record_for_fips(self, state) -> dict:
```

test/combined_dataset_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.mark.parametrize("data_source_cls", [
+    JHUDataset,
+    CDSDataset,
+    CovidTrackingDataSource,
+    NevadaHospitalAssociationData,
+
+])
+def test_unique_timeseries(data_source_cls):
```
fridiculous, line 37: my favorite part of the pr

PR #373, opened 2020-05-03
Adding api for timeseries of actuals

goldblatt:
Adding the api for the actual timeseries

### DIFF
--- a/libs/datasets/latest_values_dataset.py
+++ b/libs/datasets/latest_values_dataset.py
@@ -176,7 +176,8 @@ class LatestValuesDataset(object):
 
         Returns: Dictionary with all data for a given state.
         """
-        data = self.state_data
+        # we map NaNs to none here so that they can be generated via the API easier
+        data = self.state_data.where(pd.notnull(self.state_data), None)
         row = data[data[self.Fields.STATE] == state]
         if not len(row):
             return {}
@@ -191,7 +192,8 @@ class LatestValuesDataset(object):
 
         Returns: Dictionary with all data for a given fips code.
         """
-        row = self.data[self.data[self.Fields.FIPS] == fips]
+        # we map NaNs to none here so that they can be generated via the API easier
+        row = self.data[self.data[self.Fields.FIPS] == fips].where(pd.notnull(self.data), None)
         if not len(row):
             return {}
 
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -124,13 +124,28 @@ class TimeseriesDataset(object):
 
         return self.__class__(data)
 
-    def get_data_for_fips(self, fips) -> List[dict]: 
-        pd_data = self.get_subset(fips=fips).data
-        return pd_data.where(pd.notnull(pd_data), None).to_dict(orient="records")
+    def get_data_for_fips(self, fips) -> List[dict]:
+        """Get data for FIPS code.
 
-    def get_data_for_state(self, state) -> List[dict]: 
-        """ state abbrev here """
-        pd_data = self.get_subset(aggregation_level=AggregationLevel.STATE, state=state).data
+        Args:
+            fips: 2 letter state abbrev.
+
+        Returns: List of dictionary records with NA values replaced to be None
+        """
+
+        pd_data = self.get_subset(None, fips=fips).data
+        pd_data = pd_data.where(pd.notnull(pd_data), None)
+        return pd_data.to_dict(orient="records")
+
+    def get_data_for_state(self, state) -> List[dict]:
+        """Get data for state.
+
+        Args:
+            state: 2 letter state abbrev.
+
+        Returns: List of dictionary records with NA values replaced to be None.
+        """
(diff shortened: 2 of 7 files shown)


libs/functions/generate_api.py
```diff
@@ -94,60 +97,60 @@ def _generate_state_actuals(
...
-            "capacity": state_beds_data[BedsDataset.Fields.ICU_BEDS],
+            "capacity": latest_actuals[CommonFields.ICU_BEDS],
             "currentUsage": None,
-            "typicalUsageRate": state_beds_data[BedsDataset.Fields.ICU_TYPICAL_OCCUPANCY_RATE],
+            "typicalUsageRate": latest_actuals[CommonFields.ICU_TYPICAL_OCCUPANCY_RATE],
         },
     )
 
 
 def _generate_county_actuals(projection_row, state_intervention, county_beds_data):
```
ghop02, line 130: We probably don't need to pass in "county_beds_data" any more.

libs/functions/generate_api.py
```diff
@@ -83,75 +85,68 @@ def _generate_api_for_projections(projection_row):
...
 
+def _generate_actuals_timeseries( actuals_timeseries_dataset, intervention): 
+    actual_timeseries_api_response = []
+    for row in actuals_timeseries_dataset: 
+        actual = _generate_actuals(row, intervention.name)
+        timeseries_actual = CANActualsTimeseriesRow(**actual.dict(), date=row[CommonFields.DATE])
+        actual_timeseries_api_response.append(timeseries_actual)
+    return actual_timeseries_api_response
+
+def _generate_state_actuals(
```
ghop02, line 128: I think if you were to pass in `latest_actuals` here and do the logic for choosing which source to get the data for in the calling functions, you could just call `_generate_actuals` everywhere instead of needing another county/state split

libs/functions/generate_api.py
```diff
@@ -83,75 +85,68 @@ def _generate_api_for_projections(projection_row):
...
-        cumulativeNegativeTests=_get_or_none(
-            projection_row[rc.CUMULATIVE_NEGATIVE_TESTS]
-        ),
-        hospitalBeds={
-            "capacity": projection_row[rc.PEAK_BED_CAPACITY],
+def _generate_actuals(actual_data, intervention_str): 
+    hospital_beds = None
+    if CommonFields.MAX_BED_COUNT in actual_data and _get_or_none(actual_data[CommonFields.MAX_BED_COUNT]):
+        hospital_beds = {
+            "capacity": actual_data[CommonFields.MAX_BED_COUNT],
```
ghop02, line 92: Can you make capacity an optional field? then we can have this field without any if statements.

Also, you should be able to get `currentUsage` from `actual_data[CommonFields.CURRENT_HOSPITALIZED]` (or something like that).

libs/functions/generate_api.py
```diff
@@ -83,75 +85,68 @@ def _generate_api_for_projections(projection_row):
...
-def _generate_county_actuals(projection_row, state_intervention, county_beds_data):
-    intervention_str = state_intervention.name
-
-
-    icu_beds = None
-    if county_beds_data:
-        icu_beds = {
-            "capacity": county_beds_data[BedsDataset.Fields.ICU_BEDS],
+            "capacity": actual_data[CommonFields.ICU_BEDS],
             "currentUsage": None,
```
ghop02, line 105: Can add in current icu usage from the actuals row

libs/datasets/timeseries.py
```diff
@@ -124,6 +124,13 @@ def get_subset(
 
         return self.__class__(data)
 
+    def get_data_for_fips(self, fips): 
+        return self.get_subset(aggregation_level=AggregationLevel.COUNTY, fips=fips).data.to_dict(orient="records")
+
+    def get_data_for_state(self, state):
```
ghop02, line 130: Can you add a return type here? I think it's just `-> List[dict]:`

libs/datasets/latest_values_dataset.py
```diff
@@ -191,7 +191,7 @@ def get_data_for_fips(self, fips) -> dict:
 
         Returns: Dictionary with all data for a given fips code.
         """
-        row = self.data[self.data[self.Fields.FIPS] == fips]
+        row = self.data[self.data[self.Fields.FIPS] == fips].where(pd.notnull(self.data), None)
```
ghop02, line 194: Would be helpful to add this as either as a comment/docstring or maybe have an option that defaults to true (`replace_nan_with_none` or something).

Mostly since the resulting rows of the option are changed to object, that may not always be what we want.

libs/datasets/timeseries.py
```diff
@@ -124,6 +124,15 @@ def get_subset(
 
         return self.__class__(data)
 
+    def get_data_for_fips(self, fips): 
+        pd_data = self.get_subset(aggregation_level=AggregationLevel.COUNTY, fips=fips).data
```
ghop02, line 128: fwiw, I don't think this has to be county - if you pass a state fips code, I wouldn't expect it to return none

PR #382, opened 2020-05-06
Revert revert and gracefully handle error

ghop02:
Reverts the changes and fixes the error that was failing.

This is a quick fix for the errant borough - lets us not fail but still get alerted in sentry.

I have another task up next for handling this in a more cohesive way.

### DIFF
--- a/.github/workflows/python_unit_tests.yml
+++ b/.github/workflows/python_unit_tests.yml
@@ -29,6 +29,14 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Cache Pip
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('requirements*.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+          ${{ runner.os }}-
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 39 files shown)


TomGoBravo:
Just looking at the non-revert part of the change, looks good. Thank you for introducing me to the codebase.

```
$ git diff origin/master^..origin/revert-revert
diff --git a/pyseir/inference/infer_rt.py b/pyseir/inference/infer_rt.py
index 63c9fd6..518e4f5 100644
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -1,6 +1,7 @@
 import math
 from datetime import datetime, timedelta
 import numpy as np
+import sentry_sdk
 import logging
 import pandas as pd
 from scipy import stats as sps
@@ -484,8 +485,12 @@ class RtInferenceEngine:
 
     @classmethod
     def run_for_fips(cls, fips):
-        engine = cls(fips)
-        return engine.infer_all()
+        try:
+            engine = cls(fips)
+            return engine.infer_all()
+        except Exception as e:
+            sentry_sdk.capture_exception(e)
+            return None
 
 
 def run_state(state, states_only=False):
```

igorkofman:
It's expected that we run Bronx and it error out?

ghop02:
Yeah, for right now we do expect this to fail.  It's mostly going along with the principle of alert us but don't break everything so we can continue QAing.  I think the ticket we created should fix the source of this issue

PR #384, opened 2020-05-07
Removing dod pipeline

ghop02:
This removes all traces of the dod pipeline


mikelehen:
@asciimike says we can move forward with this.  But you have a merge conflict now.  Sorry! 😢

PR #386, opened 2020-05-07
make pyseir load from combined datasets rather than pulling in covid tracking

ghop02:
We moved the logic for where NV beds from covid tracking to the combined dataset.  use that for hospitalization numbers. 

Without this fix pyseir was not picking up the NV bed overrides

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -114,6 +114,21 @@ def build_us_latest_with_all_fields() -> LatestValuesDataset:
     )
 
 
+def get_us_latest(state=None, fips=None) -> dict:
+    """Gets latest values from from either a fips code or state."""
+    if state and fips:
+        raise ValueError("Must specify only state or fips")
+
+    if not (state or fips):
+        raise ValueError("Must specify either state or fips")
+
+    us_latest = build_us_latest_with_all_fields()
+    if state:
+        return us_latest.get_record_for_state(state)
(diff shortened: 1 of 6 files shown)


libs/datasets/combined_datasets.py
```diff
@@ -114,6 +114,21 @@ def build_us_latest_with_all_fields() -> LatestValuesDataset:
     )
 
 
+def get_us_latest(state=None, fips=None) -> dict:
```
TomGoBravo, line 117: why not have separate functions for state and fips? They are called from different places and have mostly separate code paths.

ghop02: yeah I think I agree with your thinking here.  I think I went with this because it returns the data in pretty much the same format.

Side note - we recently standardized all of the state level data to have a state fips code, we might be able to deprecate having to query this data by the state name - A lot of the modeling code handles getting state and county level data separately.

pyseir/parameters/parameter_ensemble_generator.py
```diff
@@ -49,31 +36,32 @@ def __init__(self, fips, N_samples, t_list,
         if self.agg_level is AggregationLevel.COUNTY:
             self.county_metadata = load_data.load_county_metadata().set_index('fips').loc[fips].to_dict()
```
goldblatt, line 37: lol this method to get the fips

PR #395, opened 2020-05-10
Adding cache dir to combined datasets

ghop02:
Alrighty, so i found out that one of the main reasons the modeling code is very slow now is that it's reloading the combined datasets multiple times. 

We attempt to get around this with the lru_cache decorator, but it seems that the `lru_cache` does not cache across processes (which makes sense). So, as a result, it's reloading the combined datasets a lot of times.  This is a very slow operation which is not really intended to be done more than once. 

So, I added in a decorator that caches these datasets on disk which greatly speeds up the operations.  I don't really love this approach in general - I would much rather refactor the pyseir code to load the required data frames once and figure out the best way to pass that data through - but that requires a lot more refactoring and more time than i have right now.

Locally, for just massachusetts, this cuts the execution time on my 6-core machine from 5:00 -> 2:00 (with basically the first minute just being loading the dataset the first time).
There are a handful of other improvements i have that should drop the time as well, but this is a good start.

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -1,4 +1,5 @@
 from typing import Dict, Type, List, NewType
+import logging
 import functools
 import pandas as pd
 from libs.datasets import dataset_utils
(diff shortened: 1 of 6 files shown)


TomGoBravo:
Do you think it may be possible to call the functions with lru_cache once before forking, then each subprocess will get a copy of the cache when created and we avoid creating additional state on disk?

https://repl.it/@TomGoBravo/ShorttermMealyFormulas suggests that this works.

TomGoBravo:
Cache on disk does seem like a reliable way to share information between processes. This seems like a tradeoff between a slightly more complex bit of code dealing with more state vs slightly more delicate code timing the cache population to happen before any forks. I don't see a clear winner. I suspect we will eventually put more intermediate state on disk but probably not quite like this.

PR #396, opened 2020-05-11
call cache at beginning of pyseir

ghop02:
A bit easier solution


pyseir/cli.py
```diff
@@ -38,6 +39,9 @@
 
 
 def _cache_global_datasets():
+    combined_datasets.build_us_latest_with_all_fields()
```
TomGoBravo, line 42: Please add a comment explaining why we're calling them here and discarding the returned values.

PR #397, opened 2020-05-11
Rewrite fill_fields_with_data_source to ease timeseries processing

TomGoBravo:
## Background

Part of
https://trello.com/c/oGNg8jGK/17-update-icu-hospital-data-from-daily-nevada-emails
involves changing how timeseries data that appear in multiple sources get combined. https://trello.com/c/tQj1yUjh/145-cds-data-source-missing-proper-county-level-data also increased the number of places that fields from multiple sources are combined.

https://trello.com/c/rAeUgddz/117-add-a-warning-about-dataset-merging-so-we-know-when-it-starts-happening

## This PR

* Rewrite fill_fields_with_data_source to make it easier to handle entire an timeseries at once in a future change
* add dependency on structlog, configure it with sentry and use it log the the dataset SOURCE_NAME when a duplicate field is found

## Testing

* Tests added in https://github.com/covid-projections/covid-data-model/pull/394 still pass and added a new one for empty case not covered
* `pytest test` passes
* Ran `py-spy record --output combine.svg --rate 10 --subprocesses --native pytest test/combined_dataset_test.py` at master and in this branch. Results were variable but 92-102s on my laptop. This did reveal that around 50% of the time is spent getting timezone information while panda builds indexes.

### Please Check
- [ ] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/master/api/schemas) updated? *n/a*
- [ ] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated? *n/a*
- [X] Are tests passing?


TomGoBravo:
Flame chart as an SVG: [combine-flame.zip](https://github.com/covid-projections/covid-data-model/files/4611649/combine-flame.zip)

TomGoBravo:
Hmmm, not sure what to make of https://github.com/covid-projections/covid-data-model/pull/397/checks?check_run_id=664588534 so far....

```
=================================== FAILURES ===================================
________________ [pylint] pyseir/backtest/timeseries_metrics.py ________________
E:113,28: Function 'any' has no 'mean' member (no-member)
E:116,28: Function 'any' has no 'mean' member (no-member)
```

libs/datasets/dataset_utils.py
```diff
@@ -330,48 +425,7 @@ def fill_fields_with_data_source(
...
-    # Fill in values for rows that match in both data frames.
-    existing_df.loc[existing_df_in_new_data, columns_to_fill] = new_data.loc[
-        new_data_in_existing_df, columns_to_fill
-    ]
-    # Get rows that do not exist in the existing data frame
-    missing_new_data = new_data[~new_data_in_existing_df]
-
-    data = pd.concat(
-        [existing_df.reset_index(), missing_new_data[columns_to_fill].reset_index(),]
+    assert len(columns_to_fill) == 1
```
ghop02, line 428: We have interspersed asserts in various parts of the code, but I'd love to move away from them.  Could you maybe make this check raise a `ValueError` if not true?

test/dataset_utils_test.py
```diff
@@ -41,10 +44,56 @@ def to_dict(keys: List[str], df: pd.DataFrame):
...
-        df = df.reset_index()
-    df = df.set_index(keys)
-    return df.to_dict(orient="index", into=NoNanDict)
+    try:
+        if any(df.index.names):
+            df = df.reset_index()
+        df = df.set_index(keys)
+        return df.to_dict(orient="index", into=NoNanDict)
+    except Exception:
+        print(f"Problem with {df}")
```
ghop02, line 53: What is the reason for catching and re-raising the error here?

libs/datasets/dataset_utils.py
```diff
@@ -284,6 +285,100 @@ def summarize(data, aggregate_level, groupby):
     print(index_size[index_size > 1])
 
 
+def _clear_common_values(existing_df, data_source, index_fields, column_to_fill):
+    """For labels shared between existing_df and data_source, clear column_to_fill of existing_df inplace."""
```
ghop02, line 289: I'm not sure labels is immediately obvious to me? Is a label an index row?

ghop02:
> Hmmm, not sure what to make of https://github.com/covid-projections/covid-data-model/pull/397/checks?check_run_id=664588534 so far....
> 
> ```
> =================================== FAILURES ===================================
> ________________ [pylint] pyseir/backtest/timeseries_metrics.py ________________
> E:113,28: Function 'any' has no 'mean' member (no-member)
> E:116,28: Function 'any' has no 'mean' member (no-member)
> ```
> 
> but the same failure happens in `master`

I wonder if a dependency changed. I'll have to look at that code that's throwing the error

TomGoBravo:
> Overall looks great!
> 
> I think you'll have to change how this is called from `combined_datasets` correct? Just since I think it's calling the same function for both the latest values and timeseries datasets (unless i missed something)

Yes. I'll make another PR that changes how timeseries are combined. This PR aims to preserve existing behavior. An easy way to compare snapshots would be handy here.

> Curious about the performance of the timezone indexes. Do you have a good idea of how long it took on master vs now?
CPU profile flames looked pretty similar in this branch and master. 
[combine-flames.zip](https://github.com/covid-projections/covid-data-model/files/4613865/combine-flames.zip)

libs/datasets/dataset_utils.py
TomGoBravo: Yes. Adding more docs.

libs/datasets/dataset_utils.py
TomGoBravo: What do you see as the difference between ValueError and AssertionError? They are siblings in https://docs.python.org/3/library/exceptions.html#exception-hierarchy
I see an assert as fine for checking things that are only expected with a fairly local programming error and in a long-running server it is okay to except+log+continue when one raises. ValueError is more likely to be caught due to bad user input, for a very broad interpretation of "user".

libs/datasets/dataset_utils.py
```diff
@@ -284,6 +285,100 @@ def summarize(data, aggregate_level, groupby):
...
+    if column_to_fill not in existing_df.columns:
+        existing_df[column_to_fill] = None
+
+    if date_field:
+        _clear_common_values(existing_df, new_df, index_fields, column_to_fill)
+        # From here down treat the date as part of the index label for joining rows of existing_df and new_df
+        index_fields.append(date_field)
+
+    new_df.set_index(index_fields, inplace=True)
+    if len(existing_df):
```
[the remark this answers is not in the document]
TomGoBravo: empty LGTM

libs/datasets/dataset_utils.py
```diff
@@ -284,6 +285,100 @@ def summarize(data, aggregate_level, groupby):
...
+    timeseries with a blend of values from existing_df and new_df.
+
+    See examples in dataset_utils_test.py
+
+    Args:
+        existing_df: Existing data frame
+        new_df: Data used to fill existing df columns
+        index_fields: List of columns to use as common index.
+        date_field: the time column name if the data frames represent timeseries, otherwise ''
+        column_to_fill: List of columns to add into existing_df from data_source
```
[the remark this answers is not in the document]
TomGoBravo: Oops, good catch

test/dataset_utils_test.py
TomGoBravo: added comment

libs/datasets/dataset_utils.py
ghop02: Good point re `ValueError`.  I've actually never really used `AssertionError` specifically, and probably should consider it more. 

I think my biggest problem with assert is that it doesn't actually run in non-"debug" [modes](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).  I don't necessarily think that we'll be running the python code with `-o`, but I think I shy away from asserts because of that assumption/reliance on how the python code is executed.
How do you feel about explicitly raising `AssertionError` ?

TomGoBravo: Oh wow, TIL that about `-O`. I thought https://www.python.org/dev/peps/pep-0488/#completely-dropping-optimization-levels-from-cpython was the status-quo and have even commented them out of code to save CPU :-/
Changed to raising `AssertionError`. I plan to delete this function shortly.

TomGoBravo:
I started looking at the timezones but it turned into a rathole. My best guess right now is dates are not parsed consistently between data sources (maybe in some subtle way such as naive vs UTC timezone or ns vs D resolution) which slows down joins. I created https://trello.com/c/DiCcGWlR/134-factor-out-readcsv-date-parsing as tech debt and would like to merge this change because it is blocking getting Kentucky data in.

TomGoBravo:
I'd like to get this unblocked because we're merging more data and I want more visibility into what is happening. I'm punting on performance for now, with expectation that we're about to get a bunch of low hanging fruit.

libs/datasets/combined_datasets.py
```diff
@@ -169,12 +171,16 @@ def build_combined_dataset_from_sources(
...
     # Build feature columns from feature_definition_config.
     data = pd.DataFrame({})
+    log = structlog.get_logger()
     for field, data_source_classes in feature_definition_config.items():
         for data_source_cls in data_source_classes:
             dataset = intermediate_datasets[data_source_cls]
-
             data = dataset_utils.fill_fields_with_data_source(
-                data, dataset.data, target_dataset_cls.INDEX_FIELDS, [field]
+                log.bind(dataset_name=data_source_cls.SOURCE_NAME, field=field),
```
ghop02, line 179: I'm not sure I understand why structlog is being used here? (mostly because I'm just not familiar with it)

pyseir/cli.py
```diff
@@ -59,6 +61,12 @@ def entry_point():
     """Basic entrypoint for cortex subcommands"""
     dataset_cache.set_pickle_cache_tempdir()
     sentry_sdk.init(os.getenv("SENTRY_DSN"))
+    structlog.configure(
+        processors=[
+            structlog.stdlib.add_log_level,  # required before SentryProcessor()
+            SentryProcessor(level=logging.INFO),
```
ghop02, line 67: Info feels like it might be a bit verbose?

requirements.txt
```diff
@@ -34,3 +34,5 @@ adtk==0.6.0
 pydantic==1.4
 dill
 dictdiffer==0.8.1
+structlog==20.1.0
+structlog-sentry==1.2.2
```
ghop02, line 38: Need to include more_itertools also?

pyseir/cli.py
TomGoBravo: I don't think it creates an event for every message passed to the sentry_sdk. Added comment `sentry_sdk creates events for level >= ERROR and keeps level >= INFO as breadcrumbs`

libs/datasets/combined_datasets.py
TomGoBravo: > LGTM! It would be good to run the api generation locally on recent model output to make sure that the results match before and after this change (at least on the state level)

I ran 
`python ./run.py deploy-states-api -i api-results-371/ -o api-results-371-output --summary-output api-results-371-output-states` and again with master checked
` python ./run.py deploy-states-api -i api-results-371/ -o api-results-371-output-master --summary-output api-results-371-output-states-master`
and `diff -r api-results-371-output-states api-results-371-output-states-master` found no differences.

PR #398, opened 2020-05-11
Feature/nytimes snapshot

jeremiq:
This PR addresses issue of nytimes data being read directly from the nytimes github repo instead of using our existing snapshot infra. 

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/datasets/sources/nytimes_dataset.py
+++ b/libs/datasets/sources/nytimes_dataset.py
@@ -45,7 +45,7 @@ class NYTimesDataset(data_source.DataSource):
         super().__init__(data)
 
     @classmethod
-    def load(cls) -> "NYTimesDataset":
+    def local(cls) -> "NYTimesDataset":
         data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
         return cls(data_root / cls.DATA_FOLDER / cls.COUNTIES_DATA_FILE)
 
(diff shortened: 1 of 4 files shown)


jeremiq:
This PR should not be merged until this is, https://github.com/covid-projections/covid-data-public/pull/42 -- I think?

libs/datasets/sources/nytimes_dataset.py
```diff
@@ -45,8 +45,9 @@ def __init__(self, input_path):
         super().__init__(data)
 
     @classmethod
-    def load(cls) -> "NYTimesDataset":
-        return cls(cls.DATA_URL)
+    def local(cls) -> "NYTimesDataset":
```
ghop02, line 48: 👍 thank you for this, been meaning to do it for a while

jeremiq:
This is great, the failing test should start passing when the other PR is merged.

libs/datasets/sources/nytimes_dataset.py
```diff
@@ -1,18 +1,18 @@
-from typing import List
-import logging
-import numpy
 import pandas as pd
 from libs.datasets.timeseries import TimeseriesDataset
 from libs.datasets import data_source
 from libs.datasets import dataset_utils
 from libs.datasets.common_fields import CommonIndexFields
-from libs.datasets.common_fields import CommonFields
```
ghop02, line 9: Ah, this is a bit of cleanup for code I wrote, but could you replace `TimeseriesDataset.Fields` with `CommonFields` in the `COMMON_FIELD_MAP` below? 

Just would like to keep that consistent if possible!  It is true that the `TimeseriesDataset.Fields` inherits common fields I believe, but just using `CommonFields` makes it a bit more explicity.

jeremiq: sure, will do!

PR #400, opened 2020-05-12
Adding Raw Data QA Scripts and Updating Internal Rt Plots

tashwoods:
This PR adds a separate module for comparing raw NYT data caches and updates aesthetics of internal Rt plots.

### Please Check
- [ Not Relevant here] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/master/api/schemas) updated?
- [Not Relevant ] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [So far yes ] Are tests passing?


raw_data_QA/get_data.sh
```diff
@@ -0,0 +1,8 @@
+rm -r data
+mkdir data
+curl -L https://raw.githubusercontent.com/nytimes/covid-19-data/master/us-counties.csv > data/us-counties-latest.csv
```
jeremiq, line 3: I'm not sure we need this on an on-going basis. Today I added snapshotting of nytimes data in the public repo, here: https://github.com/covid-projections/covid-data-public/tree/master/data/cases-nytimes

raw_data_QA/check_raw_data.py
```diff
@@ -0,0 +1,78 @@
+import pandas as pd
```
jeremiq, line 1: I'd re-name this file to be more specific as it's checking nytimes data only.

raw_data_QA/check_raw_data.py
```diff
@@ -0,0 +1,78 @@
...
+  parser.add_argument('-state_name', '--state_name', type = str, dest = 'state_name', default = 'state', help = 'name of state variable in input csv')
+  parser.add_argument('-new_cases_name', '--new_cases_name', type = str, dest = 'new_cases_name', default = 'cases', help = 'name of new cases column in input csv')
+  parser.add_argument('-new_deaths_name', '--new_deaths_name', type = str, dest = 'new_deaths_name', default = 'deaths', help = 'name of new deaths name in input csv')
+  parser.add_argument('-date_name', '--date_name', type = str, dest = 'date_name', default = 'date', help = 'name of datesin input csv')
+  parser.add_argument('-updated_date_name', '--updated_date_name', type = str, dest = 'updated_date_name', default = 'Date', help = 'name of date variable that is computed in this script, really just renaming')
+  args = parser.parse_args()
+
+  for state in args.state:
+    df1 = get_state(pd.read_csv("data/us-counties-may8build.csv", parse_dates=[args.date_name]), state, args)
+    df2 = get_state(pd.read_csv("data/us-counties-may11.csv", parse_dates=[args.date_name]), state, args)
```
jeremiq, line 64: is there a big benefit of having two historical datapoints to compare the current day with? 

Idea: 

Add an input parameter which is the start 'comparison date' or 'baseline date'  -- then refactor thiis script to fetch the nytimes data for that day and do your comparison with the current latest version of the data available in covid-data-public? 

This could all be done inside your python script, which would eliminate the need for the bash.

raw_data_QA/check_raw_data.py
```diff
@@ -0,0 +1,78 @@
...
+  current_month = calendar.month_name[datetime.now().month]
+  current_day = datetime.now().day
+  return current_month + ' ' + str(current_day)
+
+
+if __name__ == '__main__':
+  parser = argparse.ArgumentParser(description = 'arg parser for process.py')
+  #Output Dir------
+  parser.add_argument('-output_dir', '--output_dir', type = str, dest = 'output_dir', default = 'output', help = 'output_dir')
+  parser.add_argument('-state', '--state', type = list, dest = 'state', default = ['Idaho', 'Vermont', 'Texas'], help = 'name of state to process')
```
jeremiq, line 54: does this work with type=list? I think this gets a bit funky if you want to override this: https://stackoverflow.com/questions/15753701/how-can-i-pass-a-list-as-a-command-line-argument-with-argparse

raw_data_QA/get_data.sh
tashwoods: Yes, this was more for ad-hoc reviews.

For automated checks, it would be great to add that as an option and link to the snapshosts you take (especially if they match the API snapshots). Are your snapshots in sync with the API snapshots? If so, I could further develop this to link to your snapshots.

raw_data_QA/check_raw_data.py
tashwoods: Agreed, but I would like this script for the user to choose the specific csv's that want to compare. 

What you are describing I want to develop separately to compare API snapshots.

raw_data_QA/check_raw_data.py
tashwoods: Good point, I changed this in the last commit

raw_data_QA/check_raw_data.py
tashwoods: Good point, I removed this and renamed it in the last commit

PR #403, opened 2020-05-12
Add TestAndTrace to api output

TomGoBravo:
## Background

See https://trello.com/c/eYwZ8roa/108-add-number-of-tracers-employed-to-our-api and 
https://covidactnow.slack.com/archives/C0110QQAPJA/p1589227207223300

### Please Check
- [ ] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated? Not really, docs seem to be in flux
- [x] Are tests passing?

### DIFF
--- a/api/can_api_definition.py
+++ b/api/can_api_definition.py
@@ -80,7 +80,7 @@ class _Actuals(pydantic.BaseModel):
     hospitalBeds: Optional[_ResourceUtilization] = pydantic.Field(...)
     ICUBeds: Optional[_ResourceUtilization] = pydantic.Field(...)
     contactTracers: Optional[int] = pydantic.Field(
-        ..., description="# of Contact Tracers"
+        default=None, description="# of Contact Tracers"
     )
 
 
--- a/api/can_api_definition.py
+++ b/api/can_api_definition.py
@@ -79,6 +79,9 @@ class _Actuals(pydantic.BaseModel):
     )
     hospitalBeds: Optional[_ResourceUtilization] = pydantic.Field(...)
     ICUBeds: Optional[_ResourceUtilization] = pydantic.Field(...)
+    contactTracers: Optional[int] = pydantic.Field(
+        ..., description="# of Contact Tracers"
+    )
 
 
 class CovidActNowAreaSummary(pydantic.BaseModel):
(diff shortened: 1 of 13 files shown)


api/can_api_definition.py
```diff
@@ -79,6 +79,10 @@ class _Actuals(pydantic.BaseModel):
     )
     hospitalBeds: Optional[_ResourceUtilization] = pydantic.Field(...)
     ICUBeds: Optional[_ResourceUtilization] = pydantic.Field(...)
+    # contactTracers count is available for states, not counties.
+    contactTracers: Optional[int] = pydantic.Field(
```
ghop02, line 83: You'll also have to add this into the place where actuals get set.  see [here](https://github.com/covid-projections/covid-data-model/blob/master/libs/functions/generate_api.py#L89)

libs/datasets/combined_datasets.py
```diff
@@ -85,6 +86,7 @@
     CommonFields.ICU_TYPICAL_OCCUPANCY_RATE: [],
     CommonFields.POSITIVE_TESTS: [CDSDataset, CovidTrackingDataSource],
     CommonFields.NEGATIVE_TESTS: [CDSDataset, CovidTrackingDataSource],
+    CommonFields.CONTACT_TRACERS_COUNT: [TestAndTraceData],
```
ghop02, line 89: Can you add this to the latest values dataset above as well?

TomGoBravo: Done

api/can_api_definition.py
TomGoBravo: Done, thank You for the pointer.

libs/datasets/sources/test_and_trace.py
```diff
@@ -0,0 +1,50 @@
...
+        data[cls.Fields.COUNTRY] = "USA"
+        data[cls.Fields.AGGREGATE_LEVEL] = AggregationLevel.STATE.value
+        return data
+
+    @classmethod
+    def local(cls):
+        data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
+        input_path = data_root / cls.DATA_PATH
+        data = pd.read_csv(
+            input_path, parse_dates=[cls.Fields.DATE], dtype={cls.Fields.FIPS: str}
```
goldblatt, line 47: is this data per fips?

TomGoBravo: It is state level only. The rows have fips and state set to the correct 2 character codes.

PR #404, opened 2020-05-12
fix inferred rt calculation

tashwoods:
In this PR I updated the gaussian smoothing used to calculate Rt so that it begins counting from the first non-zero entry. Previously is started with the last non-zero entry, which would then exclude meaningful data for calculating Rt.

### Please Check
- [ we'll see!] Are tests passing?

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -320,7 +320,7 @@ class RtInferenceEngine:
 
         return dates[start_idx:], times[start_idx:], posteriors
 
-    def infer_all(self, plot=True, shift_deaths=0):
+    def infer_all(self, plot=False, shift_deaths=0):
         """
         Infer R_t from all available data sources.
 
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -190,12 +190,11 @@ class RtInferenceEngine:
                              .mean(std=self.kernel_std)\
                              .round()
 
-        zeros = smoothed.index[smoothed.eq(0)]
+        nonzeros = [idx for idx, val in enumerate(smoothed) if val!= 0]
         if len(zeros) == 0:
             idx_start = 0
         else:
-            last_zero = zeros.max()
-            idx_start = smoothed.index.get_loc(last_zero) + 1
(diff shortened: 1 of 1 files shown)


pyseir/inference/infer_rt.py
```diff
@@ -190,12 +190,11 @@ def apply_gaussian_smoothing(self, timeseries_type, plot=False):
                              .mean(std=self.kernel_std)\
                              .round()
 
-        zeros = smoothed.index[smoothed.eq(0)]
-        if len(zeros) == 0:
+        nonzeros = [idx for idx, val in enumerate(smoothed) if val!= 0]
+        if len(nonzeros) == 0:
             idx_start = 0
```
mikelehen, line 195: So if there are any zeros (the below case), we skip to the first non-zero.  If there are _only_ zeros (this case), we include them all?

Intuitively, I'd expect that we would skip them all instead...  but when you smooth zeros, you just get back zeros, so it doesn't matter?

Is it worth a comment?

> If we have a mix of zeros and nonzeros, skip to the first nonzero.

tashwoods: Right, so we just want to start with the first day that has a non-zero value. 

Case 1: 
all values are non-zero, start with the first entry

Case 2:
there are values that are zero, start with the first entry that is non-zero

Does this answer your question?

mikelehen: I think case 1 is actually "all values are zero" though (because `len(nonzeros) == 0` means there are only zeros). 😁

tashwoods: So, if all entries are zero - the data isn't providing any information, and it should not affect future predictions. I don't anticipate this happening, but we could code up a different way of dealing with that - but I don't think it is too worthwhile

mikelehen: Yeah, I agree 100%.  I just wanted to confirm my understanding.  No need to spend more time on it.  Thanks!

mikelehen:
Thanks @tashwoods and @jeremiq!  I went ahead and merged, and I'll kick off a new snapshot build that we can plug into the website.  Hopefully we see fewer r(t) charts moving around going forward. 😁

PR #405, opened 2020-05-13
first pass at QAing the api result for all states

goldblatt:

### Please Check
- [ ] Are tests passing?


qa/compare_snapshots.py
```diff
@@ -0,0 +1,38 @@
+import click
+import requests
+
+from qa.comparitor import Comparitor
+
+@click.command("compare-snapshots")
```
ghop02, line 6: Can this live in `cli/`? maybe we need more documentation around it - and pyseir is currently the exception - but all of the other cli interfaces live in `cli/`

qa/comparitor.py
```diff
@@ -0,0 +1,180 @@
+import csv
```
ghop02, line 1: Can we move this `qa` package to `libs/qa`?  I think it's confusing to have a lot of top level packages (i already don't love that pyseir/ + api/ is outside of libs) and not clear why some code is top level vs others not.

qa/comparitor.py
```diff
@@ -0,0 +1,180 @@
...
+        return self.results
+            
+    @property
+    def results(self): 
+        return self._results.values()
+
+    def write_to_csv(self, file_name):
+        with open(file_name, 'w+') as csv_file: 
+            write = csv.writer(csv_file) 
+            write.writerows(self.results)
```
ghop02, line 180: We should write the header as well, I'd recommend using the `csv.DictWriter`: https://docs.python.org/3/library/csv.html#csv.DictWriter

qa/metrics.py
```diff
@@ -0,0 +1,84 @@
+from abc import ABC, abstractmethod
+
+class QAMetric(ABC):
```
ghop02, line 3: This could also be written as a dataclass which might be a bit more clear:

then metrics would be instantiated like:
```
HospitalBedsRequired = QAMetric(
    name="HospitalBedsRequired",
    projection_path=["timeseries", "hospitalBedsRequired"],
    actual_path=None,
    threshold=1
)
```

qa/comparitor.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __init__(self, snapshot1, snapshot2, state, intervention, fips = None):
+        self.snapshot1 = snapshot1
+        self.snapshot2 = snapshot2
+        self.state = state
+        self.fips = fips
+        self.intervention = intervention
+        # TODO (sgoldblatt) swap out here if it's counties
+        self.api1, self.api2 = self._get_state_data(snapshot1, snapshot2, state, intervention)
+        self._generate_date_dictionary()
```
ghop02, line 75: Thoughts on separating out the loading of the state data from the comparitor? 

Would be great to pass in the loaded data dictionary to separate concerns a bit more. Also makes it easier to test this class without having to hit the actual url.

qa/compare_snapshots.py
```diff
@@ -0,0 +1,38 @@
...
+@click.option(
+    "--snapshot2",
+    '-s2',
+    type=int,
+    help="The number of the snapshot to compare with",
+    required=True,
+)
+@click.option(
+    "--intervention",
+    type=str,
```
ghop02, line 23: I think you can make this `type=Intervention` (though it might be some variation on it), and you can add a default:
`default=Intervention.SELECTED_INTERVENTION`

qa/comparitor.py
```diff
@@ -0,0 +1,180 @@
...
+        self.snapshot1 = snapshot1
+        self.snapshot2 = snapshot2
+        self.value1 = value1
+        self.value2 = value2
+        self.threshold = threshold
+        self.difference = difference
+
+
+class Comparitor(object): 
+    snapshot1 = None
```
ghop02, line 40: small note, but this defines them as class variables, not instance variables - could make this a dataclass too!

qa/run.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
+import click
+import requests
+from dictdiffer import diff
+
+@click.group()
```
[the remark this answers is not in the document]
goldblatt: ya i guess i was hoping to keep them as seperate as possible but eh?

qa/run.py
```diff
@@ -0,0 +1,60 @@
...
+)
+def compare_snapshot(snapshot1, snapshot2, intervention=None):
+    if not intervention:
+        # wow ugh i hate that this lives in a string....
+        intervention = "SELECTED_INTERVENTION"
+    states_json_1, states_json_2 = _get_states_county_data(snapshot1, snapshot2, intervention)
+    # complciations 1) the dates are off from the projects so need to match ond atse
+    # each field maybe needs it's own thing it's own threshold it cares about
+    # each this output is somewhat useless
+    print(list(diff(states_json_1, states_json_2, tolerance=2)))
```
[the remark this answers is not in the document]
goldblatt: yeah was just messing around

qa/comparitor.py
```diff
@@ -0,0 +1,180 @@
+import csv
+import requests
+from datetime import datetime, timedelta
+
+from dictdiffer import diff
+
+from qa.metrics import METRICS
+
+class MetricDiff(object):
```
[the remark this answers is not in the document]
goldblatt: woot

qa/comparitor.py
goldblatt: yesss was just messing around

qa/metrics.py
goldblatt: i think that the ABC makes it so that you could overwrite the diff/ compare function which is kind of nice

libs/qa/comparitor.py
```diff
@@ -0,0 +1,216 @@
...
+        for api, snapshot in [(self.api1, self.snapshot1), (self.api2, self.snapshot2)]:
+            self._generate_last_updated_helper(api, snapshot)
+
+    def _getMetricValue(self, metric: QAMetric, date: str, snapshot: int, isProjected: bool):
+        metric_path = metric.projection_path if isProjected else metric.actual_path
+        if not metric_path:
+            return None
+        data_at_date = self._date_to_data.get(date)
+        if not data_at_date:
+            capture_event(
```
ghop02, line 97: Can you add a logging warning here as well?

cli/compare_snapshots.py
```diff
@@ -0,0 +1,128 @@
...
+        raise Exception("Need to specify either snapshot or input dir not both")
+    if not (compare_snapshot):
+        compare_snapshot = _get_compare_snapshot()
+    states = [state] if state else STATES_50.values()
+    results = []
+    report = []
+
+    for state_abbrev in states:
+        if not intervention:
+            intervention = Intervention.OBSERVED_INTERVENTION
```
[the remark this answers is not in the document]
goldblatt: cool yes switched to use the name here so it's a bit more human readable

libs/qa/comparitor.py
goldblatt: yeah also this didn't seem to do the thing so messing around a bit with it now

goldblatt: okay fixed

libs/us_state_abbrev.py
```diff
@@ -68,6 +68,60 @@
     "Wyoming": "WY",
 }
 
+STATES_50 = {
```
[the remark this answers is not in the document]
goldblatt: it's the states we have data for (like without the missing territories)

libs/qa/metrics.py
```diff
@@ -0,0 +1,202 @@
+from abc import ABC, abstractmethod
+from datetime import datetime
+
+
+class QAMetric(ABC):
+    def __init__(self):
+        # ...
+        pass
+
+    @property
```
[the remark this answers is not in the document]
goldblatt: mm this is how you get the classes to implement these as class methods, otherwise they could just not have them set

libs/qa/comparitor.py
```diff
@@ -0,0 +1,216 @@
...
+        projected_value1 = self.get_metric_projected_value(metric, date, self.snapshot1)
+        projected_value2 = self.get_metric_projected_value(metric, date, self.snapshot2)
+
+        actual_value1 = self.get_metric_actual_value(metric, date, self.snapshot1)
+        actual_value2 = self.get_metric_actual_value(metric, date, self.snapshot2)
+
+        if projected_value1 and projected_value2:
+            if metric.isAboveThreshold(projected_value1, projected_value2):
+                self._results.append(
+                    MetricDiff(
```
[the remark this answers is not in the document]
goldblatt: ohh yes okay that's nice

PR #410, opened 2020-05-14
Update covid_tracking_source to check and scrub non-sense ICU values

BrettBoval:
Removes the 'current icu' data in the covid tracking dataset anytime it is greater than the 'current hospitalized' amount. Motivated by covid-tracking providing a new value for Delaware current ICU that is not consistent with the capacity and hospitalized data.

Also added an assert statement to check for this condition going forward once the Delaware problem is fixed upstream.

This PR compensates for https://github.com/COVID19Tracking/issues/issues/416.

### Please Check
- [N/A] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/master/api/schemas) updated?
- [N/A] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [Yes] Are tests passing?

### DIFF
--- a/libs/datasets/sources/covid_tracking_source.py
+++ b/libs/datasets/sources/covid_tracking_source.py
@@ -1,5 +1,6 @@
 import logging
 import pandas as pd
+import numpy as np
 from libs.datasets import data_source
 from libs.datasets import dataset_utils
 from libs.datasets.dataset_utils import AggregationLevel
@@ -45,7 +46,6 @@ class CovidTrackingDataSource(data_source.DataSource):
         IN_ICU_CURRENTLY = "inIcuCurrently"
         IN_ICU_CUMULATIVE = "inIcuCumulative"
 
-        IN_ICU_CURRENTLY = "inIcuCurrently"
         TOTAL_IN_ICU = "inIcuCumulative"
 
         ON_VENTILATOR_CURRENTLY = "onVentilatorCurrently"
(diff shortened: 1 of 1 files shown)


libs/datasets/sources/covid_tracking_source.py
```diff
@@ -130,13 +130,25 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
         # TODO(chris): Handle this in a more sane way.
         data = data.loc[data.state != "PR", :]
 
+        # Removing bad data from Delaware.
+        # Once that is resolved we can remove this while keeping the assert below.
+        icu_mask = data[cls.Fields.IN_ICU_CURRENTLY] > data[cls.Fields.CURRENT_HOSPITALIZED]
```
ghop02, line 135: Can you add a sentry error here that reports how many states/rows this affects?
could look something like this: https://github.com/covid-projections/covid-data-model/blob/master/libs/build_processed_dataset.py#L91-L95. 

Would be good to know which states this is happening to

BrettBoval: How detailed do we want to know? The example given is a pretty simple string do we want to dump all lines of the data frame where we drop a specific value?

BrettBoval: Or would a simple "X number of lines were changed due to icu correction"

ghop02: Good question! Maybe something like `X number of rows in N states...`?

ghop02: also, I'm fine to punt on this for now, getting this fix out definitely the priority!

BrettBoval:
Check out the sentry and logging now.

BrettBoval:
One this merges I'll build the snapshots and compare the Delaware data.

PR #411, opened 2020-05-14
Feature/additional rt thresholding

jeremiq:
This PR addresses imbalance of death and case rt curves on composit result.

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -200,13 +200,12 @@ class RtInferenceEngine:
                              .round()
 
         nonzeros = [idx for idx, val in enumerate(smoothed) if val!= 0]
-        if len(nonzeros) == 0:
-            idx_start = 0
-        else:
-            idx_start = nonzeros[0]
         if max(smoothed) < smoothed_max_threshold:
             # skip the entire array.
             idx_start = len(smoothed)
+        else:
+            idx_start = nonzeros[0]
+ 
         smoothed = smoothed.iloc[idx_start:]
         original = timeseries.loc[smoothed.index]
 
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -167,6 +167,13 @@ class RtInferenceEngine:
             Which type of time-series to use.
         plot: bool
             If True, plot smoothed and original data.
+        smoothed_max_threshold: int
+            This parameter allows you to filter out entire series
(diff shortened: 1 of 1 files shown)


pyseir/inference/infer_rt.py
```diff
@@ -195,6 +195,9 @@ def apply_gaussian_smoothing(self, timeseries_type, plot=False):
             idx_start = 0
         else:
             idx_start = nonzeros[0]
+        if max(smoothed) < threshold:
```
ghop02, line 198: Can you give a bit more context around what the threshold is? 

Just to make sure I'm reading it right, this is saying if the max value in the entire smoothed array is less than the threshold, we skip the entire array.

I'm wondering if threshold could be renamed to make it clear what that threshold is.

pyseir/inference/infer_rt.py
```diff
@@ -155,7 +155,7 @@ def get_timeseries(self, timeseries_type):
         elif timeseries_type in (TimeseriesType.NEW_HOSPITALIZATIONS, TimeseriesType.CURRENT_HOSPITALIZATIONS):
             return self.hospital_dates, self.hospital_times, self.hospitalizations
 
-    def apply_gaussian_smoothing(self, timeseries_type, plot=False):
+    def apply_gaussian_smoothing(self, timeseries_type, plot=False, threshold=5):
```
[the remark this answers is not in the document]
jeremiq: done. good call!

pyseir/inference/infer_rt.py
jeremiq: I updated the name and doc string to try to make it a bit clearer, but this is a very "pragmatic" addition. 

We're seeing a lot of noise from the inclusion of death data in our R(t) calculation whenever there are very few deaths (as in Alaska and Idaho, most recently). This parameter just gives us an easy binary way of filtering out (our use of) deaths for such cases.

pyseir/inference/infer_rt.py
```diff
@@ -195,6 +204,9 @@ def apply_gaussian_smoothing(self, timeseries_type, plot=False):
             idx_start = 0
```
tashwoods, line 204: So, I think this line is now redundant, as an input array that is all zeros would not pass the later max threshold requirement. I think it would be good to remove this, to make the code more clear. What do you think?

tashwoods: For v2, we could override values below the threshold to NaN, or back in infer_rt.py after bayesian updating and time lag.

PR #413, opened 2020-05-14
Woodsn qa raw

tashwoods:
This is just to work on structure with Jeremiah

### Please Check
- [ ] Are tests passing?


raw_data_QA/check_nyt_raw_data.py
```diff
@@ -0,0 +1,244 @@
...
+def get_rmse(df1, df2, var):
+  rmse = np.sqrt(mean_squared_error(df1[[var]], df2[[var]]))
+  return round(rmse,2)
+
+def get_equal_len_df(df1, df2):
+  df2_length = len(df2.index)
+  truncated_df1 = df1.head(df2_length)
+  return truncated_df1
+
+def get_compare_metrics(df1, df2, var):
```
jeremiq, line 35: This seems like a really valuable function, good idea!

raw_data_QA/get_data.sh
```diff
@@ -0,0 +1,36 @@
+#Remove old data directory if it exists and make new one
+OUTPUT_DIR="data/"
```
jeremiq, line 2: I'm still not sure how useful this data fetching script is for our ongoing data quality monitoring. 

2 ideas: 

Move your git_sha --> csv fetcher function into your python script, so you can avoid storing the intermediate csvs. 
Maybe make the sha a parameter that is passed in on the commandline and then use pd.read_csv to fetch the file directly from the public site. 

You can get the exact version of the file at a given hash at this URL: https://github.com/covid-projections/covid-data-public/raw/<YOUR COMMIT HASH HERE>/data/cases-nytimes/us-counties.csv

raw_data_QA/get_hash.py
```diff
@@ -0,0 +1,9 @@
+import json, sys
+
+if __name__ == '__main__':
```
jeremiq, line 3: I think this could be done in a simpler way using a couple of calls to the requests library -- and then incorporated in your check_nyt_raw_data.py script. 


this is not very clean -- can be split up over more lines, but you can get the snapshot version like this, for example:
```python
import requests
prod_snapshot_version = requests.get("https://raw.githubusercontent.com/covid-projections/covidprojections/master/src/assets/data/data_url.json").json()['data_url'].split("/")[-2]

master_hash = requests.get(f"https://data.covidactnow.org/snapshot/{prod_snapshot}/version.json").json()['covid-data-public']['hash']

```

raw_data_QA/get_data.sh
tashwoods: Yeah, so the way it is written now it will grab the git commit of covid-data-public from the current website and compare it to the current NYT dataset. I didn't really want to include this in check_nyt_datasets.py, so it is easier for the user to configure. If you have strong feelings I could change this, but would prefer to leave it as is.

raw_data_QA/get_hash.py
tashwoods: Yeah good point, but this gets back to my earlier point about putting all the user settings about which CSVs in the top level get_data.sh script. I can change this but prefer to keep it as is.

PR #419, opened 2020-05-15
Removing execute summaries and associated code

ghop02:
This removes the summaries:
 - latest case values
 - list of counties with data.


ghop02:
oh good call!

PR #420, opened 2020-05-15
NYT Raw Data QA

tashwoods:
This PR includes QA for NYT Data, because their data keeps changing 👍 
### Please Check
- [ they better!] Are tests passing?


raw_data_QA/get_data.sh
```diff
@@ -0,0 +1,36 @@
...
+PROD_VERSION_FILE="version.json"
+PROD_FILE=$OUTPUT_DIR"prod.version"
+HASH_FILE=$OUTPUT_DIR"hashinfo.txt"
+curl $WEBSITE_PROD > $PROD_FILE
+json_prod=$(grep -o 'http[^"]*' $PROD_FILE)$PROD_VERSION_FILE
+curl $json_prod > $HASH_FILE
+COVID_DATA_PUBLIC_MASTER_PROD=$(python get_hash.py $HASH_FILE)
+
+#The current prod site does not have the counties csv counties cached so we will use old ones here
+COVID_DATA_PUBLIC_MASTER_OLD="6f24dcb9a6b2203572c7509e506182be62117a38"
```
jeremiq, line 19: does your use of these scripts involve updating these commit hashes? 

Add something in your readme about how to track down good commit hashes for comparison?

raw_data_QA/check_nyt_raw_data.py
```diff
@@ -0,0 +1,508 @@
...
+
+        z_scores.append(z_score)
+        diff.append(p_diff)
+        number_of_days_percent_threshold = sum(i > args.percent_threshold for i in diff)
+        average_Z = round(average(z_scores), 2)
+        latest_Z = round(average(z_scores[: -args.n_days_z_score_mean]), 2)
+    return diff, z_scores, average_Z, latest_Z, number_of_days_percent_threshold
+
+
+def average(list):
```
jeremiq, line 77: maybe you answered this in the other PR, but any reason to define this function rather than using something in `np`?

TomGoBravo:
@tashwoods I'd like to get to a world where the entire covid-data-model has zero dependencies on data outside the covid-data-public repo. Perhaps it is a bit fanciful but what do you think of putting QA checks on NYT data in covid-data-public? If you think covid-data-public has deviates from the nytimes/covid-19-data repo by even single number lets fix get to the bottom of why our copy is broken. Perhaps a release process for covid-data-public makes sense where we run checks on fetched data then bless a particular covid-data-public sha for use by the run done in covid-data-model.

I tried to list out the tradeoffs...

QA checks in covid-data-model:
* don't need to change how NYT data is fetched
* lets QA display reference particular API snapshot outputs
* makes it easier of QA analysis to reuse code/tools used for other analysis in the model repo

QA checks in covid-data-public:
* changing how NYT data is fetched is already queued https://trello.com/c/CUdt78CV/107-add-nyt-data-to-covid-data-public
* helps to separate data changes (in data-public) from code changes (in data-model) which
  * makes running the model more deterministic
  * enables more caching of intermediate values for faster builds
  * provides a clean separation of concerns where someone can contribute to the data-public repo with little knowledge and no dependency on data-model

raw_data_QA/get_data.sh
tashwoods: Ok, I reformatted all of these (no more shell scripts 😀). Lmk if you still have thoughts. I agree for v2 it would be good to pull this into libs.datasets

raw_data_QA/check_nyt_raw_data.py
tashwoods: Fixed

tashwoods:
@TomGoBravo 
That is an interesting idea, to only 'release' data for covid-data-model to use once it has passed checks within covid-data-public. This seems to make sense to me, as long as we don't end up stuck on an old version of the dataset, because all the new datasets differ 'too much.' I think we will have discuss this more with @igorkofman.

For now, this works and at least provides some easy debugging. But will still have to be added to run.sh so we can see the differences between api snapshots.

If people are cool with this merge, we could add 
python check_nyt_raw_data.py
to run.sh

raw_data_QA/check_nyt_raw_data.py
```diff
@@ -309,6 +311,46 @@ def check_new_data(df1, df2, args, var):
...
+def get_df_from_url_hash(thishash, basepath, filepath, args, name):
+    this_file = requests.get(f"{BASE_PATH}/{thishash}/{filepath}").content
+    this_df = pd.read_csv(io.StringIO(this_file.decode("utf-8")), parse_dates=[args.date_name])
+    this_df.to_csv(f"{args.output_dir}/{args.output_data_dir}/{name}.csv")
+    return this_df
+
+
+def get_df_from_url(url, args, name):
+    this_file = requests.get(url).content
+    this_df = pd.read_csv(io.StringIO(this_file.decode("utf-8")), parse_dates=[args.date_name])
```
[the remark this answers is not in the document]
tashwoods: that failed. still not sure why

TomGoBravo:
Replaced by https://github.com/covid-projections/covid-data-model/pull/436

PR #424, opened 2020-05-15
ICU Estimation from Cumulative ICU Data

BrettBoval:
Improve the estimation of current icu covid patients in the case where we are only provided with cumulative icu patients.

Sometimes covidtracking provides cumulative, but not current, data for hospitalization and icu. We then estimate the number of patients currently in the ICU. There are multiple things with this implementation that need to be improved, however they can be done in stages. The first stage is to standardize the calculation of ICU. Right now the estimate uses a variable drawn from a wide distribution (roughly 8.6 +/- 2.6 days). This introduces a large source of noise in the snapshot to snapshot calculations. In other places this code is usually called multiple times and then averaged, but for here it is just called once. Instead of bumping up the call/average, I just calculated the value from the means. The consistency day-to-day is key, especially since this parameter isn't used in any monte-carlo modeling.

I left the hospitalization branch still drawing from the distribution, but averaging over 250 samples. Timeit shows an increase from 8.4ms to 27.2ms. If this becomes significant, we can refactor the params code and surface this constant.

Finally, our new linter just came online, so there will be misc line cleanups for a bit.

### Please Check
- [YES] Are tests passing?

### DIFF
--- a/pyseir/backtest/backtester.py
+++ b/pyseir/backtest/backtester.py
@@ -30,7 +30,7 @@ class Backtester:
     increases until it reaches max_observation_days_blinded.
     To remove unwanted feature (smooth) of the time series of the
     observation data, one can set window of ts_rolling_args greater than one.
-    To accelorate backtesting, one can set num_days_per_step greater than 1.
+    To accelerate backtesting, one can set num_days_per_step greater than 1.
     However it is recommended that num_days_per_step <= ts_rolling_args[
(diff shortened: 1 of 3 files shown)


BrettBoval:
I checked this change to the model with Max earlier today and he is on board.

BrettBoval:
Looking into why the build test failed.

PR #426, opened 2020-05-17
Some timeseries get_subset cleanup

TomGoBravo:
* replace `DatasetFilter.__init__` with pydantic dataclass and explicitly list the filters used
* add timeseries_test.py
* rm get_subset county= param that seems unused

Done while working towards https://trello.com/c/HLf7QmJD/118-fix-pyseir-settingwithcopywarning

### DIFF
--- a/libs/datasets/dataset_filter.py
+++ b/libs/datasets/dataset_filter.py
@@ -16,4 +16,5 @@ class DatasetFilter(object):
 
     def apply(self, dataset: Union[TimeseriesDataset, LatestValuesDataset]):
         """Applies filter to dataset by returning subset that matches filter arguments."""
-        return dataset.get_subset(aggregation_level=None, country=self.country, states=self.states)
+        aggregation_level = None
+        return dataset.get_subset(aggregation_level, country=self.country, states=self.states)
--- a/libs/datasets/dataset_filter.py
+++ b/libs/datasets/dataset_filter.py
@@ -1,22 +1,19 @@
-from typing import Union
+from typing import Union, List
+
+from pydantic.dataclasses import dataclass
+
 from libs.datasets.dataset_utils import AggregationLevel
 from libs.datasets.timeseries import TimeseriesDataset
 from libs.datasets.latest_values_dataset import LatestValuesDataset
 
 
+@dataclass
 class DatasetFilter(object):
     """Defines a filter to apply to a dataset."""
 
-    def __init__(self, aggregation_level: AggregationLevel = None, **filters):
-        """
-        Args:
-            aggregation_level: AggregationLevel of dataset
-            filters: Filters to apply to dataset.  Must match arguments of the respective
-                `get_subset` function.
-        """
-        self.aggregation_level = aggregation_level
-        self.filters = filters
+    country: str
+    states: List[str]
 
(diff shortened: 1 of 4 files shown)


libs/datasets/dataset_filter.py
```diff
@@ -1,22 +1,19 @@
...
-        """
-        self.aggregation_level = aggregation_level
-        self.filters = filters
+    country: str
+    states: List[str]
 
     def apply(self, dataset: Union[TimeseriesDataset, LatestValuesDataset]):
         """Applies filter to dataset by returning subset that matches filter arguments."""
-        return dataset.get_subset(self.aggregation_level, **self.filters)
+        return dataset.get_subset(aggregation_level=None, country=self.country, states=self.states)
```
ghop02, line 19: small nit, but since aggregation level isn't a keyword argument, i generally prefer not call it by the var name, maybe like:
```
aggregation_level = None
dataset.get_subset(aggregation_level)
```

libs/datasets/timeseries.py
```diff
@@ -128,7 +125,7 @@ def get_records_for_fips(self, fips) -> List[dict]:
 
         Args:
             fips: FIPS code.
-
+t
```
ghop02, line 128: t :)

libs/datasets/timeseries.py
```diff
@@ -147,16 +144,12 @@ def get_records_for_state(self, state) -> List[dict]:
         pd_data = self.get_subset(AggregationLevel.STATE, state=state).data
         return pd_data.where(pd.notnull(pd_data), None).to_dict(orient="records")
 
-    def get_data(
-        self, country=None, state=None, county=None, fips=None, states=None
-    ) -> pd.DataFrame:
+    def get_data(self, country=None, state=None, fips=None, states=None) -> pd.DataFrame:
```
ghop02, line 147: can you make this change on the latest values dataset as well? (not sure if it has county or not)

libs/datasets/dataset_filter.py
TomGoBravo: Done.
What do you think of moving to Python 3.8? `PEP 570 -- Python Positional-Only Parameters` was added for people like you :-P Personally I rather fancy access to the walrus operator but I haven't suggested it because I don't see any real value to switching.

libs/datasets/timeseries.py
TomGoBravo: Oops, thanks for noticing.

libs/datasets/timeseries.py
TomGoBravo: Done

libs/datasets/dataset_filter.py
ghop02: haha big pep 570 fan (more the functionality, jury's still out on if I like the syntax)!

But I have no strong opinion - I worry that because it's a bit newer, the community in general isn't as setup for 3.8 just yet, but that's not really founded on anything concrete.  
Also, if we were to make `covid-data-model` more of a traditional python pip installable package, I worry that a lot of the environments might not be working with 3.8 just yet.

TomGoBravo: :100: share concerns about ease of setting up the repo when it depends on a fairly new Python release.

PR #427, opened 2020-05-17
Fix ICU Utilization Metric When Incorporating New 'Cumulative ICU' Datastream

BrettBoval:
This is a solution for https://trello.com/c/CrcF4MsE/112-model-utah-icu-utilization-unreasonably-low-in-latest-snapshot. It adds features to the estimate current hospital load from daily cumulative datastreams. It is now more robust to the settling time errors that Michael caught with the Utah data.

It unblocks the frontend from having to mask the Utah ICU data, and will stop this from happening again as places start posting ICU specific data.

There is some more work I would like to do... refactoring ParameterEnsembleGenerator to allow quick access to the underlying attributes of a distribution (to make a deterministic average without significant/slow sampling) and consolidating load hospitalization data function (load_hospitalization_data_by_state and get_current_hospitalized share some code).

However, this is viable and I think we should merge it.

Snapshot 301 is built from this -> https://covidactnow.org/compare?left=299&metric=2&right=301&sort=1 shows there isn't much significant changes in the non-Utah based ICU metrics.

Utah is masked in that tool, but inspecting the json shows the expected outputs.
The left is 299 and the right is 301. Note that ICUBedsInUse is 34 in the new, and 21 in the old. This is consistent with the deviations fixed in the model.

image

I'm happy to explain the logic in this model. I didn't change any underlying assumptions (at the moment at least). Just initialized it with a deterministic parameter for icu length of stay, and changed the initial parameters so it could converge fast enough to be useful.

Once this is merged we can trigger a build, update covid-projections, and bump the snapshot number.

### Please Check
- [N/A] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/master/api/schemas) updated?
- [N/A] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [YES] Are tests passing?


pyseir/load_data.py
```diff
@@ -508,44 +503,185 @@ def load_hospitalization_data_by_state(
...
+    MIN_DATAPOINTS_TO_CONVERT = 3  # We wait until the third datapoint to have 2 deltas to forecast
+
+    abbr = us.states.lookup(state).abbr
+    df = (
+        combined_datasets.build_us_timeseries_with_all_fields()
+        .get_subset(AggregationLevel.STATE, country="USA", state=abbr)
+        .get_data(country="USA", state=abbr)
+    )
+
+    categories = ["icu", "hospitalized"]
```
jeremiq, line 544: pull this into a high level constant maybe, `HOSPITALIZATION_CATEGORIES`

pyseir/load_data.py
```diff
@@ -508,44 +503,185 @@ def load_hospitalization_data_by_state(
...
-                    * params["hospitalization_length_of_stay_icu_and_ventilator"]
-                )
-
-            # Now compute a cumulative sum, but at each day, subtract the discharges from the previous count.
-            new_hospitalizations = np.append([0], np.diff(cumulative))
-            current = [0]
-            for i, new_hosps in enumerate(new_hospitalizations[1:]):
-                current.append(current[i] + new_hosps - current[i] / average_length_of_stay)
-            return times_new, current, HospitalizationDataType.CURRENT_HOSPITALIZATIONS
+    if len(df) == 0:
```
jeremiq, line 548: not sure if you run into this, but I've seen a lot of 

`if df is None or if df.empty` in the code. This check will raise an exception if it gets a `None`

pyseir/load_data.py
```diff
@@ -508,44 +503,185 @@ def load_hospitalization_data_by_state(
...
+            return times_new_latest, current_latest  # Return current estimate from cumulative
         else:
-            return times_new, cumulative, HospitalizationDataType.CUMULATIVE_HOSPITALIZATIONS
+            return None, None  # No current, not enough cumulative
     else:
-        return None, None, None
+        return None, None  # No current nor cumulative
+
+
+def estimate_current_from_cumulative(cumulative, category):
```
jeremiq, line 580: This is excellent. I'm not sure any of this code should live in the `load_data` module though -- maybe we should split out a separate module for ICU estimation, even. 

That can be done in a future PR though.

BrettBoval: Agreed that it should be pulled out of load_data and that it should be a separate PR.

pyseir/load_data.py
BrettBoval: Yeah. I think this will be part of a bigger refactor. Especially if we pull out the code path for ICU utilization (that goes into the big visualization) vs. hospitalization which goes into the extrapolated days until hospital is overwhelmed.

PR #429, opened 2020-05-18
Load state data (currently only ky.csv) in combined_datasets

TomGoBravo:
https://trello.com/c/m8a0aV4T/126-get-kentucky-icu-hospitalization-testing-data-on-public-site

I don't see an obvious reliable way to check the effect this has on the output. :-(

### Please Check
- [ ] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/master/api/schemas) updated?
- [ ] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [ ] Are tests passing?

PR #432, opened 2020-05-19
Revert "Load state data (currently only ky.csv) in combined_datasets"

BrettBoval:
Reverts covid-projections/covid-data-model#429

The provided Kentucky data significantly conflicts with covidtracker and local KY news reports about ICU numbers. Alan from Partnerships is in conversation with the KY Contact, but until then we want to use the covidtracking data. Since KY is currently the only state exercising this codepath, Tom and I think it's best to just revert and leave as a branch for now. If another state comes on board this way, then we can add a KY-specific exclusion and merge these changes.


TomGoBravo:
I think it is pretty reasonable to remove code as soon as it is no longer used, then bring it back when needed again.

libs/datasets/data_source.py
```diff
@@ -13,11 +13,8 @@ class DataSource(object):
     class Fields(object):
         pass
 
-    # A map from column names in the source data to common field names. The keys of this map
```
TomGoBravo, line 16: Please keep the comments in this file.

TomGoBravo:
[Get Kentucky ICU/Hospitalization + Testing data on public site](https://trello.com/c/m8a0aV4T/126-get-kentucky-icu-hospitalization-testing-data-on-public-site)

PR #437, opened 2020-05-21
Feature/interpolated rt

jeremiq:

### Please Check

- [YES ] Are tests passing?

Results from this change can be seen here: 
snapshot 334: 

https://www.covidactnow.org/internal/compare?left=332&metric=0&right=334&sort=2

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -568,7 +568,7 @@ class RtInferenceEngine:
             output_path = get_run_artifact_path(self.fips, RunArtifact.RT_INFERENCE_REPORT)
             plt.savefig(output_path, bbox_inches="tight")
             # plt.close()
-        if df_all.empty:
+        if df_all is None or df_all.empty:
             log.warning("Inference not possible for fips: %s", self.fips)
         return df_all
 
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -366,7 +366,6 @@ class RtInferenceEngine:
                 # this is necessary since otherwise NaN values
                 # will be inferred for all future days, after seeing
                 # a single (smoothed) zero value.
-
                 posteriors[current_day] = prior0
             else:
                 posteriors[current_day] = numerator / denominator
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -362,10 +362,18 @@ class RtInferenceEngine:
 
             # Execute full Bayes' Rule
             if denominator == 0:
-                # restart the baysian learning for the remaining series
-                # this is necessary since otherwise NaN values
+                # Restart the baysian learning for the remaining series.
+                # This is necessary since otherwise NaN values
                 # will be inferred for all future days, after seeing
                 # a single (smoothed) zero value.
+                #
+                # We understand that restarting the posteriors with the
+                # initial prior may incur a start-up artifact as the posterior
+                # restabilizes, but we believe it's the current best
+                # solution for municipalities that have smoothed cases and
+                # deaths that dip down to zero, but then start to increase
+                # again.
+
                 posteriors[current_day] = prior0
             else:
                 posteriors[current_day] = numerator / denominator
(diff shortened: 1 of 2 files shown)


pyseir/inference/infer_rt.py
```diff
@@ -468,6 +475,10 @@ def infer_all(self, plot=False, shift_deaths=0):
                     for col in df_all.columns:
                         if timeseries_type.value in col:
                             df_all[col] = df_all[col].shift(shift_in_days)
+                            # Extend death and hopitalization rt signals beyond shift to avoid sudden jumps in composit metric.
+                            df_all[col] = df_all[col].interpolate(
```
jeremiq, line 479: some handy docs: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.interpolate.html

BrettBoval: I think we need to put more notes in here about how this call is doing double duty and being a little abused to do extrapolation too. 

For example, it does linear interpolation between two points (so 1, np.nan, 3 becomes 1,2,3) but it does forward fill for extrapolation (so 1, 2 , np.nan becomes 1, 2, 2). I think those behaviors are reasonable, but it's not clear from the code (or the docstring of the function) that it is following two separate paths.

So two options:
- Keep code as is and just add a more detailed comment
- Make it explicit by calling it interpolate twice with the 'limit_area' flag switched. This would also give us finer control over rules for extrapolation verse rules for interpolation. We could also set different 'limit' values for the max number of nan we are willing to replace. Depending on how the 'limit_area'='outer' is designed this might be a bit slower.

image

pyseir/inference/infer_rt.py
```diff
@@ -468,6 +475,10 @@ def infer_all(self, plot=False, shift_deaths=0):
                     for col in df_all.columns:
                         if timeseries_type.value in col:
                             df_all[col] = df_all[col].shift(shift_in_days)
+                            # Extend death and hopitalization rt signals beyond shift to avoid sudden jumps in composit metric.
```
BrettBoval, line 478: Do you have a sense of how often this will happen? Would adding a logging.debug call with 'we interpolated X points' just swamp any possible usefulness? I have haven't seen any instances where we have feature specific loggers, so I'm guessing this might swamp anyone's reasonable expectation of the general debug logger.

pyseir/inference/infer_rt.py
```diff
@@ -361,7 +361,14 @@ def get_posteriors(self, timeseries_type, plot=False):
             denominator = np.sum(numerator)
 
             # Execute full Bayes' Rule
-            posteriors[current_day] = numerator / denominator
+            if denominator == 0:
+                # restart the baysian learning for the remaining series
+                # this is necessary since otherwise NaN values
+                # will be inferred for all future days, after seeing
+                # a single (smoothed) zero value.
+                posteriors[current_day] = prior0
```
BrettBoval, line 369: Can you add another note that captures 'We understand that restarting the posteriors with the initial prior may incur a start-up artifact as the posterior restabilizes, but that we believe it's the current best solution for edge cases like X.' Let's give future selves and others more hints into the quirks of the system.

tashwoods: But I guess this implies we aren't going to test artificially increasing the denominator to say 1E-10?

pyseir/inference/infer_rt.py
tashwoods: Oh, nice catch Brett. Is that the behavior we want for forward extrapolation? I think it would be better to have it do some version of a linear extrapolation...

Also, just pinning I can imagine edge cases where linearly interpolating between these points won't model the true distribution well, but I don't think we should try to deal with that now.

tashwoods: But also, if while we are adding information, are you back-extrapolating? (e.g. when the death series starts after the case series and we see a sudden uptick in the inferred Rt when the death series 'begins')

tashwoods: [Rt_results__Kansas__20.pdf](https://github.com/covid-projections/covid-data-model/files/4664432/Rt_results__Kansas__20.pdf)

pyseir/inference/infer_rt.py
jeremiq: This will happen for all death series, since the timeshifting is almost always bigger than 0. I'll log it though.

pyseir/inference/infer_rt.py
jeremiq: I'll make it clearer exactly what the behavior is here. I evaluated other ways of extrapolating the end of the series but didn't see anything that was obviously better.

PR #438, opened 2020-05-22
Add more build_processed_data tests

TomGoBravo:
This PR adds some tests so I can more easily fix
https://trello.com/c/tQj1yUjh/145-cds-data-source-missing-proper-county-level-data
without breaking existing code.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/test/build_processed_data_test.py
+++ b/test/build_processed_data_test.py
@@ -3,38 +3,55 @@ from unittest.mock import patch
 import pytest
 
 from libs import build_processed_dataset, validate_results
-from libs.datasets import CommonFields
+from libs.datasets import CommonFields, CDSDataset, CovidTrackingDataSource
 import pandas as pd
+from libs.datasets.results_schema import CUMULATIVE_POSITIVE_TESTS, CUMULATIVE_NEGATIVE_TESTS
 
 from libs.enums import Intervention
+from more_itertools import one
 
 
 def test_get_usa_by_states_df():
-    empty_df_with_state = pd.DataFrame([], columns=['state',])
-    with patch('libs.build_processed_dataset.get_state_projections_df', return_value=empty_df_with_state) as mock:
-        df = build_processed_dataset.get_usa_by_states_df(input_dir="/tmp", intervention_type=Intervention.OBSERVED_INTERVENTION.value)
-    validate_results.validate_states_df("MA", df)
-    df.set_index([CommonFields.DATE, CommonFields.STATE], inplace=True)
-    print(df.info())
-    print(df.head())
-    assert df.loc[("2020-04-01", "MA"), CommonFields.POSITIVE_TESTS] > 0
+    empty_df_with_state = pd.DataFrame([], columns=["state",])
+    with patch(
+        "libs.build_processed_dataset.get_state_projections_df", return_value=empty_df_with_state
+    ) as mock:
+        df = build_processed_dataset.get_usa_by_states_df(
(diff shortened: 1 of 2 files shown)


TomGoBravo:
[CDS Data Source missing proper county level data](https://trello.com/c/tQj1yUjh/145-cds-data-source-missing-proper-county-level-data)

test/build_processed_data_test.py
```diff
@@ -1,6 +1,57 @@
...
+def test_get_testing_df_by_state():
+    positive_field = CovidTrackingDataSource.Fields.POSITIVE_TESTS
+    negative_field = CovidTrackingDataSource.Fields.NEGATIVE_TESTS
+    df = build_processed_dataset.get_testing_timeseries_by_state("MA")
+    assert positive_field in df.columns
+    assert negative_field in df.columns
+    # TODO(tom): have build_processed_dataset return a real datetime dtype so callers don't need to parse a string.
+    df[CommonFields.DATE] = pd.to_datetime(df[CommonFields.DATE], format="%m/%d/%y")
+    df.set_index(CommonFields.DATE, inplace=True)
+    # No joke, our only source of state-level timeseries testing data is CDSDataset and it has NaN for MA until April 6.
```
ghop02, line 38: Do you mean covid tracking ?

ghop02: Also I wonder if we can just import this from the MA data - I feel like they reported it?

test/build_processed_data_test.py
```diff
@@ -1,6 +1,57 @@
...
+    assert 0 < df.at["2020-04-10", positive_field] < df.at["2020-05-01", positive_field]
+    assert 0 < df.at["2020-04-10", negative_field] < df.at["2020-05-01", negative_field]
+
+
+# Check San Francisco/06075 and Houston (Harris County, TX)/48201
```
ghop02, line 43: Is there a reason you chose these couple counties?

test/build_processed_data_test.py
```diff
@@ -1,6 +1,57 @@
...
+import pandas as pd
+from libs.datasets.results_schema import CUMULATIVE_POSITIVE_TESTS, CUMULATIVE_NEGATIVE_TESTS
+
+from libs.enums import Intervention
+from more_itertools import one
+
+
+def test_get_usa_by_states_df():
+    empty_df_with_state = pd.DataFrame([], columns=["state",])
+    with patch(
```
ghop02, line 16: this is totally fine and no need to change, just wanted to enumerate a couple other pytest fixture options:
 - monkeypatch fixture that pytest has [included](https://docs.pytest.org/en/latest/monkeypatch.html)
 - i've liked [pytest-mock](https://pypi.org/project/pytest-mock/)

an di

test/build_processed_data_test.py
TomGoBravo: The next PR will be tricky... `grep 20200401,MA ../covid-data-public/data/covid-tracking/covid_tracking_states.csv` does find positive and negative values, but the way the data is combined this test fails when it is produced from combined data. Lets get this merged and move on to that next...

test/build_processed_data_test.py
TomGoBravo: Rooting for the home team... spot checking the two USA counties I've lived in the longest.

test/build_processed_data_test.py
TomGoBravo: Thank you for the pointers. I find fixtures a bit voodoo but could get into `with monkeypatch.context() as m:`

PR #439, opened 2020-05-22
build_processed_data gets test data from combined_datasets

TomGoBravo:
## Background

https://trello.com/c/tQj1yUjh/145-cds-data-source-missing-proper-county-level-data

https://github.com/covid-projections/covid-data-model/pull/438 added tests.

## This PR

Changes build_processed_data to get COVID negative/positive test data from combined_datasets instead of directly from the data source classes. It mostly transforms the columns back to their original form to reduce the chance of accidentally breaking code that depends on the current return value.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/build_processed_dataset.py
+++ b/libs/build_processed_dataset.py
@@ -4,6 +4,7 @@ import logging
 import sentry_sdk
 from functools import lru_cache
 
+from libs.datasets.combined_datasets import build_us_timeseries_with_all_fields
 from libs.us_state_abbrev import US_STATE_ABBREV
 from libs.us_state_abbrev import abbrev_us_state
 from libs.us_state_abbrev import us_fips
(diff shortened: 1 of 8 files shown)


TomGoBravo:
[CDS Data Source missing proper county level data](https://trello.com/c/tQj1yUjh/145-cds-data-source-missing-proper-county-level-data)

libs/build_processed_dataset.py
```diff
@@ -40,55 +41,48 @@ def _get_interventions_df():
...
-    fips_testing_df = fips_testing_df[~fips_testing_df.index.duplicated(keep="last")]
-    if before != len(fips_testing_df):
-        _logger.warning(
-            f"Testing DF contained duplicate rows for {fips}: {before} -> {len(fips_testing_df)}"
-        )
-        sentry_sdk.capture_message(f"Testing DF contained duplicate rows for {fips}")
-    return fips_testing_df
+    """Called by generate_api"""
+    all_data = build_us_timeseries_with_all_fields().data
+    print(all_data.info())
```
ghop02, line 72: print left in

libs/build_processed_dataset.py
```diff
@@ -175,18 +169,26 @@ def get_usa_by_county_with_projection_df(input_dir, intervention_type):
     return counties
 
 
-def get_usa_by_states_df(input_dir, intervention_type):
+def get_usa_by_states_df(input_dir: str, intervention_type):
```
ghop02, line 172: is intervention_type here a string too? or `Intervention`

libs/datasets/dataset_utils.py
```diff
@@ -334,9 +334,16 @@ def fill_fields_with_data_source(
...
+        print("Intersection")
         print(new_data.loc[new_data_in_existing_df, columns_to_fill])
+        print("new data not in existing")
+        print(new_data.loc[~new_data.index.isin(existing_df.index), columns_to_fill])
+        if columns_to_fill[0] in existing_df.columns:
+            print("existing not in new")
+            print(existing_df.loc[~existing_df.index.isin(new_data.index), columns_to_fill])
         existing_in_new = sum(existing_df_in_new_data)
         new_in_existing = sum(new_data_in_existing_df)
+        # If this trips consider adding verify_integrity=True to the set_index calls above.
```
ghop02, line 346: 👍

libs/datasets/sources/cds_dataset.py
```diff
@@ -111,21 +124,27 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
...
         fips_data = dataset_utils.build_fips_data_frame()
         data = dataset_utils.add_fips_using_county(data, fips_data)
+        no_fips = data[CommonFields.FIPS].isna()
+        if no_fips.sum() > 0:
+            logging.warning(f"Removing rows without fips id: {str(data.loc[no_fips])}")
+            data = data.loc[~no_fips]
+
+        data.set_index(["date", "fips"], inplace=True)
+        if data.index.has_duplicates:
+            logging.warning(f"Removing duplicates: {str(data.index.duplicated(keep=False))}")
```
ghop02, line 141: Do we want sentry errors if there are too many duplicates?

libs/datasets/sources/cds_dataset.py
```diff
@@ -111,21 +124,27 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
...
+            data = data.loc[~data.index.duplicated(keep=False)]
```
ghop02, line 142: What does keep=False do?

test/build_processed_data_test.py
```diff
@@ -1,6 +1,57 @@
...
+    assert 0 < df.at["2020-04-10", positive_field] < df.at["2020-05-01", positive_field]
+    assert 0 < df.at["2020-04-10", negative_field] < df.at["2020-05-01", negative_field]
+
+
+# Check San Francisco/06075 and Houston (Harris County, TX)/48201
+@pytest.mark.parametrize(
+    "fips", ["06075", "48201"],
+)
+def test_get_testing_by_fips(fips):
+    df = build_processed_dataset.get_testing_timeseries_by_fips(fips)
```
ghop02, line 48: I know this uses the combined datasets under the hood now, but do you think we should explicitly check combined dataset county test data to make sure we have at least some values?

libs/datasets/sources/cds_dataset.py
TomGoBravo: The tests pass with this as logging.fatal so I'm guessing there are no dups. I changed to logging.error which according to https://docs.sentry.io/platforms/python/logging/ makes a sentry event by default.
dealing with alerts well is a road map level TODO :-)

libs/datasets/sources/cds_dataset.py
TomGoBravo: keep=False makes the log message easier to read because it includes all of the duplicated row instead of just the first or last instance of each duplicate.

libs/build_processed_dataset.py
TomGoBravo: fixed

test/build_processed_data_test.py
TomGoBravo: Good idea. I added a check of cases and deaths. there isn't much data about tests at the county level.

PR #440, opened 2020-05-22
Add Rt based suppression policy

erccarls:

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/inference/model_fitter.py
+++ b/pyseir/inference/model_fitter.py
@@ -9,6 +9,7 @@ import numpy as np
 # from jax import numpy as np
 import os
 import us
+from enum import Enum
 import dill as pickle
 import numpy as np
 from pprint import pformat
@@ -34,6 +35,11 @@ from pyseir.inference.fit_results import load_inference_result
 
 def calc_chi_sq(obs, predicted, stddev):
     return np.sum((obs - predicted) ** 2 / stddev ** 2)
+
+
+class SuppressionPolicy(Enum):
+    TWO_STAGE = 'two_stage'
+    RT = 'rt'
 
 
 class ModelFitter:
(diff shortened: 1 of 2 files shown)


pyseir/inference/model_fitter.py
```diff
@@ -366,9 +393,17 @@ def run_model(self, R0, eps, t_break, log10_I_initial):
...
+        if self.suppression_policy is SuppressionPolicy.TWO_STAGE:
+            suppression_policy = suppression_policies.generate_two_step_policy(
+                self.t_list, eps, t_break
+            )
+        elif self.suppression_policy is SuppressionPolicy.RT:
+            suppression_policy = suppression_policies.generate_rt_inference_policy(
+                t0=self.ref_date + timedelta(days=t0),
+                fips=self.fips,
+                future_suppression='current',
+                # R0=R0 # TODO: TBD if this should be fit...
```
erccarls, line 405: Here we need to probably instead make this a scalar which makes`R_seir(t) = 1+ alpha * (R(t) - 1)`

pyseir/models/suppression_policies.py
```diff
@@ -22,6 +22,75 @@
...
+    if R0 is None:
+        R0 = get_r0_from_rt(fips)
+    try:
+        df_rt = fit_results.load_Rt_result(fips)
+    except ValueError:
+        df_rt = fit_results.load_Rt_result(fips[:2])
+
+    df_rt = df_rt[df_rt['Rt_MAP_composite'].notnull()]
+
+    days = [(t - t0).days for t in df_rt['Rt_MAP_composite'].index.to_pydatetime()]
```
erccarls, line 80: I need to think more about whether this shift is right...  

Rt is known at a given time, but we may want to allow for a small shift to account for delays in that vs the simulation.  Right now this is shifting the entire Rt timeseries along with the simulation, which may not be what we want (i.e. we may want to allow it to move *within* the simulation).

PR #442, opened 2020-05-23
Adding a command line function to download the latest model output artifact from github

ghop02:
Added in a helper function to make it easy to download the latest model output snapshot - just makes it easier to debug api generation locally

### DIFF
--- a/cli/utils.py
+++ b/cli/utils.py
@@ -1,0 +1,25 @@
+import logging
+import pathlib
+import click
+from libs import github_utils
+
+_logger = logging.getLogger(__name__)
+
+
+@click.group("utils")
+def main():
+    pass
+
+
+@main.command()
+@click.argument("run-number", type=int, required=False)
+@click.option(
+    "--github-token",
+    envvar="GITHUB_TOKEN",
+    required=True,
+    help="Github Token, can be an option or set as env variable GITHUB_TOKEN",
+)
+@click.option("--output-dir", "-o", type=pathlib.Path, default=pathlib.Path("."))
+def download_model_artifact(github_token, run_number, output_dir):
+    """Download model output from github action publish and deploy workflow. """
+    github_utils.download_model_artifact(github_token, output_dir, run_number=run_number)
--- a/libs/github_utils.py
+++ b/libs/github_utils.py
@@ -1,0 +1,85 @@
+from typing import Tuple
+import shutil
+import tempfile
+import io
+import zipfile
+import pathlib
+import logging
+import requests
+
+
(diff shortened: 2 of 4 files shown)


cli/utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+@click.group("utils")
+def main():
+    pass
+
+
+@main.command()
+@click.argument("run-number", type=int, required=False)
+@click.option(
+    "--github-token",
```
TomGoBravo, line 17: Thinking aloud... I'm not sure why, but my impression is values in the env are harder to steal than in args... though I guess if someone can read your command line history your machine is already owned and they can likely read the `export GITHUB_TOKEN=...`.

libs/github_utils.py
```diff
@@ -0,0 +1,85 @@
...
+        # Workflows are returned latest first.
+        run = runs[0]
+    else:
+        # TODO: Only requests the first page of workflows, if run id is not from the
+        # 30 most recent runs, it will not be able to be found.
+        for run in runs:
+            if run["run_number"] == run_number:
+                break
+        else:
+            raise Exception("Could not find workflow with run number {run_number}.")
```
TomGoBravo, line 35: add something in this message about only supporting recent runs so users don't need to dig up the TODO in the source.

libs/github_utils.py
```diff
@@ -0,0 +1,85 @@
...
+def download_model_artifact(
+    github_token: str, output_dir: pathlib.Path, run_number: int = None, overwrite=True
+):
+    """Downloads and extracts model output artifact.
+
+    Args:
+        github_token: GitHub token with read access.
+        output_dir: Output directory to extract artifact to.
+        run_number: If specified, will download the specific run number.
+        overwrite: If true, removes and overwrites exsting output artifact.
```
TomGoBravo, line 60: typo: existing

libs/github_utils.py
```diff
@@ -0,0 +1,85 @@
...
+        raise AssertionError(
+            "Multiple artifacts detected, missing logic to choose correct artifact."
+        )
+
+    artifact = artifacts[0]
+    return artifact["archive_download_url"], run["run_number"]
+
+
+def download_model_artifact(
+    github_token: str, output_dir: pathlib.Path, run_number: int = None, overwrite=True
```
TomGoBravo, line 52: I strongly prefer having an explicit `--overwrite` to enable less safe behavior.

TomGoBravo:
I tried this tool out and it worked for me. I'd still like `--overwrite` to default to not overwriting what is already in the directory.

PR #447, opened 2020-05-26
WIP: Third stage tests

tashwoods:

### Please Check
- [ ] Are tests passing?


BrettBoval:
The 'how impacted will we be in the future' code needs the final_eps stuff to forecast. The seir model needs that whole eps timeseries (which it should get from the three stage fit). This just fixes the "suppression looks at the outdating code".

PR #451, opened 2020-05-27
Feature/save webui adaptor for county icu

jeremiq:
This PR is my first attempt at getting actual county-level hospitalization and icu data to be used for our icu headroom metric, when available. I've added a new datasource to our combined dataset to collect this data, and I've re-purposed some existing logic to convert from cumulative values to current values. 

I suspect we may want to update the logic in webui_data_adaptor further, but I'd like some input on where to take this next. 

### Please Check
- [ YES] Are tests passing?


pyseir/cli.py
```diff
@@ -63,7 +63,7 @@ def entry_point():
 
 @entry_point.command()
 def download_data():
-    cache_all_data()
```
BrettBoval, line 66: I would rather we clean up cache_all_data instead of replacing it with cache_county_case_data. We could just comment out the mobility and cache_public_implementations_data lines. I know they end up the same, but the cache_all_data function name is more informative, and in the future I'd like it to cache all relevant data.

jeremiq: I know what you mean! I started implementing exactly what you described, but I decided that it might be a bit deceptive/confusing, since there would then be data sources in `load_data` that would not be considered `all_data`. 

If you strongly prefer just refactoring `cache_all_data` to only cache the county_case data, I will, though.

pyseir/load_data.py
```diff
@@ -522,16 +523,29 @@ def get_current_hospitalized(state, t0, category):
...
+def get_current_hospitalized_for_county(fips: str, t0: datetime, category: str):
+    df = (
+        get_hospitalization_data()
+        .get_subset(AggregationLevel.COUNTY, country="USA", fips=fips)
+        .get_data(AggregationLevel.COUNTY, country="USA", fips=fips)
+    )
+    return _get_current_hospitalized(df, t0, category)
+
+
+def _get_current_hospitalized(df: pd.DataFrame, t0: datetime, category: str):
```
BrettBoval, line 543: I like pulling out the loading of the dfs into those two wrappers. We should include in the docstring for this function the fact that we won't provide results until there are 3 cumulatives. I could see that being a gotcha that would be easier to catch in documentation.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -176,9 +219,13 @@ def map_fips(self, fips):
             output_model[schema.INFECTED_B] = hosp_rescaling_factor * np.interp(
                 t_list_downsampled, t_list, output_for_policy["HGen"]["ci_50"]
             )  # Hosp General
-            output_model[schema.INFECTED_C] = icu_rescaling_factor * np.interp(
-                t_list_downsampled, t_list, output_for_policy["HICU"]["ci_50"]
-            )  # Hosp ICU
+
```
BrettBoval, line 222: All these conversions should be done on the pyseir side somewhere. There's a lot of scaling/math happening in a output mapper script. Apply icu_rescaling_fator upstream somewhere closer to the creation point (where the context of what is going on is clearer). Probably a different PR.

pyseir/load_data.py
BrettBoval: And the fact that it will convert cumulatives to current using a small model. It will mix derived and presented values.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -60,6 +63,28 @@ def __init__(
         self.df_whitelist = load_data.load_whitelist()
         self.df_whitelist = self.df_whitelist[self.df_whitelist["inference_ok"] == True]
 
+    @staticmethod
+    def _get_county_hospitalization_from_actuals(fips, t0_simulation):
```
BrettBoval, line 67: This method name is misleading. If a fips has cumulative, but not current, this will derive an estimate of current that is not an 'actual' from the datasources.

pyseir/load_data.py
```diff
@@ -522,16 +523,29 @@ def get_current_hospitalized(state, t0, category):
...
+    MIN_DATAPOINTS_TO_CONVERT = 3  # We wait until the third datapoint to have 2 deltas to forecast
+    ALLOWED_CATEGORIES = ["icu", "hospitalized"]
```
TomGoBravo, line 545: An Enum is much cleaner, for example:
 https://github.com/covid-projections/covid-data-model/blob/31656d9272bf21a9b9b24fdd92f34d5d413054fa/libs/datasets/dataset_utils.py#L19-L22

pyseir/load_data.py
TomGoBravo, line 543: I think this function expects df to contain data from exactly one region (state or county). please clarify in a docstring.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -60,6 +63,28 @@ def __init__(
...
+    def _get_population(self, fips: str):
+        if len(fips) == 5:
+            return self.population_data.get_record_for_fips(fips)[CommonFields.POPULATION]
+        return self.population_data.get_record_for_state(self.state_abbreviation)[
+            CommonFields.POPULATION
+        ]
+
     def map_fips(self, fips):
         """
         For a given county fips code, generate the CAN UI output format.
```
TomGoBravo, line 90: the implementation looks like it may handle a 2 char state fips code. please clarify the docstring while you are deep in the code.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -69,74 +94,92 @@ def map_fips(self, fips):
         fips: str
             County FIPS code to map.
         """
-        logging.info(f"Mapping output to WebUI for {self.state}, {fips}")
+        log.info("Mapping output to WebUI for %s, %s", self.state, fips)
```
TomGoBravo, line 97: To the best of my knowledge f"" string formatting is the modern Python way and % formatting is out of style.
For logging I much prefer a completely static message string and extra={} for variables or to be even more cutting edge try
```
log = structlog.get_logger()
log.info("Mapping output to WebUI", state=self.state, fips=fips)
```

pyseir/deployment/webui_data_adaptor_v1.py
jeremiq: good catch, I didn't even notice this doc string, but it's clearly wrong.

pyseir/deployment/webui_data_adaptor_v1.py
jeremiq: oO! I just saw that you started layering in structured logging. I'll convert over to that. 

Regarding the f-strings vs % formatting, I think logging is still the exception -- but I'm sure the preferred style still varies: https://docs.python.org/3/howto/logging.html#optimization

pyseir/deployment/webui_data_adaptor_v1.py
jeremiq: I'm going to save this for another PR, but I completely agree. This conversion needs to live much closer to the model -- the interpolation and integer conversion too, IMO.

pyseir/load_data.py
jeremiq: added a doc string!

pyseir/deployment/webui_data_adaptor_v1.py
jeremiq: Updated, added a doc string too.

pyseir/deployment/webui_data_adaptor_v1.py
jeremiq: Hmm, I can't actually get this to work with structured logging. I tried, but keep getting this error: 
```
Traceback (most recent call last):
  File "/Users/flyrrecovery/.pyenv/versions/covid-data-model/bin/pyseir", line 11, in <module>
    load_entry_point('pyseir', 'console_scripts', 'pyseir')()
  File "/Users/flyrrecovery/.pyenv/versions/3.7.7/envs/covid-data-model/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/flyrrecovery/.pyenv/versions/3.7.7/envs/covid-data-model/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/flyrrecovery/.pyenv/versions/3.7.7/envs/covid-data-model/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/flyrrecovery/.pyenv/versions/3.7.7/envs/covid-data-model/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/flyrrecovery/.pyenv/versions/3.7.7/envs/covid-data-model/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/flyrrecovery/src/github/covid-data-model/pyseir/cli.py", line 435, in build_all
Revert "Switches webui adaptor to structured logging."
    states_only=states_only,
  File "/Users/flyrrecovery/src/github/covid-data-model/pyseir/cli.py", line 211, in _build_all_for_states
    p.map(states_only_func, states)
  File "/Users/flyrrecovery/.pyenv/versions/3.7.7/lib/python3.7/multiprocessing/pool.py", line 268, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/Users/flyrrecovery/.pyenv/versions/3.7.7/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
TypeError: msg() got an unexpected keyword argument 'state'
```

jeremiq:
This code clearly needs more (any!) testing. I'd like to get this in and then follow up with a bit more code cleanup and testing separately.

pyseir/load_data.py
```diff
@@ -654,7 +696,7 @@ def get_average_dwell_time(category):
     # TODO: This value is temporarily in this limited scope.
     # Will be added into the params once I decide on some refactoring.
     params["hospitalization_length_of_stay_icu_avg"] = 8.6
-    if category == "hospitalized":
+    if str(category) == "hospitalized":
```
TomGoBravo, line 699: embrace the enum by avoiding str operations: `category == HospitalizationCategory.HOSPITALIZED:`

jeremiq: fixed in the latest commit!

PR #455, opened 2020-05-28
Feature/more structured logging

jeremiq:
I'm not sure this is the final form of this, but adding the renderer allowed me to get around the exceptions I was seeing around key:value parsing before.

### DIFF
--- a/pyseir/cli.py
+++ b/pyseir/cli.py
@@ -66,6 +66,8 @@ def entry_point():
             structlog.stdlib.add_log_level,  # required before SentryProcessor()
             # sentry_sdk creates events for level >= ERROR and keeps level >= INFO as breadcrumbs.
             SentryProcessor(level=logging.INFO),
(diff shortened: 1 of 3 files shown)


pyseir/cli.py
```diff
@@ -66,6 +66,8 @@ def entry_point():
             structlog.stdlib.add_log_level,  # required before SentryProcessor()
             # sentry_sdk creates events for level >= ERROR and keeps level >= INFO as breadcrumbs.
             SentryProcessor(level=logging.INFO),
+            structlog.processors.TimeStamper(fmt="iso"),
```
TomGoBravo, line 69: Please add to https://github.com/covid-projections/covid-data-model/blob/master/run.py#L34-L38 too. My mistake for not factoring it out.

jeremiq: good call! I hadn't thought about the other entry point. added in the latest commit.

PR #456, opened 2020-05-29
Automatically trigger covid-projections update-snapshot job after snapshot build completes.

mikelehen:
Fixes https://trello.com/c/6shjpYaM

### DIFF
--- a/tools/maybe-trigger-web-snapshot-update.sh
+++ b/tools/maybe-trigger-web-snapshot-update.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+# maybe-trigger-web-snapshot-update.sh
+#
+# Helper script called at the end of the deploy_api.yml workflow to potentially
+# trigger the update-snapshot.yml workflow in the covid-projections repo to
+# update to the newly generated snapshot.
+
+set -o nounset
+set -o errexit
+
+CMD=$0
+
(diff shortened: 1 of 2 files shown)


tools/maybe-trigger-web-snapshot-update.sh
```diff
@@ -0,0 +1,70 @@
...
+    COVID_DATA_PUBLIC_REF=$3
+  fi
+
+  if ! [[ $SNAPSHOT_ID =~ ^[0-9]+$ ]] ; then
+    echo "Error: Specified Snapshot ID ($SNAPSHOT_ID) should be a plain number."
+    echo
+    exit_with_usage
+  fi
+
+  if [[ $COVID_DATA_MODEL_REF != "master" ]] || [[ $COVID_DATA_PUBLIC_REF != "master" ]]; then
```
goldblatt, line 33: why is this the public ref and the non public ref? i've been a lil confused why the public ref distinction exists as it never worked for me

mikelehen: Not 100% sure what you mean by non-public...

These are the git refs of covid-data-model and covid-data-public that were used to build the snapshot.  These are _usually_ "master" but can be changed to point elsewhere when people are manually testing their code in branches, etc.

PR #457, opened 2020-05-29
Added Jupyter Lab to Requirements.txt

BrettBoval:
Thought I'd bring up a quick package managing question (everyone's favorite thing).

I added the newer version of jupyter (jupyterlab) to requirements.

If it was a heavier lift I'd say that all the debug tools like jupyter/jupyterlab don't need to be in the remote build, but I don't think it's too bad atm.

Also, @ghop02 did we get burned sometime in the past and decided to pin everything? I don't particularly want to sign up for us using poetry or pipenv. Just curious.

### Please Check
- [YES] Are tests passing?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ cycler==0.10.0
 idna==2.9
 isort==4.3.21
 kiwisolver==1.1.0
-lazy-object-proxy==1.3.1
+lazy-object-proxy==1.4.1
 GitPython==3.1.0
 numpy==1.18.2
 matplotlib==3.2.1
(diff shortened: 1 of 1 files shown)


ghop02:
@BrettBoval we did run into a couple small version dependency issues in the past.  I generally prefer to keep requirements.txt version locked, just to minimize moving pieces for everyone, but agree for things like jupyterlab i don't have as strong of an opinion.  

I continually go back and forth on pipenv/poetry (haven't actually used poetry).  It definitely seems like it could be useful. pipenv wasn't very well maintained for a while (but seems like its more active as of late?) but at the same time, feels like poetry is gaining more general support. overall /shrug haha

BrettBoval:
I pinned jupyterlab, bumped another dependency that was complaining, and added a trello card backlog to consolidate on poetry (right now we have semi-orphaned environment.yaml).

PR #459, opened 2020-06-02
Model/add testing compensation to rt inference

erccarls:
Mistakenly removed this in last PR. Reverting.


pyseir/inference/infer_rt.py
```diff
@@ -512,6 +512,20 @@ def infer_all(self, plot=True, shift_deaths=0):
         if plot and df_all is not None:
             plt.figure(figsize=(10, 6))
 
+            if "Rt_MAP_composite" in df_all:
```
tashwoods, line 515: :D ty!

PR #463, opened 2020-06-02
Adds pre-commit hook for stripping notebook output.

jeremiq:


jeremiq:
After this PR is merged everyone will need to run either: 

```
make setup-dev
```

or simply
```
pre-commit install
```

in order to get the new hook definition.

jeremiq:
Good call, @ghop02 . Done in the latest commit.

jeremiq:
This is failing because of the data issue identified by Brett and Tom. I'll come back to this as soon as the fix is found for the combined dataset

PR #466, opened 2020-06-02
use covid tracking date column, revert data pins

TomGoBravo:
## Background

See https://trello.com/c/1HA3Egdm/189-adding-jupyterlab-appears-to-break-build-all-at-the-moment and https://covidactnow.slack.com/archives/C0110QQAPJA/p1591105732043200

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/datasets/sources/covid_tracking_source.py
+++ b/libs/datasets/sources/covid_tracking_source.py
@@ -111,9 +111,7 @@ class CovidTrackingDataSource(data_source.DataSource):
         data[cls.Fields.COUNTY] = None
         data[cls.Fields.COUNTRY] = "USA"
         data[cls.Fields.AGGREGATE_LEVEL] = AggregationLevel.STATE.value
-        # Date checked is the time that the data is actually updated.
-        # assigning the date field as the date floor of that day.
-        data[cls.Fields.DATE] = data[cls.Fields.DATE_CHECKED].dt.tz_localize(None).dt.floor("D")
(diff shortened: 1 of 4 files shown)


libs/datasets/sources/covid_tracking_source.py
```diff
@@ -111,9 +111,7 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
         data[cls.Fields.COUNTY] = None
         data[cls.Fields.COUNTRY] = "USA"
         data[cls.Fields.AGGREGATE_LEVEL] = AggregationLevel.STATE.value
-        # Date checked is the time that the data is actually updated.
-        # assigning the date field as the date floor of that day.
-        data[cls.Fields.DATE] = data[cls.Fields.DATE_CHECKED].dt.tz_localize(None).dt.floor("D")
+        data[cls.Fields.DATE] = pd.to_datetime(data[cls.Fields.DATE], format="%Y%m%d")
```
ghop02, line 114: Could this be fixed by adding `cls.Fields.DATE` to `parse_dates`? or does it not recognize the date format

TomGoBravo:
[Adding Jupyterlab Appears to Break Build-All at the Moment](https://trello.com/c/1HA3Egdm/189-adding-jupyterlab-appears-to-break-build-all-at-the-moment)

libs/datasets/sources/covid_tracking_source.py
TomGoBravo: Fixed in c126577

PR #467, opened 2020-06-02
Feature/delphi kinsa notebooks

jeremiq:


jeremiq:
good call, I'll change these

PR #470, opened 2020-06-03
SEIR Third Stage 

tashwoods:
This PR creates a third stage for the SEIR model

### Please Check
- [ so far! ] Are tests passing?

### DIFF
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -9,7 +9,7 @@ on:
   schedule:
    - cron: '30 0,12 * * *'
 
-  push:
+  #push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
@@ -17,7 +17,7 @@ on:
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'feature/third-stage-clean'
+  COVID_DATA_MODEL_REF: 'master'
 
   # To pin to an old data sets, put the branch/tag/commit here:
   COVID_DATA_PUBLIC_REF: 'master'
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -9,7 +9,7 @@ on:
   schedule:
    - cron: '30 0,12 * * *'
 
-  # push:
+  #push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
@@ -17,10 +17,11 @@ on:
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  COVID_DATA_MODEL_REF: 'feature/third-stage-clean'
 
   # To pin to an old data sets, put the branch/tag/commit here:
   COVID_DATA_PUBLIC_REF: 'master'
+  #TODO(tom): Put back to master once latest breakage addressed.
 
   # S3 Bucket (used by s3-sync-action tasks) to store final API snapshot.
   AWS_S3_BUCKET: 'data.covidactnow.org'
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -9,7 +9,7 @@ on:
   schedule:
    - cron: '30 0,12 * * *'
 
-  #push:
+  # push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
@@ -21,7 +21,6 @@ env:
 
   # To pin to an old data sets, put the branch/tag/commit here:
   COVID_DATA_PUBLIC_REF: 'master'
-  #TODO(tom): Put back to master once latest breakage addressed.
 
   # S3 Bucket (used by s3-sync-action tasks) to store final API snapshot.
   AWS_S3_BUCKET: 'data.covidactnow.org'
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -9,7 +9,7 @@ on:
   schedule:
    - cron: '30 0,12 * * *'
 
-  #push:
+  push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
--- a/pyseir/ensembles/ensemble_runner.py
+++ b/pyseir/ensembles/ensemble_runner.py
@@ -207,9 +207,12 @@ class EnsembleRunner:
 
             # Rescale state values to the county population and replace county
             # specific params.
+            # TODO: get_average_seir_parameters should return the analytic solution when available
+            # right now it runs an average over the ensemble (with N_samples not consistently set
+            # across the code base).
             default_params = ParameterEnsembleGenerator(
                 self.fips,
-                N_samples=1,
+                N_samples=1,  # change back to 250?
                 t_list=model.t_list,
                 suppression_policy=model.suppression_policy,
             ).get_average_seir_parameters()
@@ -230,10 +233,11 @@ class EnsembleRunner:
         else:
             eps_final = sp.get_future_suppression_from_r0(inferred_params["R0"], scenario=scenario)
 
-        model.suppression_policy = sp.generate_two_step_policy(
-            self.t_list,
+        model.suppression_policy = sp.get_epsilon_interpolator(
             eps=inferred_params["eps"],
             t_break=inferred_params["t_break"],
+            eps2=inferred_params["eps2"],  # check
+            t_delta_phases=inferred_params["t_delta_phases"],
             t_break_final=(
                 datetime.datetime.today()
                 - datetime.datetime.fromisoformat(inferred_params["t0_date"])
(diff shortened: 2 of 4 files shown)


pyseir/inference/model_fitter.py
```diff
@@ -99,7 +106,16 @@ class ModelFitter:
...
-    PARAM_SETS = {("HI",): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90])}
+    PARAM_SETS = {
+        ("HI"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
+        ("AK"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
+        ("MT"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
+        ("ID"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
+        ("LA"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
+        ("ND"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
+        ("WV"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
+        ("WY"): dict(eps=0.25, t0=75, t_break=10, limit_t0=[50, 90]),
```
jeremiq, line 117: Since the dict being used is the same across all of these states, I'd pull it out into a variable to avoid the duplication.

pyseir/inference/model_fitter.py
```diff
@@ -121,6 +137,8 @@ def __init__(
 
         self.fips = fips
         self.ref_date = ref_date
+        self.max_fit_date = (dt.date.today() - timedelta(days=7) - ref_date.date()).days  # natasha
```
jeremiq, line 140: extract the 7 day time delta into a named constant.

pyseir/inference/model_fitter.py
tashwoods: Good point, but I am not sure we won't want to update these. So let's make that a future decision : )

pyseir/inference/model_fitter.py
tashwoods: Ah, I forgot to delete that. I refactored the code

pyseir/inference/model_fitter.py
```diff
@@ -418,12 +459,28 @@ def _fit_seir(self, R0, t0, eps, t_break, test_fraction, hosp_fraction, log10_I_
...
           : float
-            Chi square of fitting model to observed cases and deaths.
+            Chi square of fitting model to observed cases, deaths, and hospitalizations.
         """
         l = locals()
         model_kwargs = {k: l[k] for k in self.model_fit_keys}
-        model = self.run_model(**model_kwargs)
 
+        # Last data point used in Fit
+        last_data_point_used = t0 + t_break + 14 + t_delta_phases + 14
```
erccarls, line 468: I don't understand this part.. Why the second 14 here? The ramp is still fit over those 14 days and after.

pyseir/inference/model_fitter.py
```diff
@@ -418,12 +459,28 @@ def _fit_seir(self, R0, t0, eps, t_break, test_fraction, hosp_fraction, log10_I_
...
+        # Number of future days used in second ramp period
+        number_of_future_days_used = last_data_point_used - self.ref_future_date
+        # Multiplicative chi2 penalty if future_days are used in second ramp period (set to 1 by default)
+        future_days_penalty = 1.0
```
erccarls, line 472: This is very low. seems it could be at least 10x this to even be of order the size of the same size as the data chi2

pyseir/inference/model_fitter.py
```diff
@@ -418,12 +459,28 @@ def _fit_seir(self, R0, t0, eps, t_break, test_fraction, hosp_fraction, log10_I_
...
+
+        # If using more future days than allowed, updated future_days_penalty
+        if number_of_future_days_used > self.future_days_allowed:
+            future_days_penalty = number_of_future_days_used
+
+        # Only run fit when last_data_point_used does not use more than max_future_days_fitted
+        if last_data_point_used < self.ref_future_date + self.max_future_days_fitted:
```
erccarls, line 479: Why have this instead of just increasing the penalty multiplier? This still leads to a huge discontinuity that the optimizer doesn't know what to do with.

pyseir/inference/model_fitter.py
```diff
@@ -486,7 +543,11 @@ def _fit_seir(self, R0, t0, eps, t_break, test_fraction, hosp_fraction, log10_I_
         self.dof_deaths = (self.observed_new_deaths > 0).sum()
         self.dof_cases = (self.observed_new_cases > 0).sum()
 
-        return chi2_deaths + chi2_cases + chi2_hosp
+        not_penalized_score = chi2_deaths + chi2_cases + chi2_hosp
+        # Calculate the final score as the product of the future_days_penalty and not_penalized_score
+        score = future_days_penalty * (chi2_deaths + chi2_cases + chi2_hosp)
```
erccarls, line 548: Seems this should be additive not multiplicative. else you are putting a very strong and non-convex prior on not going over the fit time. This might be too aggressive to fit recent data, and doesn't have easy to understand tradeoffs against the other chi2.

pyseir/inference/model_fitter.py
erccarls: I think I understand now, but this name is very confusing.  I think this has nothing to do with the data right? It is just the end of the *model's* ramp up to the final eps value.  But this also is not the last point used in the fit since eps 2 is fit over all valid data with `t > t0 + t_break + 14 + t_delta_phases`.

pyseir/inference/model_fitter.py
tashwoods: This is the default penalty. I only change it if future days are used

pyseir/inference/model_fitter.py
tashwoods: I agree, in the past the optimizer didn't do well without this constraint, but I could test it again.

pyseir/inference/model_fitter.py
tashwoods: fixed

pyseir/inference/model_fitter.py
```diff
@@ -121,6 +137,8 @@ def __init__(
 
         self.fips = fips
         self.ref_date = ref_date
+        self.max_fit_date = (dt.date.today() - timedelta(days=7) - ref_date.date()).days  # natasha
+        self.ref_future_date = (dt.date.today() - ref_date.date()).days
```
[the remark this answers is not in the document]
tashwoods: fixed

pyseir/ensembles/ensemble_runner.py
```diff
@@ -207,9 +207,12 @@ def _load_model_for_fips(self, scenario="inferred"):
 
             # Rescale state values to the county population and replace county
             # specific params.
+            # TODO: get_average_seir_parameters should return the analytic solution when available
+            # right now it runs an average over the ensemble (with N_samples not consistently set
+            # across the code base).
             default_params = ParameterEnsembleGenerator(
                 self.fips,
-                N_samples=1,
+                N_samples=1,  # change back to 250?
```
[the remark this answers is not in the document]
tashwoods: Fixed! I wasn't sure if this should be a separate PR, but I just added it.

.github/workflows/deploy_api.yml
```diff
@@ -9,18 +9,19 @@ on:
...
+  push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
     types: publish-api
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  COVID_DATA_MODEL_REF: 'feature/third-stage-clean'
```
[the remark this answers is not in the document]
tashwoods: hahah, yep yep

pyseir/inference/model_fitter.py
```diff
@@ -89,6 +90,12 @@ class ModelFitter:
         t_break=20,
         limit_t_break=[5, 40],
         error_t_break=1,
+        eps2=0.3,
+        limit_eps2=[0.20, 1.2],
```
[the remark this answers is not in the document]
tashwoods: I'm going to update this in the next PR

pyseir/inference/model_fitter.py
```diff
@@ -89,6 +90,12 @@ class ModelFitter:
...
+        error_eps2=0.005,
+        t_delta_phases=14,  # number of days between second and third ramps
+        limit_t_delta_phases=[14, 60],
```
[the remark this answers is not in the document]
tashwoods: comment added! Agreed, we need to watch this going forward. But hopefully we have a more dynamic model before this becomes outdated : D
@BrettBoval

pyseir/inference/model_fitter.py
tashwoods: Yes, sorry I guess a better name would be:
last_data_point_ramp_2, I'll rename in the next PR

pyseir/inference/model_fitter.py
tashwoods: PR #2

pyseir/inference/model_fitter.py
tashwoods: PR #2

pyseir/ensembles/ensemble_runner.py
```diff
@@ -230,10 +233,11 @@ def _load_model_for_fips(self, scenario="inferred"):
         else:
             eps_final = sp.get_future_suppression_from_r0(inferred_params["R0"], scenario=scenario)
 
-        model.suppression_policy = sp.generate_two_step_policy(
-            self.t_list,
+        model.suppression_policy = sp.get_epsilon_interpolator(
             eps=inferred_params["eps"],
             t_break=inferred_params["t_break"],
+            eps2=inferred_params["eps2"],  # check
```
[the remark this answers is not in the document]
tashwoods: old debugging. removed

PR #475, opened 2020-06-04
Update Scenario Planner Code for Third Stage

BrettBoval:
This is a regression bug that was fixed on one of the 3rd stage test branches, but not on the one that actually got merged into master :(

### DIFF
--- a/pyseir/inference/model_fitter.py
+++ b/pyseir/inference/model_fitter.py
@@ -2,11 +2,7 @@ import logging
 import iminuit
 
 # TODO use JAX for numpy XLA acceleration
-# from jax.config import config
-# config.update("jax_enable_x64", True) # enable float64 precision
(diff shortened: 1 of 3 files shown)


pyseir/inference/model_fitter.py
```diff
@@ -2,19 +2,14 @@
 import iminuit
 
 # TODO use JAX for numpy XLA acceleration
-# from jax.config import config
-# config.update("jax_enable_x64", True) # enable float64 precision
-import numpy as np
```
jeremiq, line 7: np is used in the code below. did you mean to remove it?

BrettBoval: That was a duplicate call. It's still there. This all needs to be reordered, but that's for another PR.

PR #476, opened 2020-06-04
Adds kinsa datasource.

jeremiq:
This PR adds a datasource for kinsa data. Data is being loaded directly via the API to avoid any issues that could come from snapshotting it in covid-data-public.


tashwoods:
This looks great to me! It would be rad to add a population-weighted average (or similar method) for the state level.

PR #481, opened 2020-06-09
Feature/infer rt outlier removal

BrettBoval:
This PR addresses corrections for catching reporting changes that are applied as a one time patch to daily values. It applies only to the infer_rt codepath. It looks for values that are ~10 std devs away from the trailing 14 day average. This PR also implements structured logging of those corrections.


BrettBoval:
Hey @TomGoBravo. Please give feedback on the structured logging and enums. I've been pretty free with the scope of the logging messages. I thought it was better than the current print style logging, but I could be more rigorous about passing the logs through the class if you see a pressing need.

PR #482, opened 2020-06-09
change cmdc back to mapping CURRENT_ICU and do same in test.

TomGoBravo:
This change follows https://github.com/covid-projections/covid-data-model/pull/480 and https://github.com/covid-projections/covid-data-public/pull/57

### Please Check
- [ ] Are tests passing?


TomGoBravo:
> LGTM, is there a larger scale test that would've caught this?

Short term we had a test in `test_combined_county_has_some_timeseries_data` but the test was wrong.
combined_datasets tests did fail when ALL_FIELDS_FEATURE_DEFINITION was out of sync with ` CmdcDataSource.COMMON_FIELDS`. Medium term I'd like to eliminate COMMON_FIELDS for data sources where we convert to common fields before the model pipeline. Currently ALL_FIELDS_FEATURE_DEFINITION explicitly lists each <dataset, field> that is loaded. Do you think we can change to import all common fields from all datasets and have a single `List[dataset]` for the order of loading them?

PR #485, opened 2020-06-09
Adding nytimes to combined datasets

ghop02:
This PR addresses issue https://trello.com/c/OZ789q0g/177-fix-whitelist-code-slowness-when-rebuilding-single-state

This pull out all code that accesses nytimes datasets directly.  I need to do a bit more testing on the pyseir code, but would love a pass through!

This will change the cases/deaths to use cmdc when available as well - is that alright?

Also, it looks like we have some duplicate data in NYTimes - In MO, there is an "unknown" county + kansas city is it's own county (which probably means it's an aggregate of a couple counties). I would rather find another way to deal with that rather than just dropping out.

Also - I'm running into a test error locally, but seems to pass here - want to investigate a bit more.

### Please Check
- [ ] Are tests passing?


pyseir/inference/initial_conditions_fitter.py
```diff
@@ -12,11 +12,13 @@
...
 from functools import partial
 import seaborn as sns
+from libs.datasets import combined_datasets
+from libs.datasets import CommonFields
 from pyseir import load_data
 from pyseir import OUTPUT_DIR
 
 
-class InitialConditionsFitter:
+class InitialConditionsFitter(object):
```
TomGoBravo, line 21: I don't think this does anything https://stackoverflow.com/questions/1238606/is-it-necessary-or-useful-to-inherit-from-pythons-object-in-python-3-x

pyseir/inference/initial_conditions_fitter.py
```diff
@@ -54,20 +56,23 @@ def __init__(
...
         self.min_days_required = min_days_required
 
-        # Load case data
-        case_data = load_data.load_county_case_data()
-        self.cases = case_data[case_data["fips"] == fips]
+        timeseries = combined_datasets.get_timeseries_for_fips(fips)
+        latest_values = combined_datasets.get_us_latest_for_fips(fips)
+
+        self.cases = timeseries.data
+        # NOTE(chris): Should cases drop any empty rows? Thinking if there is a timeseries
```
TomGoBravo, line 63: How about logging warning if the cases column is all nan?

pyseir/inference/initial_conditions_fitter.py
ghop02: Ah you're right - I generally have followed the google style guid pretty strictly, but it's probably a bit outdated in a purely python 3 world

pyseir/inference/initial_conditions_fitter.py
ghop02: I wonder if it would be better to drop NaN rows.  The code below checks `if n_days < min_days_required` - maybe that n_days should represent non null days.

libs/datasets/combined_datasets.py
```diff
@@ -74,8 +77,8 @@
 }
 
 ALL_TIMESERIES_FEATURE_DEFINITION: FeatureDataSourceMap = {
-    CommonFields.CASES: [JHUDataset],
-    CommonFields.DEATHS: [JHUDataset],
+    CommonFields.CASES: [NYTimesDataset],
+    CommonFields.DEATHS: [CmdcDataSource, NYTimesDataset],
```
ghop02, line 81: Hmm looking at the current values again, I think this should just be NYTimes

pyseir/load_data.py
```diff
@@ -325,6 +331,9 @@ def load_new_case_data_by_fips(
     """
     timeseries = combined_datasets.build_us_timeseries_with_all_fields()
     county_case_data = timeseries.get_data(None, fips=fips)
+    # Only including rows with cases reported.
+    county_case_data = county_case_data.loc[county_case_data[CommonFields.CASES].notnull(), :]
```
TomGoBravo, line 335: There was already a potential bug in that the returned arrays could be missing days in the middle of the range without a way to represent the gap. The filter added here makes that problem more likely to occur.

TomGoBravo, line 335: What about county_case_data["deaths"].notnull() ?

TomGoBravo: A clean fix, that potentially is out of scope for this change is returning a Series with a `date` index.

pyseir/load_data.py
```diff
@@ -602,6 +611,8 @@ def load_new_case_data_by_state(
     state_abbrev = us.states.lookup(state).abbr
     state_timeseries = combined_datasets.get_timeseries_for_state(state_abbrev)
     state_case_data = state_timeseries.data
+    # Only including rows with cases reported.
+    state_case_data = state_case_data.loc[state_case_data[CommonFields.CASES].notnull(), :]
```
TomGoBravo, line 615: same issue as load_new_case_data_by_fips

pyseir/load_data.py
ghop02: Yeah, I more and more don't like this.  I'm trying to think of the best way to handle missing data.  another options is to fill in the missing days more explicitly. 

the issue that this is indirectly trying to fix is: 
https://sentry.io/organizations/covidactnow/issues/1734408534/?project=5203044&query=is%3Aunresolved

I think a better solution might be to just leave the nans and add an epsilon to the denominator here: https://github.com/covid-projections/covid-data-model/blob/master/pyseir/inference/infer_rt.py#L740

pyseir/cli.py
```diff
@@ -52,8 +51,6 @@ def _cache_global_datasets():
     global nyt_dataset, cds_dataset
```
BrettBoval, line 51: I think if we dig into the places that use this we will find that they are stale and can be removed. I just removed a call from ensemble. Let's land both changes as they are and check then. But I'm pretty sure these can be removed from the function calls in cli.py and we can remove them!

ghop02: oh good call!

libs/datasets/combined_datasets.py
```diff
@@ -138,6 +141,69 @@ def get_us_latest_for_fips(fips) -> dict:
...
+        return df.loc[[False] * len(df), :]
+
+    first_valid_index = min(df[column].first_valid_index() for column in columns)
+    last_valid_index = max(df[column].last_valid_index() for column in columns)
+    df = df.iloc[first_valid_index:last_valid_index]
+    return df.reset_index(drop=True)
+
+
+def get_timeseries_for_fips(
+    fips: str, columns: List = None, remove_padding_nans: bool = False
```
TomGoBravo, line 155: does columns need to have a default? refactoring later will likely be easier if callers always explicitly list the columns requested.

libs/datasets/combined_datasets.py
```diff
@@ -190,6 +256,7 @@ def build_combined_dataset_from_sources(
     # structlog makes it very easy to bind extra attributes to `log` as it is passed down the stack.
     log = structlog.get_logger()
     for field, data_source_classes in feature_definition_config.items():
+        print(field, data_source_classes)
```
TomGoBravo, line 259: rm debug output

libs/datasets/sources/nytimes_dataset.py
```diff
@@ -52,5 +52,6 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
         # Super hacky way of filling in new york.
         data.loc[data[cls.Fields.COUNTY] == "New York City", "county"] = "New York County"
         data.loc[data[cls.Fields.COUNTY] == "New York County", "fips"] = "36061"
+        data.loc[data[cls.Fields.COUNTY] == "Unknown", "fips"] = "99999"
```
TomGoBravo, line 55: What happens if you instead drop these rows?

ghop02:
> There could still be padded nans in the return value if the columns have different valid index. How about minimum_range_with_some_value?

What do you mean by different valid index?

TomGoBravo:
> > There could still be padded nans in the return value if the columns have different valid index. How about minimum_range_with_some_value?
> 
> What do you mean by different valid index?

Columns have different dates with real values. For example
```
date cases death
Ju20 12345   Nan
Ju21 12345  1234
```

libs/datasets/sources/nytimes_dataset.py
ghop02: Nothing :) I generally don't love to drop unknown data unless we have a real plan for it, but that ship has sailed a bit.  I'm fine dropping if you feel strongly about it.  it would be a good task to better understand the implications of dropping our unknown data at some point.

TomGoBravo: No need to drop the data now. I'd like to run through the csv diff tool with you and see how far we are from having confidence making changes like that.

libs/datasets/dataset_utils.py
```diff
@@ -71,27 +71,6 @@ def plot_grouped_data(data, group, series="source", values="cases"):
     cases_by_source.plot(kind="bar", figsize=(15, 7), title=f"{values} by data source vs date")
 
 
-def build_aggregate_county_data_frame(jhu_data_source, cds_data_source):
```
TomGoBravo, line 74: :-)

libs/datasets/combined_datasets.py
ghop02: hmm I guess that's true - I guess in terms of how this function is used I wouldn't expect to have to pass the columns.  I primarily see it as a wrapper over getting the entire dataset with a bit more optionality.

PR #486, opened 2020-06-10
Update Posteriors if Denominator is Zero

BrettBoval:
https://trello.com/c/KXNyFaYc/219-low-n-rt-posterior-reset-causes-red

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -383,7 +383,7 @@ class RtInferenceEngine:
             # (5b) Calculate the numerator of Bayes' Rule: P(k|R_t)P(R_t)
             numerator = likelihoods[current_day] * current_prior
 
-            # (5c) Calcluate the denominator of Bayes' Rule P(k)
+            # (5c) Calculate the denominator of Bayes' Rule P(k)
             denominator = np.sum(numerator)
 
             # Execute full Bayes' Rule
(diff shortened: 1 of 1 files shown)


BrettBoval:
Hey @jeremiq:

The reseting posterior to the initial prior is breaking county colors because the prior is so high. Can we either do this fix (just keep passing forward the previous day's posterior) or can we have a second "reset prior" that is centered around 1?

pyseir/inference/infer_rt.py
```diff
@@ -394,13 +397,13 @@ def get_posteriors(self, timeseries_type, plot=False):
...
                 # We understand that restarting the posteriors with the
-                # initial prior may incur a start-up artifact as the posterior
+                # re-initial prior may incur a start-up artifact as the posterior
                 # restabilizes, but we believe it's the current best
                 # solution for municipalities that have smoothed cases and
                 # deaths that dip down to zero, but then start to increase
                 # again.
 
-                posteriors[current_day] = prior0
+                posteriors[current_day] = reinit_prior
```
jeremiq, line 406: This looks good to me. Another option would be, like you say, to just carry forward the last valid day, e.g. ```posteriors[current_day] = posteriors[previous_day]```
I like this solution more, though, as it gives more flexibility for adjusting the re-initialization condition going forward.

PR #490, opened 2020-06-11
Refactor/remove convert cumulatives to current

BrettBoval:
This PR removes a feature that originally was put in to estimate prevalence from cumulatives for acute and icu hospitalizations with the goal of scaling model output to latest observed. Now that more agencies are reporting current, we no longer want to support this conversion. In a few remaining locations, we will no longer scale outputs if we don't have current. This allows us to support one less model with obscure assumptions and effects.

There is also some streamlining of functions around this business logic.

### DIFF
--- a/pyseir/load_data.py
+++ b/pyseir/load_data.py
@@ -563,7 +563,7 @@ def get_current_hospitalized(fips, t0, category: HospitalizationCategory):
 
     if len(fips) == 2:
         kwargs = dict(
-            aggregation_level=AggregationLevel.State,
+            aggregation_level=AggregationLevel.STATE,
             country="USA",
             state=us.states.lookup(fips).abbr,
         )
--- a/pyseir/load_data.py
+++ b/pyseir/load_data.py
@@ -596,10 +596,7 @@ def get_current_hospitalized_for_county(fips: str, t0: datetime, category: Hospi
 
 
 def _get_current_hospitalized(
-    df: pd.DataFrame,
-    t0: datetime,
-    category: HospitalizationCategory,
-    # min_cumulative_datapoints_to_convert: int = MIN_CUMULATIVE_DATAPOINTS_TO_CONVERT,
+    df: pd.DataFrame, t0: datetime, category: HospitalizationCategory,
 ):
     """
     Given a DataFrame that contains values icu or hospitalization data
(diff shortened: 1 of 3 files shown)


pyseir/load_data.py
```diff
@@ -399,11 +398,16 @@ def load_new_case_data_by_fips(
 
 
 def get_hospitalization_data():
+    """
+    Since we're using this data for hospitalized data only, only returning
+    values with hospitalization data.  I think as the use cases of this data source
+    expand, we may not want to drop. For context, as of 4/8 607/1821 rows contained
```
TomGoBravo, line 404: Did you mean: as of June 11 the timeseries per region has hospitalization data for April 8? Or the latest data as of June 8?

TomGoBravo: Oh, this isn't a new comment :-)

pyseir/load_data.py
```diff
@@ -553,183 +558,69 @@ def get_current_hospitalized_for_state(state: str, t0: datetime, category: Hospi
...
-    df = get_hospitalization_data().get_data(AggregationLevel.COUNTY, country="USA", fips=fips)
+    if len(fips) == 2:
+        kwargs = dict(
+            aggregation_level=AggregationLevel.STATE,
+            country="USA",
+            state=us.states.lookup(fips).abbr,
+        )
+    else:
+        kwargs = dict(aggregation_level=AggregationLevel.COUNTY, country="USA", fips=fips)
+    df = combined_datasets.build_us_timeseries_with_all_fields().get_data(**kwargs)
```
TomGoBravo, line 572: kwargs make reasoning a little harder. instead
ts = combined_datasets.build_us_timeseries_with_all_fields()
```
if len(fips) == 2:
  df = ts.get_data(AggregationLevel.STATE, country="USA", state=us.states.lookup(fips).abbr)
else:
  df = ts.get_data(AggregationLevel.COUNTY, country="USA", fips=fips)
```

pyseir/load_data.py
```diff
@@ -553,183 +558,69 @@ def get_current_hospitalized_for_state(state: str, t0: datetime, category: Hospi
...
-    initial_daily_estimate = daily_admits[:max_window].mean()
+    NUM_DAYS_TO_LOOK_BACK = 3
+    # Agencies will start and stop reporting values. Also, depending on the time of day some columns
+    # in a day row may propagate before others. Therefore, we don't want to just take the most
+    # recent value which may be None, nor take just the most recent any value, which may be weeks
+    # ago.
 
-    # And use that steady state solution as if that data had been entered in the past to initialize.
-    t0_patients = initial_daily_estimate * average_length_of_stay
+    # Datetimes are in UTC. Look at possible values that are within this time window.
```
TomGoBravo, line 608: Oh, I thought they are naive (without TZ) so you'd want to compare them to a naive time to avoid doing a bunch of timezone aware comparisons. How about avoiding naive vs not by using dates? `datetime.date.today() - datetime.timedelta(days=3)`?

BrettBoval: Agree on simplifying dates. The combined dataset has UTC naive datetimes with the times set to 0s. So I need to roughly get the UTC date right now (which is often a "day" ahead of pacific time). That's why I call utcnow and then extract the date.

PR #492, opened 2020-06-13
Feature/make seir simple ish

tashwoods:
This PR sets initial guesses for SEIR parameters for all states based on good fit values from June 10, 2020. Counties inherit some initial SEIR parameters from states, but there is still work to be done to improve county-level fits. Setting initial parameters improves fit stability.

### Please Check
- [ Oui ] Are tests passing?


pyseir/inference/model_fitter.py
```diff
@@ -210,9 +204,23 @@ def set_inference_parameters(self):
         """
         self.fit_params = self.DEFAULT_FIT_PARAMS
         # Update any state specific params.
-        for k, v in self.PARAM_SETS.items():
-            if self.state_obj.abbr in k:
-                self.fit_params.update(v)
+        initial_params_df = pd.read_csv("./pyseir/inference/world_according_to_pyseir.csv")
```
BrettBoval, line 207: I know I initially named this, but can we make it more meaningful. It should also be in the pyseir_data directory. So something like "pyseir_fitter_initial_conditions_2020_06_10.csv"

pyseir/inference/model_fitter.py
```diff
@@ -210,9 +204,23 @@ def set_inference_parameters(self):
...
+            "t0",
+            "eps",
+            "t_break",
+            "eps2",
+            "t_delta_phases",
+            "test_fraction",
+            "hosp_fraction",
+            "log10_I_initial",
+        ]
+        if int(self.fips) in initial_params_df["fips"]:
```
BrettBoval, line 220: This call to int feels wrong. We handle fips everywhere else as strings, so in the pd.read_csv you should call that column as a string (zfill-ed correctly as needed).

BrettBoval: As in the fips for alabama is "01" not just "1".

pyseir/inference/model_fitter.py
BrettBoval: And call the df something semantic like 'overwrite_params_df'?

pyseir/inference/model_fitter.py
```diff
@@ -210,9 +204,23 @@ def set_inference_parameters(self):
...
+            this_fips_df = initial_params_df.loc[initial_params_df["fips"] == int(self.fips)]
+            for param in INITIAL_PARAM_SETS:
```
BrettBoval, line 222: This is fine, but why did you change it from the original? `self.fit_params.update(v)` is pretty clear, especially if you make the key the str fips index.

pyseir/inference/model_fitter.py
```diff
@@ -624,7 +637,7 @@ def fit(self):
                 eps_error=self.fit_results[f"{epsilon}_error"],
                 lower_bound_reff=ModelFitter.REFF_LOWER_BOUND,
             )
-            # TODO: Add structured logging if this change is significant
+            log.info(f"epsilon: {self.fit_results[epsilon]} adjusted: {adjusted_epsilon}")
```
BrettBoval, line 640: Think of the struct log as an object that you can add key-value pairs to like : `log.info("Adjusted Epsilon", original=self.fit_results[epsilon], adjusted=adjusted_epsilon)`

BrettBoval: And throw fips in there if it's in scope.

pyseir/inference/model_fitter.py
tashwoods: done

pyseir/inference/model_fitter.py
tashwoods: No huge reason, and changing now would require refactoring and shouldn't be blocking. Feel free to change this later if you want

pyseir/inference/model_fitter.py
tashwoods: Ok, also feel free to change this if you care, we are about to remove this adjustment anyway.

pyseir/inference/model_fitter.py
tashwoods: done

PR #494, opened 2020-06-15
Disable post-fit eps posterior updating

tashwoods:
This PR removes the epsilon posterior updating after the fit that was previously applied. Now we trust the fit to find the best epsilon fit values. 

### Please Check
- [ running] Are tests passing?

### DIFF
--- a/pyseir/inference/model_fitter.py
+++ b/pyseir/inference/model_fitter.py
@@ -123,7 +123,7 @@ class ModelFitter:
 
         self.fips = fips
         self.ref_date = ref_date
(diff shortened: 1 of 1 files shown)


pyseir/inference/model_fitter.py
```diff
@@ -631,6 +631,8 @@ def fit(self):
                 f"Epsilon == 0 which implies lack of convergence."
             )
 
+        # For now we are not applying posterior updating after the fit, but trust the fit to find the best R values
+        """
```
BrettBoval, line 635: Comment out each line with #. That's for docstrings and makes all these compare tools hard to use.

PR #499, opened 2020-06-18
Add common_df_diff and utils csv-diff

TomGoBravo:
## This PR

I'd like to get this merged so I can stop dragging around the branch. It isn't used in the data build and needs quite a bit of work to be generally useful so don't think it needs a detailed review at this time.

* Adds a hacky little library `libs/qa/common_df_diff.py` that compares two `DataFrame` marked as unreviewed
* Adds `python ./run.py utils csv-diff output/combined-master-daf6ef3-20200618T073552.csv output/combined-tgb-20200618-combine-refactor-no-drop-na-855fb31-20200618T091315.csv`

### DIFF
--- a/cli/utils.py
+++ b/cli/utils.py
@@ -2,13 +2,11 @@ import logging
 import pathlib
 import subprocess
 from datetime import datetime
-import pandas as pd
-import numpy as np
 
 import click
 import structlog
 
-from covidactnow.datapublic.common_df import write_df_as_csv, sort_common_field_columns
+from covidactnow.datapublic import common_df
 from libs import github_utils
 from libs.datasets import combined_datasets
 
@@ -62,4 +60,4 @@ def save_combined_csv(csv_path_format, output_dir):
     timeseries = combined_datasets.build_us_timeseries_with_all_fields()
     timeseries_data = timeseries.data
 
-    write_df_as_csv(timeseries_data, csv_path, structlog.get_logger())
+    common_df.write_csv(timeseries_data, csv_path, structlog.get_logger())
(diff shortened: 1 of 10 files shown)


libs/qa/common_df_diff.py
```diff
@@ -0,0 +1,138 @@
...
+
+        all_variable_fips = melt.groupby("variable fips".split()).first().index
+        return DatasetDiff(duplicates_dropped=dups, melt=melt, all_variable_fips=all_variable_fips)
+
+    def compare(self, other: "DatasetDiff"):
+        # A timeseries is a set of <date, value> tuples identified by a <variable, fips> tuple. First
+        # find all the timeseries that have at least one real value in only one dataset. These are
+        # timeseries that appear in one file but not the other file.
+        self.my_ts = self.all_variable_fips.difference(other.all_variable_fips)
+        other.my_ts = other.all_variable_fips.difference(self.all_variable_fips)
```
ghop02, line 79: I'm having a bit of trouble with `my_ts`. I think I expect "ts" to be a pandas Series or DataFrame, so it's throwing me that it's a multi index? I don't think I quite understand this comment: "A timeseries is a set of <date, value> tuples identified by a <variable, fips> tuple"

cli/utils.py
```diff
@@ -88,3 +89,21 @@ def form_path_name(csv_path_format, output_dir):
...
     )
     return csv_path
+
+
+@main.command()
+@click.argument("csv_path_left", type=str, required=True)
+@click.argument("csv_path_right", type=str, required=True)
+def csv_diff(csv_path_left, csv_path_right):
+    """Compare 2 CSV files."""
+    df_l = common_df.read_csv(csv_path_left)
```
ghop02, line 99: It would be great if there was a quick expectation of what the csv's should contain (in terms of header).  I could see that either being a comment in the docstring or an exception that's thrown with instructions as to how to construct the csv.

TomGoBravo:
I added a bunch more comments but haven't changed the code much. PTAL @ghop02

PR #504, opened 2020-06-23
Test/snapshot timeout

BrettBoval:
This PR addresses issue https://trello.com/c/m7jfBoUo/254-covid-data-model-snapshots-timing-out.

The snapshot builds started to slow down over the week of 14 June 2020 and by Saturday evening the build had timed out (6 hours). Watching the build machine during a build (see htop screenshots in trello) shows a large ramp up of reserved RAM during the ensemble_runner code to the point that all RAM is assigned. There is no paging, so the machine essentially just stalls until a 6 hour watchdog timer in github kicks us out.

During the initial troubleshooting, I wrapped all the map calls in the appropriate context loops to ensure they were closed. Then I limited the nested pool calls to limit the children processes of the inner loop. Finally, I realized that each ensemble_runner.EnsembleRunner instance was load the JHU dataset and then not using it. The hypothesis is that our profligate loading didn't become a problem until the JHU dataset became large enough that having 100s of copies finally ran into our memory limits.

There are no usages of the class outside of that module. By removing that call in the init, the memory overload has been fixed and we've removed another path that accesses individual datasets instead of combined_dataset.

I removed the limits of the pools (maxtasksperchild), although these didn't significantly slow down runtime. It's clear we aren't memory-bound in normal usage, but the 186GB comes with the cores.

It's also worth nothing that the first couple minutes of execute-model, all of execute-api and all of push-snapshot-to-s3 are sequential and using a single core at the moment. I've talked with Simon and agree that removing WebUI while combining it with execute-api is the lowest hanging fruit going forward for simplicity, readability, and speed.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -17,10 +17,10 @@ on:
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  COVID_DATA_MODEL_REF: 'test/snapshot-timeout'
 
   # To pin to an old data sets, put the branch/tag/commit here:
-  COVID_DATA_PUBLIC_REF: '1d5ea08606c6d0c5e9ad29493a19baa6d5279a39'
+  COVID_DATA_PUBLIC_REF: 'master'
 
   # S3 Bucket (used by s3-sync-action tasks) to store final API snapshot.
   AWS_S3_BUCKET: 'data.covidactnow.org'
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -20,7 +20,7 @@ env:
   COVID_DATA_MODEL_REF: 'test/snapshot-timeout'
 
   # To pin to an old data sets, put the branch/tag/commit here:
-  COVID_DATA_PUBLIC_REF: 'master'
+  COVID_DATA_PUBLIC_REF: '6351ef4b2533d99f08625663fc729cf98f122da0'
 
   # S3 Bucket (used by s3-sync-action tasks) to store final API snapshot.
   AWS_S3_BUCKET: 'data.covidactnow.org'
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -9,7 +9,7 @@ on:
   schedule:
    - cron: '30 0,12 * * *'
 
-  # push:
+  push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
@@ -17,7 +17,7 @@ on:
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  COVID_DATA_MODEL_REF: 'test/snapshot-timeout'
 
   # To pin to an old data sets, put the branch/tag/commit here:
   COVID_DATA_PUBLIC_REF: 'master'
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -9,7 +9,7 @@ on:
   schedule:
    - cron: '30 0,12 * * *'
 
-  push:
+  # push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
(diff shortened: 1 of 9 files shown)


pyseir/cli.py
```diff
@@ -114,10 +116,12 @@ def _run_mle_fits(state=None, states_only=False):
 
 def _run_ensembles(state=None, ensemble_kwargs=dict(), states_only=False):
     if state:
-        run_state(state, ensemble_kwargs=ensemble_kwargs, states_only=states_only)
+        ensemble_runner.run_state(state, ensemble_kwargs=ensemble_kwargs, states_only=states_only)
```
ghop02, line 119: 🙏

BrettBoval:
Yeah. All these changes successfully built last night, the morning snapshots are to give standup two options to use. Nevertheless I'll wait until they complete before merging this.

BrettBoval:
@ghop02 @TomGoBravo 

What do you think about adding the maxtasksperchild back into the Pool calls. It doesn't noticeably slow down the build (it was within run-to-run variation when I was debugging) and it unlocks significant resources. Even with breaking out each of these pools in their own context, we still accumulate ~50GB memory usage. When I use maxtasksperchild=1 that doesn't get above 10GB (I would guess averaging 5GB). I'm somewhat in favor of it because it makes it clearer to evaluating what the actual memory usage is for the program and can guide decision making if we don't want to use such an expensive build machine in the future.

Here's the note from the multiprocessing docs:
> Note Worker processes within a Pool typically live for the complete duration of the Pool’s work queue. A frequent pattern found in other systems (such as Apache, mod_wsgi, etc) to free resources held by workers is to allow a worker within a pool to complete only a set amount of work before being exiting, being cleaned up and a new process spawned to replace the old one. The maxtasksperchild argument to the Pool exposes this ability to the end user.

ghop02:
Just want to make sure I'm understanding correctly - maxtasksperchild will essentially limit the number of times a pool worker is used. 

I think this sounds good! One thing we need to watch out for is that we're not initializing a cache inside the worker (which we shouldn't be anyway) or else it will be very slow.  I.e. we used to basically load combined datasets in each worker, meaning it would load the first time and then not need to load again for the life of the worker.  That was naive on my part, so setting maxtasksperchild to 1 would've made that misplacement much more painful. 

In short, I'm in favor of pushing as much state out of the workers as possible, so setting maxtasksperchild to 1 seems like a great step in that direction.

.github/workflows/deploy_api.yml
```diff
@@ -9,15 +9,15 @@ on:
...
+  push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
     types: publish-api
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  COVID_DATA_MODEL_REF: 'test/snapshot-timeout'
```
TomGoBravo, line 20: don't forget to revert this

pyseir/cli.py
```diff
@@ -11,7 +11,9 @@
 from pyseir.load_data import cache_county_case_data
 from pyseir.inference.initial_conditions_fitter import generate_start_times_for_state
 from pyseir.inference import infer_rt as infer_rt_module
-from pyseir.ensembles.ensemble_runner import run_state, RunMode, _run_county
+from pyseir.ensembles import ensemble_runner
+
+# from pyseir.ensembles.ensemble_runner import run_state, RunMode, _run_county
```
TomGoBravo, line 16: rm line

pyseir/inference/model_fitter.py
```diff
@@ -1081,7 +1081,7 @@ def run_state(state, states_only=False, with_age_structure=False):
         ].fips.values
 
         if len(all_fips) > 0:
-            with Pool() as p:
+            with Pool(maxtasksperchild=1) as p:
```
ghop02, line 1084: Definitely no need to do this now, but it'd be nice to have this as a common function so that we're getting all pools initialized with the same parameters. 

Could make a quick wrapping context manager function something like this: 
```
@contextlib.contextmanager
def get_pool(*args, **kwargs):
    with Pool(*args, maxtasksperchild=1, **kwargs) as p:
        yield p
```
would probably work?

TomGoBravo: A common context manager holding our preferred kwargs makes sense. An alternative is reusing the same pool, passing around `p` as a function parameter where it is needed.

PR #506, opened 2020-06-24
Apply Shim Method Consistently Across Total_Hospitalized, ICU, and Deaths

BrettBoval:
This PR addresses issue https://trello.com/c/aXp44GbY/233-fix-projection-scaling.

The method that was applied to cumulative deaths is extended to also apply to ICU and Total Hospitalized.

### Please Check
- [ ] Are tests passing?


pyseir/deployment/model_to_observed_shim.py
```diff
@@ -1,54 +1,106 @@
...
-    if observed_latest_cum_deaths is None:
-        observed_latest_cum_deaths = np.nan
+    if observed is None:
+        observed = np.nan
 
-    model_latest_cum_deaths = model_death_ts[idx]
+    if np.isnan(observed):
+        shim = 0
+    elif observed == 0:
+        # As of 19 June 2020, the observed dataset is still being validated for erroneous inputs.
```
BrettBoval, line 31: @ghop02 We should QA these cases and see if this is still true for build latest. I would like to remove these special cases.

ghop02: When you say "observed dataset" which dataset is that referring to?

BrettBoval: I use "observed" to mean ground truth data, which in this case is whatever is returned from latest combined_dataset.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -207,26 +110,52 @@ def map_fips(self, fips: str) -> None:
...
+        # Get the latest observed values to use in calculating shims
         observed_latest_dict = combined_datasets.get_us_latest_for_fips(fips)
         shim_log = structlog.getLogger(fips=fips)
 
-        death_shim = shim.shim_deaths_model_to_observations(
-            model_death_ts=model_death_ts,
-            idx=idx_offset,
-            observed_latest=observed_latest_dict,
-            log=shim_log,
+        # For Deaths
```
ghop02, line 123: Do you think it would be helpful at all to pull this code block here out into a function that's maybe like?
``` death_shim, total_host_shim, icu_shim = calculate_shims(...)```

I think it would give us a chance to maybe make the shim calculation more testable + easier to rip out if we want

File: pyseir/deployment/model_to_observed_shim.py [hunk 3]
```diff
@@ -1,54 +1,106 @@
...
+        shim = 0
+    else:
+        shim = observed - model
+
+    log.info(event="strict_shim", shim=np.round(shim), observed=observed, model=np.round(model))
+    return shim
+
+
+def intralevel_icu_shim(
+    model_acute_latest: float,
```
ghop02, line 43: nbd but having latest on all the variable's seems a bit unnecessary here - there's no real way we verify that these are latest, so would prefer to have this function be agnostic to that fact

.github/workflows/deploy_api.yml
```diff
@@ -9,15 +9,15 @@ on:
...
+  push:
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
     types: publish-api
 
 env:
   # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  COVID_DATA_MODEL_REF: 'shim-total-hospitalized'
```
mikelehen, line 20: Don't forget to undo these. :)

pyseir/deployment/model_to_observed_shim.py
```diff
@@ -1,54 +1,106 @@
-from typing import Sequence
-
 import numpy as np
 
 
-def shim_deaths_model_to_observations(
-    model_death_ts: Sequence, idx: int, observed_latest: dict, log
-):
+def strict_shim(model: float, observed: float, log) -> float:
```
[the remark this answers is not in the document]
BrettBoval: Done

pyseir/deployment/model_to_observed_shim.py
BrettBoval: Done. I wanted to reinforce that this is a value not an array, but having type annotations and docstring point this out is probably enough.

pyseir/deployment/model_to_observed_shim.py
```diff
@@ -1,54 +1,106 @@
...
     """
     Take model outputs and calculate a "shim" value that can be added to the entire modeled
     cumulative deaths series to make them match the latest (i.e. today's) actual cumulative deaths.
```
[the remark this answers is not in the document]
BrettBoval: Done.

pyseir/deployment/webui_data_adaptor_v1.py
BrettBoval: Fair point, but I think I'll leave it as for now. I'd like to continue to treat the three shims independently for now, since they serve three different parts of the front end. But testing them as a group on the same data object makes sense.

PR #509, opened 2020-06-25
Make global constant uppercase

ghop02:
Has been bugging me for a bit, just makes casing more consistent


TomGoBravo:
Looks like the unit test failure was the nytimes / jhu data issue we've fixed.

PR #513, opened 2020-06-26
Adding texas hospitalization override

ghop02:
https://github.com/covid-projections/covid-data-public/pull/71 needs to be merged before this will pass, but adds the texas hospitalization override data source


libs/datasets/sources/texas_hospitalizations.py
```diff
@@ -0,0 +1,33 @@
...
+
+    COMMON_FIELD_MAP = {
+        CommonFields.CURRENT_HOSPITALIZED: CommonFields.CURRENT_HOSPITALIZED,
+    }
+
+    @classmethod
+    def local(cls):
+        data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
+        input_path = data_root / cls.DATA_PATH
+        data = pd.read_csv(
```
TomGoBravo, line 30: `common_df.read_csv(....)`

PR #514, opened 2020-06-26
GitHub Self-Hosted Runner Queue

BrettBoval:
Enforce a queue to the GitHub self-hosted runner that treats the creation of the entire snapshot as the job.

The previous version separated the workflow into multiple "jobs" that were linearly dependent on each other. That "jobs" abstraction is built to allow parallel processes, and if we used the GitHub hosted version it would run up to 20 jobs in parallel. However, the runner built for self-hosting prioritizes the jobs across snapshots in a way that delays the time to first snapshot completion (without changing the time to last snapshot completion). Instead of completing all the jobs of the first snapshot, it will cycle through each outstanding snapshot request calling the next available job for that snapshot.

It is better to have the early snapshot complete and unblock the person waiting for it.

This is solved by switching a given snapshot to a single job with multiple steps. The "steps" abstraction is the correct mental model that enforces sequential ordering (which we want anyways in our current implementations). So this will send all the commands to the runner as one packet.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -44,15 +44,17 @@ env:
 
 
 jobs:
-  execute-model:
+  build-entire-job:
     runs-on: self-hosted
     steps:
+
     - name: Checkout covid-data-model
       uses: actions/checkout@v2
       with:
         repository: covid-projections/covid-data-model
         path: covid-data-model
         ref: '${{env.COVID_DATA_MODEL_REF}}'
+
     - name: Checkout covid-data-public
       uses: actions/checkout@v2
       with:
@@ -60,11 +62,13 @@ jobs:
         path: covid-data-public
         lfs: true
         ref: '${{env.COVID_DATA_PUBLIC_REF}}'
+
     - name: Setup Python
       uses: actions/setup-python@v1
       with:
         python-version: '3.7.6'
         architecture: 'x64'
+
     - name: Cache Pip
       uses: actions/cache@v1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_api.yml
```diff
@@ -165,12 +119,6 @@ jobs:
       run: |
         rm -rf ./tmp/data/
 
-  finish:
-    runs-on: self-hosted
-    needs:
-      - push-snapshot-to-s3
-
-    steps:
     - name: Maybe trigger covid-projections repo to use new snapshot.
```
BrettBoval, line 122: @mikelehen Here is a good opportunity to improve the naming of this step. I'd like to know what is contingent that causes you to call it "maybe". Could you call it "Propose that covid-projections repo use new snapshot"?

.github/workflows/deploy_api.yml
```diff
@@ -9,15 +9,15 @@ on:
...
   # Hook to trigger a manual run.
   # See: https://goobar.io/2019/12/07/manually-trigger-a-github-actions-workflow/
   repository_dispatch:
     types: publish-api
 
 env:
-  # !!! Change this to your BRANCH if you want to test it
-  COVID_DATA_MODEL_REF: 'master'
+  # !!! Change this to your BRANCH if you want to test it !!!
+  COVID_DATA_MODEL_REF: 'github-runner-queue'
```
mikelehen, line 20: Don't forget to undo these. :-)

.github/workflows/deploy_api.yml
```diff
@@ -44,27 +44,31 @@ env:
 
 
 jobs:
-  execute-model:
+  build-entire-job:
```
mikelehen, line 47: build-and-publish-snapshot ?

.github/workflows/deploy_api.yml
mikelehen: How about: 

> Trigger website PR generation for new snapshot if appropriate (master branch, morning build, etc.)

PR #517, opened 2020-06-27
moving back api deploys

ghop02:


.github/workflows/deploy_api.yml
```diff
@@ -7,7 +7,7 @@ on:
   # covid-data-public fetches data at 00:00 and 12:00 UTC.
   # So we rebuild / publish the API at 00:30 and 12:30 UTC.
   schedule:
-   - cron: '30 0,12 * * *'
+   - cron: '30 1,16 * * *'
```
mikelehen, line 10: Can you update the comment to match?  (or else remove / reword it so that it's not as fragile in the face of these tweaks)

ghop02: will do!

PR #518, opened 2020-06-27
Fix key error exception handling

ghop02:
This PR addresses a handful of small pyseir bugs/warnings:
 * Adds an option to `pyseir build-all` to run on a single fips. This further filters the counties that it runs on.
 * Fixes (this)[https://sentry.io/organizations/covidactnow/issues/1748441676/] sentry error where infer rt was failing if an outlier was detected on the last day
 * Fixes (this)[https://sentry.io/organizations/covidactnow/issues/1736478765/] sentry error where if there were no timeseries for a county it would error out.
 * Fixes noisy logs in webui data adapter if a county didn't have a valid rt inference - before it was trying to load a non existing file, now it checks if path exists before reading into pandas data frame.
 * Removed hampel filtering in load data (duplicating Natasha's change)

Some of the larger diffs are just from doing a couple earlier returns.

### DIFF
--- a/libs/datasets/sources/nytimes_dataset.py
+++ b/libs/datasets/sources/nytimes_dataset.py
@@ -47,11 +47,31 @@ class NYTimesDataset(data_source.DataSource):
     @classmethod
     def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
         data[cls.Fields.COUNTRY] = "USA"
+        data[cls.Fields.AGGREGATE_LEVEL] = "county"
         data = dataset_utils.strip_whitespace(data)
         data[cls.Fields.STATE] = data[cls.Fields.STATE].apply(dataset_utils.parse_state)
         # Super hacky way of filling in new york.
         data.loc[data[cls.Fields.COUNTY] == "New York City", "county"] = "New York County"
         data.loc[data[cls.Fields.COUNTY] == "New York County", "fips"] = "36061"
         data.loc[data[cls.Fields.COUNTY] == "Unknown", "fips"] = "99999"
-        data[cls.Fields.AGGREGATE_LEVEL] = "county"
-        return data
+
+        # https://github.com/nytimes/covid-19-data/blob/master/README.md#geographic-exceptions
+        # Both Joplin and Kansas City numbers are reported separately from the surrounding counties.
+        # Until we figure out a better way to spread amongst counties they are in, combining all
+        # data missing a fips into one unknown fips.
+        is_kc = data[cls.Fields.COUNTY] == "Kansas City"
+        is_joplin = data[cls.Fields.COUNTY] == "Joplin"
+        data.loc[is_kc | is_joplin, cls.Fields.FIPS] = "99999"
+        is_missouri = data[cls.Fields.STATE] == "MO"
+        is_unknown = data[cls.Fields.FIPS] == "99999"
+        missouri_unknown = data.loc[is_missouri & is_unknown, :]
+        group_columns = [
+            cls.Fields.AGGREGATE_LEVEL,
+            cls.Fields.FIPS,
+            cls.Fields.DATE,
+            cls.Fields.COUNTRY,
+            cls.Fields.STATE,
+        ]
+        missouri_unknown = missouri_unknown.groupby(group_columns).sum().reset_index()
+        missouri_unknown[cls.Fields.COUNTY] = "Aggregated City and Unknown Data"
+        return pd.concat([data.loc[~(is_missouri & is_unknown), :], missouri_unknown])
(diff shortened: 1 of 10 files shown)


libs/datasets/sources/nytimes_dataset.py
```diff
@@ -47,11 +47,31 @@ def local(cls) -> "NYTimesDataset":
...
         # Super hacky way of filling in new york.
         data.loc[data[cls.Fields.COUNTY] == "New York City", "county"] = "New York County"
         data.loc[data[cls.Fields.COUNTY] == "New York County", "fips"] = "36061"
         data.loc[data[cls.Fields.COUNTY] == "Unknown", "fips"] = "99999"
-        data[cls.Fields.AGGREGATE_LEVEL] = "county"
-        return data
+
+        # https://github.com/nytimes/covid-19-data/blob/master/README.md#geographic-exceptions
+        # Both Joplin and Kansas City numbers are reported separately from the surrounding counties.
+        # Until we figure out a better way to spread amongst counties they are in, combining all
```
BrettBoval, line 60: If we don't think NYT is going to ever fix this, should we just apportion based on relative population of the counties and call it good enough. I don't see an obviously better solution that doesn't get complicated or brittle quickly.

pyseir/cli.py
```diff
@@ -376,6 +393,7 @@ def map_outputs(state, output_interval_days, run_mode, states_only):
 @click.option(
     "--skip-whitelist", default=False, is_flag=True, type=bool, help="Skip the whitelist phase.",
 )
+@click.option("--fips", help="County level fips code to restrict runs to.")
```
BrettBoval, line 396: Maybe worth noting (1) that those fips can be across any state and (2) --states-only flag will override this and (3) I think you still have to build all for states with the current flow. So passing a single fips will still run for all states (and then just one county). Passing the -s flag will limit to one state and one county (not enforced to be the same) but if you had that in the doc that seems reasonable.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -312,8 +312,8 @@ def map_fips(self, fips: str) -> None:
             output_model = output_model.fillna(0)
 
             # Fill in results for the Rt indicator.
-            try:
-                rt_results = load_Rt_result(fips)
+            rt_results = load_Rt_result(fips)
+            if rt_results is not None:
```
BrettBoval, line 316: Does this change catch both the ValueError and the KeyError that was being handled previously? I'm curious if that try-except was handling two distinct pieces of logic and now we might have dropped one of them.

BrettBoval: But in general this is a much more readable way to capture what we were catching. Thank you!

pyseir/cli.py
ghop02: Yeah that's a good point.  There's not a ton of guidance around how to actually use it.  I'll add that to the help string, thanks!

pyseir/deployment/webui_data_adaptor_v1.py
ghop02: That's a good question.  I think I didn't see any evidence of where a value error was actually being raised (and it seems like the latest build didn't throw it).  I would lean towards waiting for it to surface again

libs/datasets/sources/nytimes_dataset.py
ghop02: Yeah I went back and forth on what the best way to handle that is, but ultimately I think it might worth a trello task to dig a bit more into it.  

I do agree that population seems reasonable but I'm not sure at what point that begins to break down - also for a while they counted some of these in the Unknown county, which is no longer around. 

I'm not totally opposed to fixing it here though if you want

pyseir/cli.py
ghop02: I added to this comment, let me know if that's more clear or there's a better way to phrase it

pyseir/inference/infer_rt.py
```diff
@@ -965,7 +968,7 @@ def replace_outliers(
             slicer = slice(idx - local_lookback_window, idx + local_lookback_window)
             changed_snippets.append(x[slicer].astype(int).tolist())
             try:
-                x[idx] = np.mean([x[idx - 1], x[idx + 1]])
+                x[idx] = np.mean([x.iloc[idx - 1], x.iloc[idx + 1]])
```
BrettBoval, line 971: I think this way of fixing the problem makes the most sense. It was confusing numpy and pandas indexing, so I just made it clearer that we are using numpy based indexing, which will throw the correct error. The switch to flatnonzero is recommended in the docs when people are using where, but not changing the value immediately inline. Future work to simplify and use np.where correctly, but it's not a high priority and this works (without the magic call to [0] at the end either).

ghop02: Ah yeah that makes sense!

PR #519, opened 2020-06-28
Dont return rt result if Rt composite doesnt exist

ghop02:
This fixes [this](https://sentry.io/organizations/covidactnow/issues/1697483066/?project=5203044&query=is%3Aunresolved) sentry error.  

I believe it was a result of the latest Rt smoothing - the smoothing expects Rt composite to be there but that only exists when there is some case/death data.  

This fix isn't the best fix, but was what I felt most comfortable with.  I'm pretty sure this happens in low case count counties, so shouldn't be too impactful on a global level

I tested out fips 31081 in nebraska, and case/death data was filtered out, but the smoothed hospitalization data wasn't.  

One fix I would be in favor of doing is changing how we include hosps as an "available timeseries" [here](https://github.com/covid-projections/covid-data-model/blob/master/pyseir/inference/infer_rt.py#L614-L626).  we have a condition `if len(hosps > 3)` but it might be better to define a minimum hospitalization threshold.  

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -716,6 +716,13 @@ class RtInferenceEngine:
             df_all["Rt_MAP_composite"] = df_all["Rt_MAP__new_cases"]
             df_all["Rt_ci95_composite"] = df_all["Rt_ci95__new_cases"]
 
+        if "Rt_MAP_composite" not in df_all.columns:
+            # Lacking a composite Rt indicates that there was not case or death data for fips,
+            # but there may be other timeseries (such as hospitalizations).
+            # Returning early here assumes we don't want to just report only hospitalization Rt.
+            log.warning("Rt composite not found, skipping", fips=self.fips)
+            return None
+
         # Optionally Smooth just Rt_MAP_composite.
         # Note this doesn't lag in time and preserves integral of Rteff over time
         for i in range(0, InferRtConstants.SMOOTH_RT_MAP_COMPOSITE):


BrettBoval:
I can't think of an example where we hospitalization data but no case or death data. This makes me nervous that this error is actually catching something else. Are you saying that we drop case/death data in such a way that it seems empty, but we don't dump the associated hospitalization data?

ghop02:
> I can't think of an example where we hospitalization data but no case or death data. This makes me nervous that this error is actually catching something else. Are you saying that we drop case/death data in such a way that it seems empty, but we don't dump the associated hospitalization data?

yeah so here's one spot that could happen:
```python
        if np.sum(cases) > self.min_cases:
            available_timeseries.append(TimeseriesType.NEW_CASES)
        if np.sum(deaths) > self.min_deaths:
            available_timeseries.append(TimeseriesType.NEW_DEATHS)
        if (
            self.hospitalization_data_type
            is load_data.HospitalizationDataType.CURRENT_HOSPITALIZATIONS
            and len(hosps > 3)
        ):
            # We have converted this timeseries to new hospitalizations.
            available_timeseries.append(TimeseriesType.NEW_HOSPITALIZATIONS)
```
there is no minimum number of hospitalizations required for the hospitalization timeseries to be included.

Also, even though, for the case of `fips=31081`, there are `new_cases`, `new_deaths`, and `new_hosps`, the cases and deaths get smoothed out in the gaussian smoothing and the timeseries come back empty. 

For some reason the hospitalizations don't get removed, I think that's because the value is 6 all the way across (though i thought those should be new hosps, so not sure why that values is 6 all the way across)).

image

BrettBoval:
I think we may be mixing CURRENT_HOSPITALIZED and CUMULATIVE_HOSPITALIZED. We convert between the two in PySEIR for scoring and plotting, but it's sometimes not obvious unless you're paying attention to ModelFitter().hospitalization_data_type

PR #520, opened 2020-06-29
Feature/save raw data and rt

tashwoods:
In this PR I save the raw case and death series to the dataframe returned by infer_all for forecasting. I also factorized the get_available_timeseries code to make the code more readable.

### Please Check

- [ ] Are tests passing?

### DIFF
--- a/pyseir/inference/infer_rt.py
+++ b/pyseir/inference/infer_rt.py
@@ -661,7 +661,13 @@ class RtInferenceEngine:
         available_timeseries = self.get_available_timeseries()
 
         for timeseries_type in available_timeseries:
-            dates, times, timeseries = self.get_timeseries(timeseries_type)
+            # Add Raw Data Output to Output Dataframe
+            dates_raw, times_raw, timeseries_raw = self.get_timeseries(timeseries_type)
+            df_raw = pd.DataFrame()
+            df_raw["date"] = dates_raw
+            df_raw = df_raw.set_index("date")
+            df_raw[f"{timeseries_type.value}"] = timeseries_raw
+
             df = pd.DataFrame()
             dates, times, posteriors, start_idx = self.get_posteriors(timeseries_type)
             # Note that it is possible for the dates to be missing days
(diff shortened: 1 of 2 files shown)


pyseir/inference/infer_rt.py
```diff
@@ -842,6 +865,9 @@ def infer_all(self, plot=True, shift_deaths=0):
             plt.close()
         if df_all.empty:
             logging.warning("Inference not possible for fips: %s", self.fips)
+
+        log.info("about to save df_all.csv")
```
BrettBoval: This seems uninformative.

pyseir/inference/infer_rt.py
```diff
@@ -842,6 +865,9 @@ def infer_all(self, plot=True, shift_deaths=0):
...
+        df_all.to_csv("df_all.csv")
```
BrettBoval, line 865: Unclear how this doesn't overwrite every time? Debug code?

pyseir/inference/infer_rt.py
```diff
@@ -647,6 +669,7 @@ def infer_all(self, plot=True, shift_deaths=0):
 
             df["date"] = dates
             df = df.set_index("date")
+            df[f"{timeseries_type.value}"] = timeseries[start_idx:]
```
BrettBoval, line 668: Two things.

1. Can you add a comment why this is being added in with the given slicing on the timeseries. I can't easily reason about what indexes are being aligned here.

2. Why use the f-string instead of just `df[timeseries_type.value`?

pyseir/inference/infer_rt.py
```diff
@@ -578,26 +589,9 @@ def get_posteriors(self, timeseries_type, plot=False):
...
-            Shift the death time series by this amount with respect to cases
-            (when plotting only, does not shift the returned result).
-
-        Returns
-        -------
-        inference_results: pd.DataFrame
-            Columns containing MAP estimates and confidence intervals.
-        """
-        df_all = None
+    def get_available_timeseries(self):
```
BrettBoval, line 594: add docstring for what it does now that it is abstracted from the main infer_all call.

pyseir/inference/infer_rt.py
```diff
@@ -456,7 +467,7 @@ def get_posteriors(self, timeseries_type, plot=False):
             log.info(
                 "%s: empty timeseries %s, skipping" % (self.display_name, timeseries_type.value)
             )
-            return None, None, None
+            return None, None, None, None
```
BrettBoval, line 470: Update the docstring so that it reflects the new return signature.

pyseir/inference/infer_rt.py
tashwoods: Argh, my bad I meant to remove that

PR #521, opened 2020-06-29
WIP: Rt forecast

tashwoods:

### Please Check
- [ ] Are tests passing?


pyseir/inference/infer_rt.py
```diff
@@ -626,10 +712,266 @@ def infer_all(self, plot=True, shift_deaths=0):
...
+        # original_df = pd.DataFrame(X_scaler.inverse_transform(output_df))
+        return original_df
+
+    @staticmethod
+    def build_model(
+        MASK_VALUE, epochs, n_batch, hidden_layer_dimensions, dropout, final_train_X, final_train_Y
+    ):
+        patience = 50
+        validation_split = 0.1
+        model = Sequential()
```
ghop02, line 869: Is there anyway we could separate the code for creating the model out from inside infer_rt? This file is getting pretty big

pyseir/inference/infer_rt.py
```diff
@@ -626,10 +712,266 @@ def infer_all(self, plot=True, shift_deaths=0):
...
+            verbose=1,
+            shuffle=False,
+            validation_split=validation_split,
+            callbacks=[es],
+        )
+        logging.info("fit")
+        logging.info(history.history["loss"])
+        logging.info(history.history["val_loss"])
+        plot = True
+        if plot:
```
ghop02, line 909: Could also add a tensorboard callback if you're feeling adventurous

pyseir/inference/infer_rt.py
```diff
@@ -626,10 +712,266 @@ def infer_all(self, plot=True, shift_deaths=0):
...
+            plt.legend()
+            plt.xlabel("Epochs")
+            plt.ylabel("RMSE")
+            plt.savefig("lstm_loss_final.png")
+            plt.close("all")
+
+        return model, history
+
+    @staticmethod
+    def get_X_Y(sample_list, PREDICT_DAYS, PREDICT_VARIABLE, MASK_VALUE):
```
ghop02, line 930: I'd prefer these variables to be all lower case for consistency. What do you think about making a configuration data class that you could pass around?

```
import dataclasses

@dataclasses.dataclass
class LSTMModelConfig:
    batch_size: int = 1
    predict_days: int = ...


model_config = LSTMModelConfig()
```

I think all of the variables you have capitalized could go in the config.  The nice thing is that you can specify defaults, but then easily override values:
```
model_config = LSTMModelConfig(predict_days=100)
```

PR #526, opened 2020-06-30
Less FIPS 99999

TomGoBravo:
This PR reduces the use of FIPS 99999 by datasources, which is a step towards a world where we use only the FIPS as the region identifier.

## Test

Tested by comparing the combined latest data from master and this branch. I saw that the state aggregation values didn't change, ( 'cases', '99999') and ('deaths', '99999') were removed and bunch of new timeseries were added with FIPS such as 02999, 04999 etc.

```
git reset --hard origin/master
python ./run.py utils save-combined-csv --output-dir output/
python ./run.py utils save-combined-latest-csv --output-dir output/

git checkout tgb-fips-99999-no-more
python ./run.py utils save-combined-csv --output-dir output/
python ./run.py utils save-combined-latest-csv --output-dir output/

python ./run.py utils csv-diff output/combined-master-cc45347-20200630T141544.csv output/combined-tgb-fips-99999-no-more-0840909-20200630T143510.csv
```

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/test/us_state_abbrev_test.py
+++ b/test/us_state_abbrev_test.py
@@ -1,0 +1,17 @@
+from libs.us_state_abbrev import (
+    US_STATE_ABBREV,
+    abbrev_us_state,
+    ABBREV_US_FIPS,
+    ABBREV_US_UNKNOWN_COUNTY_FIPS,
+)
+
+
+def test():
+    assert US_STATE_ABBREV["Wisconsin"] == "WI"
+    assert abbrev_us_state["WI"] == "Wisconsin"
+    assert ABBREV_US_FIPS["TX"] == "48"
+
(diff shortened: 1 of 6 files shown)


test/us_state_abbrev_test.py
```diff
@@ -0,0 +1,17 @@
+from libs.us_state_abbrev import (
+    US_STATE_ABBREV,
+    abbrev_us_state,
+    ABBREV_US_FIPS,
+    ABBREV_US_UNKNOWN_COUNTY_FIPS,
+)
+
+
+def test():
```
ghop02, line 9: Could you give this test a slightly more descriptive name?

TomGoBravo: Ha, could hardly be less descriptive :-P Done.

PR #527, opened 2020-07-01
Api results without projections

ghop02:
This is very much a draft PR, but wanted to get it out there to make sure I'm on the right track.

There are a bunch of changes here:
 1. Consolidating all API Output to be Area specific rather than state and county specific.  There is a lot of duplication in the current API Outputs and most of the time the only logic difference is adding county name.
 2. Removing the old way of generating the projections df and putting that logic in a single class "CANPyseirOutputLocation".  Most of the calculations we weren't using anymore, so I reduced it to the ones that we are using.  If there are thoughts on other approaches here, very open to them.  I decided to go by fips to make it easier to split up the api work across cores, but could have a more aggregated approach if desired.
 3. Output from pyseir webui mapper is now a json file with keys.
 4. To calculate api results, we're running over a list of fips codes.  Right now I'm only generating the actuals but no predictions for the OBSERVED_INTERVENTION, but no others.  Is this what we want? 
 5. I still need to port over the bulk API outputs + clean up logic for the actual projections metrics and add back testing to the projection timeseries (though would love to remove it).
 6. This takes away a lot of the projections data frame validation.  I'm not 100% sure where that should live now, though it feels like the checks we had are covered in other unit tests/would show up on the output diff.

Also, this pr is pretty large, one way I could simplify it is to separate out the API output name changes.


TomGoBravo:
@ghop02 It looks like this PR introduced lint errors:
```
E:247,26: Value 'self.__root__' is unsubscriptable (unsubscriptable-object)
E:258,26: Value 'self.__root__' is unsubscriptable (unsubscriptable-object)
E:269,26: Value 'self.__root__' is unsubscriptable (unsubscriptable-object)
```

I'm guessing the problem is lint doesn't know about pydantic.Field(...) but can't find an existing issue on the web or at https://github.com/PyCQA/pylint/search?q=pydantic&type=Issues and haven't tried to make a small repro to file an issue. What do you think?

PR #529, opened 2020-07-02
Refactor Infer_Rt Step 1

BrettBoval:
This PR is the first part of a refactor for our rt inference code.

This submission does not change any of the major components. It just structures the infer_rt code in its own submodule and pulls apart the RtInferenceEngine class into more manageable chunks. Now that Alex's significant improvements have landed, it's time to pay down that technical debt while allowing us to investigate new aggregation levels. 

This change is in support of refactoring the inputs into RtInferenceEngine to be agnostic to aggregation level, which will allow for more manageable testing, maintenance, extensibility, and readability. Also, the tests for infer_rt are now under the same file in the test dir.

I will create a second branch off of this spot to freeze these changes for review.

### Please Check
- [ ] Are tests passing?


BrettBoval:
Note: This adds a couple tests from Alex and some exploratory code around test correction. The exploratory code is hidden behind a flag, and isn't exercised in the expected build path.

PR #530, opened 2020-07-02
Refactor Infer_Rt Step 2

BrettBoval:
Part 2 of the Infer_Rt changes. Still failing tests, but worth getting some feedback about the shape of the solution.


pyseir/cli.py
```diff
@@ -7,12 +7,11 @@
 import pandas as pd
 from covidactnow.datapublic import common_init
 
-from structlog_sentry import SentryProcessor
 from multiprocessing import Pool
 from functools import partial
 from libs.datasets import dataset_cache
 from pyseir.inference.initial_conditions_fitter import generate_start_times_for_state
-from pyseir.inference import infer_rt as infer_rt_module
+from pyseir.rt import infer_rt as infer_rt_module
```
ghop02, line 14: I would prefer to name this in a way that we don't have to rename the import.

pyseir/rt/constants.py
```diff
@@ -0,0 +1,75 @@
+from datetime import datetime
+
+import numpy as np
+
+
+class InferRtConstants:
```
ghop02, line 6: It would be cool if we could make this a dataclass with defaults. It'll make it easier to inject different values to test rather than needing to mock out the class.  Also, this could be `InferRtConfig`

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,647 @@
...
+    if output_df is not None and not output_df.empty:
+        output_df.to_json(county_output_file)
+
+
+def _get_display_name(fips) -> str:
+    """Need to find the right function for this. Right now just return the fips"""
+    return str(fips)
+
+
+def _generate_input_data(fips, include_testing_correction=True):
```
ghop02, line 44: This is great!

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,647 @@
...
+        self.dates = data.index
+        self.cases = data.cases
+        self.deaths = data.deaths if "deaths" in data else None
+
+        self.include_deaths = include_deaths
+        self.display_name = display_name
+        self.fips = fips
+
+        # Load the InferRtConstants (TODO: turn into class constants)
+        self.r_list = InferRtConstants.R_BUCKETS
```
ghop02, line 91: I would prefer if you passed in these params as something like 

```
def __init__(self, config: InferRtConfig = InferRtConfig({}):
    self.config = config

def infer_all(self, fips, data, ...):
    
```

This will definitely require a bit more refactoring, but I think it gets to a much better separation of concerns.

pyseir/rt/utils.py
```diff
@@ -0,0 +1,236 @@
...
+                    pr=p_po_am,
+                    dd=driving_likelihood,
+                    up=c_po_am - c_pr_am,
+                    po=c_po_am,
+                    lag=lag_after_update,
+                    cmp=compare_lag,
+                    ind=ind,
+                )
+            )
+        if noLag:  # End of lagging sequence of days
```
ghop02, line 71: `no_lag`

test/mocks/inference/load_data.py
```diff
@@ -0,0 +1,117 @@
...
+        else:  # linear growth
+            self.last_value = self.scale + self.growth_rate * (time - self.t0)
+        return self.last_value
+
+
+def _get_cases_for_times(generator, times):
+    return np.array(list(map(generator.generate_data, times)))
+
+
+def load_new_case_data_by_state(state, ref_date, include_testing_correction):
```
ghop02, line 74: How are these mocks used?

pyseir/cli.py
BrettBoval: Agreed. I add a change that makes the cli function call more consistent (prepend run_), which technically will break backwards compatibility. We can decide before merge if that break is unacceptable, but I think it's only used internally.

BrettBoval:
Hey @amacaula. The tests are now back up and running. I think it's worth having you take a look. I think in the refactor I've dropped the don't start calculating until your first nonzero. The tests don't catch that yet, so I thought we could add a test for it, and then fix it if needed.

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+import numpy as np
+import pandas as pd
+from scipy import stats as sps
+
+from pyseir import load_data
+from pyseir.utils import TimeseriesType, get_run_artifact_path, RunArtifact
+from pyseir.rt.constants import InferRtConstants
+from pyseir.rt import plotting, utils
+
+rt_log = structlog.get_logger(__name__)
```
ghop02, line 15: I would prefer to call this `_logger` or `logger` to keep consistent across files

BrettBoval: Ahh. I missed the __name__ call. I wanted to have this be a new logger that was just in infer_rt scope for comprehensibility (so without args), but now I'm not sure that's what I want. I'll have to check what context gets passed in by using that logger and if it's worth it. Either way, I'll either rename it or call a new instance.

pyseir/rt/infer_rt.py
ghop02: I would still prefer to inject these values rather than directly accessing the class here. Also, I think that it would be helpful to just have a variable `self.config = rt_config` or even `self.config = InferRtConstants`, to remove the need to write out all the variables twice in the init

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+        Just used for output paths. Should remove with display_name later.
+    """
+
+    def __init__(
+        self, data, display_name, fips, include_deaths=False,
+    ):
+
+        self.dates = data.index
+        self.cases = data.cases if "cases" in data else None
+        self.deaths = data.deaths if "deaths" in data else None
```
ghop02, line 130: Have you thought about what this class would look like if dates, cases, and deaths weren't instance variables? I'm just thinking about how best to separate out the class specific logic

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+            High confidence intervals.
+        """
+        posterior_cdfs = posteriors.values.cumsum(axis=0)
+        low_idx_list = np.argmin(np.abs(posterior_cdfs - (1 - ci)), axis=0)
+        high_idx_list = np.argmin(np.abs(posterior_cdfs - ci), axis=0)
+        ci_low = self.r_list[low_idx_list]
+        ci_high = self.r_list[high_idx_list]
+        return ci_low, ci_high
+
+    def make_process_matrix(self, timeseries_scale=InferRtConstants.SCALE_SIGMA_FROM_COUNT):
```
ghop02, line 220: This isn't really accessing any instance specific variables other than constants, do you think this could be a function outside of the InferRt class?

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+        times: array-like
+            Output integers since the reference date.
+        posteriors: pd.DataFrame
+            Posterior estimates for each timestamp with non-zero data.
+        start_idx: int
+            Index of first Rt value calculated from input data series
+            #TODO figure out why this value sometimes truncates the series
+            
+        """
+        dates, timeseries = self.get_timeseries(timeseries_type=timeseries_type)
```
ghop02, line 295: Could you pass in dates and timeseries rather than access it in this function? Makes testing this function easier because we can play around with the timeseries without having to rely on the class

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+                prior_am=current_prior.argmax(),
+                like_am=likelihoods[current_day].argmax(),
+                post_am=numerator.argmax(),
+            )
+
+            # Add to the running sum of log likelihoods
+            log_likelihood += np.log(denominator)
+            loop_idx += 1
+
+        self.log_likelihood = log_likelihood
```
ghop02, line 415: instead of setting this class value, we could return it below

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+            # ------------------------------------------------
+            # Compute the indicator lag using the curvature
+            # alignment method.
+            # ------------------------------------------------
+            if (
+                timeseries_type in (TimeseriesType.NEW_DEATHS,)
+                and f"Rt_MAP__{TimeseriesType.NEW_CASES.value}" in df_all.columns
+            ):
+
+                # Go back up to 30 days or the max time series length we have if shorter.
```
ghop02, line 519: This code block feels like it could be a separate function

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+                [1.0 * np.ones(len(df_all) - len(tail_sup)), tail_sup.values]
+            )
+            # Adjust rt by undoing the suppression
+            df_all["Rt_MAP_composite"] = (df_all["Rt_MAP_composite"] - 1.0) / np.power(
+                suppression, self.tail_suppression_correction
+            ) + 1.0
+
+        # Optionally Smooth just Rt_MAP_composite.
+        # Note this doesn't lag in time and preserves integral of Rteff over time
+        for i in range(0, self.smooth_rt_map_composite):
```
ghop02, line 582: Smoothing also feels like it could be a separate function

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+
+        if np.sum(cases) > self.min_cases:
+            available_timeseries.append(TimeseriesType.NEW_CASES)
+
+        if np.sum(deaths) > self.min_deaths:
+            available_timeseries.append(TimeseriesType.NEW_DEATHS)
+
+        return available_timeseries
+
+    def infer_all(self, plot=True, shift_deaths=0):
```
ghop02, line 448: My ideal way to call this function would be:
```
def infer_all(self, timeseries: List[pd.Series], plot=True, shift_deaths=0):
```
or maybe
```
def infer_all(self, fips: str, timeseries: List[pd.Series], plot=True, shift_deaths=0):
```

that way it frees a lot of the work from the class state and easier to see how timeseries are passed through.

pyseir/rt/utils.py
```diff
@@ -0,0 +1,236 @@
...
+                ]
+            )
+            self.total_lag += lag_after_update
+            if abs(lag_after_update) > abs(self.max_lag):
+                self.max_lag = lag_after_update
+        self.last_lag = lag_after_update
+
+
+def replace_outliers(
+    x,
```
ghop02, line 107: could you add types here?

pyseir/rt/utils.py
```diff
@@ -0,0 +1,236 @@
...
+        self.last_lag = 0
+        self._reset_()
+
+    def _reset_(self):
+        self.lag_days_running = list()
+        self.max_lag = 0
+        self.total_lag = 0
+
+    def evaluate_lag_using_argmaxes(
+        self, current_day, current_sigma, prev_post_am, prior_am, like_am, post_am
```
ghop02, line 33: types would be helpful

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+        df=pd.DataFrame(dict(cases=observed_new_cases, deaths=observed_new_deaths), index=date),
+        include_deaths=include_deaths,
+    )
+    return df
+
+
+def filter_and_smooth_input_data(df: [pd.DataFrame], include_deaths=False) -> pd.DataFrame:
+    """Do Everything Strange Here Before it Gets to the Inference Engine"""
+    MIN_CUMULATIVE_COUNTS = dict(cases=20, deaths=10)
+    MIN_INCIDENT_COUNTS = dict(cases=5, deaths=5)
```
ghop02, line 67: I'd prefer if these weren't caps - I expect capitalized to be more global values - Is there a reason this doesn't belong in RtConstants?

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+        requirements = [  # All Must Be True
+            df[column].count() > InferRtConstants.MIN_TIMESERIES_LENGTH,
+            df[column].sum() > MIN_CUMULATIVE_COUNTS[column],
+            df[column].max() > MIN_INCIDENT_COUNTS[column],
+        ]
+        # Now Apply Input Outlier Detection and Smoothing
+        filtered = utils.replace_outliers(df[column], log=rt_log)
+        assert len(filtered) == len(df[column])
+        smoothed = filtered.rolling(
+            InferRtConstants.COUNT_SMOOTHING_WINDOW_SIZE,
```
ghop02, line 80: I'd recommend passing in the infer rt constants (ideally as an instantiated dataclass) as a variable: `def filter_and_smooth_input_data(config: InferRtConfig, df: List[pd.DataFrame], include_deaths=False)`

Then it makes it easy to do `config.count_smoothing_window_size`.  In my mind it decouples this function from a specific object and gives you a bit more flexibility

pyseir/rt/infer_rt.py
```diff
@@ -0,0 +1,617 @@
...
+    if output_df is not None and not output_df.empty:
+        output_df.to_json(county_output_file)
+
+
+def _get_display_name(fips) -> str:
+    """Need to find the right function for this. Right now just return the fips"""
+    return str(fips)
+
+
+def _generate_input_data(fips, include_testing_correction=True, include_deaths=True):
```
ghop02, line 44: `_fetch_and_smooth_input_data`? `_fetch_and_preprocess_input_data`?

BrettBoval:
Before and after snapshots with the same data is now complete. https://covidactnow.org/internal/compare?left=703&locations=0&metric=4&right=704&sort=2. Small rounding changes(+/- 0.01) between the two, but no color change.

BrettBoval:
Alex and I went over all the changes this evening and agreed it was ready to commit.

PR #531, opened 2020-07-02
Dont raise error on model fit failure

ghop02:
Looks like we were accidentally raising an error on a single model fitter failure rather than carrying on.  This should allow the build to not fail if one region fails


BrettBoval:
I have aspirations for this to be better, but obviously too strict now.

ghop02:
very much same

PR #533, opened 2020-07-06
Adding max tasks per child to reduce memory consumptione

ghop02:
It looks like the api build is getting killed because of an OOM kill. this will reduce the usage similarly as to how we do in pyseir by setting maxtasksperchild to 1.

### DIFF
--- a/libs/pipelines/api_pipeline.py
+++ b/libs/pipelines/api_pipeline.py
@@ -45,7 +45,7 @@ def run_on_all_fips_for_intervention(
 
     # Load interventions outside of subprocesses to properly cache.
     get_can_projection.get_interventions()
-    pool = pool or multiprocessing.Pool()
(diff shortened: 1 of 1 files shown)


libs/pipelines/api_pipeline.py
```diff
@@ -45,7 +45,7 @@ def run_on_all_fips_for_intervention(
 
     # Load interventions outside of subprocesses to properly cache.
     get_can_projection.get_interventions()
-    pool = pool or multiprocessing.Pool()
+    pool = pool or multiprocessing.Pool(maxtasksperchild=1)
```
mikelehen, line 48: Please add _some_ sort of comment describing the motivation for this so that somebody looking at this in the future knows why we set this and what they should worry about if they change it.

PR #535, opened 2020-07-07
Updating api schema

ghop02:
Updating the API json-schema outputs - it's been a while since they were last updated. 

### Please Check
- [x] Are tests passing?


api/can_api_definition.py
```diff
@@ -13,7 +13,7 @@
 """
 
 
-class _ResourceUsageProjection(base_model.BaseModel):
+class ResourceUsageProjection(base_model.APIBaseModel):
     peakShortfall: int = pydantic.Field(
         ..., description="Shortfall of resource needed at the peek utilization"
```
BrettBoval, line 18: Might as well fix the typos to "peak utilization".

PR #536, opened 2020-07-07
Updating api schema

ghop02:
Adds a few missing schemas that didn't generate because we weren't recursively finding subclasses of APIBaseModel


ghop02:
> Are there two things coming in (1) The recursion step for loading and (2) adding in/back schema defs?

Yeah, basically the recursion step enables outputting those new schema files.  As in, they were generated after the recursion was enabled

PR #537, opened 2020-07-07
Feature/rt forecast

tashwoods:

### Please Check
- [ ] Are tests passing?


requirements.txt
```diff
@@ -38,4 +38,6 @@ dictdiffer==0.8.1
 structlog==20.1.0
 structlog-sentry==1.2.2
 more_itertools==8.3.0
+tensorflow==2.1.0
+keras==2.3.1
```
ghop02, line 42: Is there anything you need from keras that tf.keras doesn't have?

PR #542, opened 2020-07-10
Explore/limit test fraction to 0.1

BrettBoval:
Check out snapshot 654. I've fixed hosp_fraction to 1 and test_fraction to 0.1 . It does as expected on my local machine for the state and counties of California. We were getting %infected numbers that were unreasonably high when we let test_fraction float, especially in places like Marin County. This seems to fix those parts of the model that aren't captured in the chi2 scoring.

I would like to merge this before tomorrow morning's 9:30am build.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/inference/model_fitter.py
+++ b/pyseir/inference/model_fitter.py
@@ -93,8 +93,8 @@ class ModelFitter:
         limit_t_delta_phases=[14, 100],  # good as of June 3, 2020 may need to update in the future
         error_t_delta_phases=1,
         test_fraction=0.1,
-        limit_test_fraction=[0.02, 1],
-        # fix_test_fraction=True,
+        # limit_test_fraction=[0.02, 1],
+        fix_test_fraction=True,
         error_test_fraction=0.02,
         hosp_fraction=1,
         # fix_hosp_fraction=True,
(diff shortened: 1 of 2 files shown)


pyseir/inference/model_fitter.py
```diff
@@ -225,18 +227,18 @@ def set_inference_parameters(self):
...
         ]
         if self.fips in overwrite_params_df["fips"].values:
             this_fips_df = overwrite_params_df.loc[overwrite_params_df["fips"] == self.fips]
             for param in INITIAL_PARAM_SETS:
                 self.fit_params[param] = this_fips_df[param]
 
-        self.fit_params["fix_hosp_fraction"] = self.hospitalizations is None
-        if self.fit_params["fix_hosp_fraction"]:
-            self.fit_params["hosp_fraction"] = 1
+        # self.fit_params["fix_hosp_fraction"] = self.hospitalizations is None
```
ghop02, line 239: Can you add some context as to why this is commented out in a comment? Generally I find it helpful to comment why it's commented (as opposed to removed), are you planning on brining this code back ?

tashwoods:
This will likely have a notable impact. The code itself looks fine, but this will remove the renormalization of the case and hospitalization curves previously from these fitted fractions. Imo, we need to compare snapshots for all states and notable counties before setting the test_fraction = 0.1 and hosp_fraction = 1. 

Long term it would be better to infer the test fraction from the IFR/CFR/other data. Not sure about inferring the hosp_fraction.

PR #543, opened 2020-07-10
Changing API Names from Area -> Region to match FE

ghop02:


ghop02:
💯  on all of the suggestions! No better time like the present to improve the naming.

PR #547, opened 2020-07-14
Streamlining pyseir cli to always take in lists of states

ghop02:
This just refactors the pyseir cli a bit.  It pushes more responsibility to calling functions to know the list of states we will call with.  The intent here is to reduce the recursion in the helper functions and make those easier for me (and hopefully others) to understand.

### DIFF
--- a/pyseir/cli.py
+++ b/pyseir/cli.py
@@ -49,63 +49,44 @@ def _generate_whitelist():
     gen.generate_whitelist()
 
 
-def _impute_start_dates(state=None, states_only=False):
+def _impute_start_dates(states, states_only=False):
     if states_only:
         raise NotImplementedError(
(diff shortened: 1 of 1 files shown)


BrettBoval:
I am happy to accept this as an incremental step.

However, I'm pretty sure we can switch more directly to using fips (regardless of aggregation level). Right now, we are calling a different function for infer_rt if it's a state or a county, even though the code is now agnostic to that distinction.

ghop02:
> I am happy to accept this as an incremental step.
> 
> However, I'm pretty sure we can switch more directly to using fips (regardless of aggregation level). Right now, we are calling a different function for infer_rt if it's a state or a county, even though the code is now agnostic to that distinction.

👍  I think I'd prefer to get this in as an incremental step - but totally agree on fips as the key/next steps!

PR #552, opened 2020-07-15
Adding ability to load previous persisted datasets from git lfs history

ghop02:
Adds some ability to query dataset history through combined datasets api.

Access the previous commit:

```
combined_datasets.load_us_timeseries_dataset(previous_commit=True)
combined_datasets.load_us_latest_dataset(previous_commit=True)
```

Access the first commit before a date:
```
combined_datasets.load_us_timeseries_dataset(on_or_before="2020-07-15")
combined_datasets.load_us_latest_dataset(on_or_before="2020-07-15")
```

This does a bit of work to fetch the git lfs object if it has not been downloaded locally which made this a bit more complicated than I had anticipated.


libs/git_lfs_object_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+        _logger.info("Fetching missing reference", fetch_command=lfs_fetch_command)
+        subprocess.check_output(lfs_fetch_command, shell=True)
+
+    return object_path.read_bytes()
+
+
+def get_data_for_path(
+    repo: git.Repo, path: str, on_or_before: str = None, previous_commit=False
+) -> bytes:
+    commit = find_commit(repo, path, on_or_before=on_or_before, previous_commit=previous_commit)
```
TomGoBravo, line 95: path is typed as a str here but a Path in the find_commit parameters. I prefer the stronger type, Path.

libs/git_lfs_object_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+    repo: git.Repo, path: pathlib.Path, on_or_before: str = None, previous_commit: bool = False
+) -> Optional[git.Commit]:
+    """Find a commit for a given path matching query options.
+
+    If no query options are specified, returns latest commit.
+
+    Args:
+        repo: Git Repo.
+        path: file path.
+        on_or_before: Optional ISO format date string.  If set, will return the first commit on or
```
TomGoBravo, line 39: how about making this a datetime.date ?

libs/git_lfs_object_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+    Args:
+        repo: Git Repo
+        path: File path.
+        commit: Commit object to read lfs data for.
+
+    Returns: Bytes for file at commit.
+    """
+    blob = commit.tree / path
+    # TODO(chris): Find a better way to identify a Git LFS pointer file.
+    is_pointer = blob.size < 200
```
TomGoBravo, line 75: how about always calling _get_object_sha_from_lfs_pointer, and if it raises assume it isn't an LFS file?

libs/git_lfs_object_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+    if not is_pointer:
+        return blob.data_stream.read()
+
+    _logger.debug("Loading commit data from path at commit", path=path, commit=commit)
+    pointer_text = blob.data_stream.read().decode()
+    object_sha = _get_object_sha_from_lfs_pointer(pointer_text)
```
TomGoBravo, line 81: instead of parsing the lfs file how about something like `return subprocess.check_output(['git', 'lfs', 'smudge'], input=pointer_text)` ?

ghop02: this is great! didn't know what smudge does

libs/datasets/dataset_pointer.py
```diff
@@ -51,7 +53,32 @@ def save_dataset(self, dataset: DatasetBase) -> pathlib.Path:
         _logger.info("Successfully saved dataset", path=str(self.path))
         return self.path
 
-    def load_dataset(self) -> DatasetBase:
+    def load_dataset(
+        self, before: str = None, previous_commit: bool = False, commit: str = None
```
TomGoBravo, line 57: instead of having a bunch of parameters how about a string you pass to `git rev-list -n` ? That supports dates, branches, HEAD^, HEAD^^ and more

libs/git_lfs_object_helpers.py
ghop02: I actually think that when you reference the blob it will always return the git lfs pointer, because it's referring to what's in the git tree, which is the lfs pointer

libs/datasets/dataset_pointer.py
ghop02: I do like the idea but don't quite have the energy right now to do it, would you be alright if I add this as a TODO and keep it how it is?

TomGoBravo: No problem. Please add the comment.

libs/git_lfs_object_helpers.py
```diff
@@ -0,0 +1,90 @@
...
+    repo: git.Repo = None,
+    before: str = None,
+    previous_commit=False,
+    commit: str = None,
+) -> bytes:
+    repo = repo or git.Repo(dataset_utils.REPO_ROOT)
+    commit = find_commit(
+        repo, path, before=before, previous_commit=previous_commit, commit_sha=commit
+    )
+    return read_lfs_data_for_commit(repo, path, commit)
```
TomGoBravo, line 90: does this still work if `path` is not in LFS?

libs/datasets/dataset_pointer.py
```diff
@@ -51,7 +53,32 @@ def save_dataset(self, dataset: DatasetBase) -> pathlib.Path:
...
+            # When loading dataset from a previous point in time, the data is not available as path.
+            # To get around this, we save the bytes of the commit data to a temporary file to then
+            # load as a csv.
+            with tempfile.NamedTemporaryFile() as tmp_file:
+                path = pathlib.Path(tmp_file.name)
+                lfs_data = git_lfs_object_helpers.get_data_for_path(
+                    self.path, before=before, previous_commit=previous_commit, commit=commit
+                )
+                path.write_bytes(lfs_data)
+                return self.dataset_type.dataset_class.load_csv(path)
```
TomGoBravo, line 80: can you change TimeseriesDataset.load_csv to accept `path_or_buf: Union[pathlib.Path, TextIO]` like common_df.read_csv ? Then I think you'll be able to keep the data in memory without using a temp file.

libs/git_lfs_object_helpers.py
ghop02: Looks like it does! Changing the name of the function to reflect that

libs/datasets/dataset_pointer.py
```diff
@@ -71,13 +63,12 @@ def load_dataset(
             # When loading dataset from a previous point in time, the data is not available as path.
             # To get around this, we save the bytes of the commit data to a temporary file to then
```
TomGoBravo, line 64: update or rm stale comment

PR #557, opened 2020-07-16
add DatasetBase.yield_records

TomGoBravo:
## This PR

* adds DatasetBase.yield_records, replacing `records` property that breaks in a future PR
* replaces use of `get_us_latest_for_state` with 2 digit code passed to `get_records_for_fips` 
* adds `test_fill_fields_with_data_source_nan_overwrite` that exercises the unexpected current behavior when combining datasets
* make get_subset aggregation_level parameter optional, it was required
* add `_` prefix to some functions that are internal to a class

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -81,9 +81,10 @@ class TimeseriesDataset(dataset_base.DatasetBase):
 
         if aggregation_level == AggregationLevel.COUNTY:
             group = [CommonFields.COUNTRY, CommonFields.STATE, CommonFields.FIPS]
-        if aggregation_level == AggregationLevel.STATE:
+        elif aggregation_level == AggregationLevel.STATE:
             group = [CommonFields.COUNTRY, CommonFields.STATE]
-        if aggregation_level == AggregationLevel.COUNTRY:
+        else:
+            assert aggregation_level == AggregationLevel.COUNTRY
             group = [CommonFields.COUNTRY]
 
         data = self.data[
@@ -128,8 +129,7 @@ class TimeseriesDataset(dataset_base.DatasetBase):
 
         Returns: List of dictionary records with NA values replaced to be None
         """
-        subset = self.get_subset(AggregationLevel.COUNTY, fips=fips)
-        return subset.records
+        return list(self.get_subset(aggregation_level=AggregationLevel.COUNTY, fips=fips).yield_records())
 
     def get_records_for_state(self, state) -> List[dict]:
         """Get data for state.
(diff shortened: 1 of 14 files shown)


libs/datasets/timeseries.py
```diff
@@ -74,16 +74,16 @@ def latest_values(self, aggregation_level=None) -> pd.DataFrame:
 
         Return: DataFrame
         """
-        if not aggregation_level:
+        if aggregation_level is None:
```
ghop02, line 77: Why this change?

libs/datasets/timeseries.py
```diff
@@ -74,16 +74,16 @@ def latest_values(self, aggregation_level=None) -> pd.DataFrame:
...
             county = self.latest_values(aggregation_level=AggregationLevel.COUNTY)
             state = self.latest_values(aggregation_level=AggregationLevel.STATE)
             return pd.concat([county, state])
-
-        if aggregation_level == AggregationLevel.COUNTY:
+        elif aggregation_level == AggregationLevel.COUNTY:
```
ghop02, line 81: apparently the preferred way of comparing enums is with `is`.  i.e. `aggregation_level is AggregationLevel.COUNTY`

pyseir/parameters/parameter_ensemble_generator.py
```diff
@@ -39,10 +39,9 @@ def __init__(self, fips, N_samples, t_list, I_initial=1, suppression_policy=None
                 load_data.load_county_metadata().set_index("fips").loc[fips].to_dict()
             )
             self.state_abbr = us.states.lookup(self.county_metadata["state"]).abbr
-            self._latest = combined_datasets.get_us_latest_for_fips(self.fips)
         else:
             self.state_abbr = us.states.lookup(fips).abbr
```
ghop02, line 43: Are `self.state_abbr` and `self.county_metadata` used anywhere else? You may be able to remove this whole block

libs/datasets/timeseries.py
TomGoBravo: Done

libs/datasets/timeseries.py
TomGoBravo: No particular reason other than touching code to do with aggregation_level and thinking it is a little cleaner, but I can also see why having the `is None` on its own is better.

pyseir/parameters/parameter_ensemble_generator.py
TomGoBravo: Good call! done.

PR #558, opened 2020-07-16
changing definition of Rt in projections field

ghop02:


api/can_api_definition.py
```diff
@@ -32,8 +32,8 @@ class Projections(base_model.APIBaseModel):
     ICUBeds: Optional[ResourceUsageProjection] = pydantic.Field(
         ..., description="Projection about ICU hospital bed utilization"
     )
-    Rt: float = pydantic.Field(..., description="Historical or Inferred Rt")
-    RtCI90: float = pydantic.Field(..., description="Rt standard deviation")
+    Rt: float = pydantic.Field(..., description="Inferred Rt")
+    RtCI90: float = pydantic.Field(..., description="Rt 90th percentile confidence interval.")
```
TomGoBravo, line 36: nit: it doesn't make sense to have a float represent an interval. "Rt 90th percentile confidence interval upper endpoint." ?

ghop02: Sounds good!

PR #559, opened 2020-07-17
fix latest combined column order

TomGoBravo:
## This PR

* sort columns and rows in LatestValuesDataset before writing
* drop `index` column in LatestValuesDataset
* Dropping rows without FIPS in Timeseries, as produced by JHU but hard to filter in data source class because of the cleanup done in the timeseries class

## Tested

`data/timeseries.csv` is identical.

`git show master:data/latest.csv | git lfs smudge > data/latest-master.csv` and then in jupyter: 
```
import pandas as pd
from libs.datasets.common_fields import CommonFields

path = 'data/latest-master.csv'
df = common_df.sort_common_field_columns(pd.read_csv(path, dtype={CommonFields.FIPS: str}).set_index(CommonFields.FIPS).sort_index())
df.to_csv(path, date_format="%Y-%m-%d", index=True, float_format="%.12g")
```
and the only difference is the empty index column dropped.

### DIFF
--- a/libs/datasets/latest_values_dataset.py
+++ b/libs/datasets/latest_values_dataset.py
@@ -1,6 +1,7 @@
 from typing import Type, List, Optional
 import pathlib
 
+import structlog
 from more_itertools import first
 
 from covidactnow.datapublic import common_df
@@ -186,7 +187,8 @@ class LatestValuesDataset(dataset_base.DatasetBase):
         """
         # Cannot use common_df.write_csv as it doesn't support data without a date index field.
         data = self.data.set_index(CommonFields.FIPS).replace({pd.NA: np.nan}).convert_dtypes()
-        data = common_df.sort_common_field_columns(data)
+        data = common_df.only_common_columns(data, structlog.get_logger())  # Drops `index`
+        data = common_df.sort_common_field_columns(data).sort_index()
         data.to_csv(path, date_format="%Y-%m-%d", index=True, float_format="%.12g")
 
     @classmethod
--- a/libs/datasets/latest_values_dataset.py
+++ b/libs/datasets/latest_values_dataset.py
@@ -3,6 +3,7 @@ import pathlib
 
 from more_itertools import first
 
+from covidactnow.datapublic import common_df
 from libs import us_state_abbrev
 import pandas as pd
 import numpy as np
@@ -185,6 +186,7 @@ class LatestValuesDataset(dataset_base.DatasetBase):
         """
         # Cannot use common_df.write_csv as it doesn't support data without a date index field.
         data = self.data.set_index(CommonFields.FIPS).replace({pd.NA: np.nan}).convert_dtypes()
+        data = common_df.sort_common_field_columns(data)
         data.to_csv(path, date_format="%Y-%m-%d", index=True, float_format="%.12g")
 
     @classmethod
--- a/libs/datasets/sources/jhu_dataset.py
+++ b/libs/datasets/sources/jhu_dataset.py
@@ -73,12 +73,14 @@ class JHUDataset(data_source.DataSource):
             loaded_data.append(data)
 
         data = pd.concat(loaded_data)
-        data = self.standardize_data(data)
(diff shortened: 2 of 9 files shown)


libs/datasets/latest_values_dataset.py
```diff
@@ -185,6 +187,8 @@ def to_csv(self, path: pathlib.Path):
         """
         # Cannot use common_df.write_csv as it doesn't support data without a date index field.
         data = self.data.set_index(CommonFields.FIPS).replace({pd.NA: np.nan}).convert_dtypes()
+        data = common_df.only_common_columns(data, structlog.get_logger())  # Drops `index`
+        data = common_df.sort_common_field_columns(data).sort_index()
```
ghop02, line 191: Should we move this into the common_df? The main reason it's not there is because `common_df` mostly expects a date column, but a good portion of datasets won't have that.

libs/datasets/sources/jhu_dataset.py
```diff
@@ -73,12 +73,14 @@ def __init__(self, input_dir):
...
-        data = self.standardize_data(data)
+        data = self._standardize_data(data)
         super().__init__(data)
 
     @classmethod
-    def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
+    def _standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
         data = dataset_utils.strip_whitespace(data)
+        # Drop all data outside the USA to speed up debugging
+        data = data.loc[data[cls.Fields.COUNTRY].isin(["US", "USA"])]
```
ghop02, line 83: 🙌

libs/datasets/timeseries.py
```diff
@@ -202,6 +203,17 @@ def from_source(
         state_fips = data.loc[is_state, CommonFields.STATE].map(us_state_abbrev.ABBREV_US_FIPS)
         data.loc[is_state, CommonFields.FIPS] = state_fips
 
+        no_fips = data[CommonFields.FIPS].isnull()
+        if no_fips.any():
+            print(f"Dropping rows without FIPS in {source}:\n{data.loc[no_fips]}")
```
ghop02, line 208: _logger.warning?

libs/datasets/timeseries.py
```diff
@@ -202,6 +203,17 @@ def from_source(
...
+            data = data.loc[~no_fips]
+
+        if data.duplicated(COMMON_FIELDS_TIMESERIES_KEYS, keep=False).any():
+            raise ValueError(
+                f"Duplicates in {source}:\n"
+                f"{data.loc[data.duplicated(COMMON_FIELDS_TIMESERIES_KEYS)]}"
```
ghop02, line 214: If possible, I'd prefer not to include the data in the Exception message.  This makes it difficult to combine common alerts and will (I believe) create individual errors in sentry.  I've gotten around this by either setting it as a variable before the exception or creating a new exception with this as an input, something like:
```python
class DuplicateRowError(Exception):
    def __init__(self, data_source, duplicated_data):
        super().__init__("Duplicates in Data Source")
        self.data_source = data_source
        self.duplicated_data = duplicated_data

```

libs/datasets/latest_values_dataset.py
TomGoBravo: I think the bad portion of datasets don't have a date column :-P Medium term I'd like to stop using latest.csv (to reduce the number of different indexes we use, so adding provenance is easier) so don't think it is worth shuffling the code around.

libs/datasets/timeseries.py
TomGoBravo: I created a new exception... though this is not expected to happen regularly because the build will stop. FYI: In the past I've used attrs to make exception subclasses but can't do that with pydantic: https://stefan.sofa-rockers.org/2020/05/29/attrs-dataclasses-pydantic/

libs/datasets/timeseries.py
TomGoBravo: done

PR #563, opened 2020-07-17
combine refactor

TomGoBravo:
This PR addresses https://trello.com/c/Itlq1xAM/302-finish-combined-dataset-refactor-to-ensure-we-use-a-single-data-source-per-column

* Replaces `fill_fields...` methods in `libs/datasets/dataset_utils.py` with `_build_dataframe` in `libs/datasets/combined_datasets.py`
* `data/latest.csv` is byte-for-byte identical
* `data/timeseries.csv` gets a new `county` column containing the county name
* `DatasetBase` and subclasses have new class attribute `COMMON_INDEX_FIELDS` listing fields used for indexing.
* tests in `test/dataset_utils_test.py` are modified to pass using the new implementation

### Please Check
- [ ] Are tests passing?


libs/datasets/combined_datasets.py
```diff
@@ -269,28 +254,126 @@ def _build_combined_dataset_from_sources(
...
+
+    # Build feature columns from feature_definitions.
+    data = pd.DataFrame(index=new_index)
+    for field_name, data_source_names in feature_definitions.items():
+        log.info("Working field", field=field_name)
+        field_out = None
+        # Go through the data sources, starting with the highest priority.
+        for datasource_name in reversed(data_source_names):
+            with tmp_bind(log, dataset_name=datasource_name, field=field_name) as log:
+                datasource_field_in = datasource_dataframes[datasource_name][field_name]
```
ghop02, line 351: Thoughts on pulling this series replacement code out into its own function? I don't think it's necessary, but would make test writing a bit easier for this (fairly complicated) logic.

libs/datasets/combined_datasets.py
```diff
@@ -73,7 +77,6 @@
         CovidTrackingDataSource,
         NevadaHospitalAssociationData,
     ],
-    CommonFields.COUNTY: [FIPSPopulation],
```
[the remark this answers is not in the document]
TomGoBravo: `preserve_columns` makes sure it doesn't get dropped and it is in the output so looks like it got included somehow :-)

libs/datasets/combined_datasets.py
TomGoBravo: Proof by example that testing the complex logic is possible as-is... at least I'm pretty sure  test_build_timeseries_override exercises each of the override types.

PR #564, opened 2020-07-21
Infer Rt Flag and Logging Cleanup

BrettBoval:
This PR combines multiple points where we use the include_deaths flag into one flag. The data loading setup steps will also exit if neither case nor death data meet any of the filtering requirements. This makes it clearer where infer_rt "failed to generate output" by separating cases where we gave it no input from cases vs where it failed to find a solution given the non-empty input.

This should have a small speedup in the infer_rt code because (1) the engine won't be initialized on empty input data and (2) the include_deaths flag will short-circuit the engine from calculating the deaths based rt (which is dropped at the end by default).

### Please Check
- [x] Are tests passing?

### DIFF
--- a/pyseir/rt/infer_rt.py
+++ b/pyseir/rt/infer_rt.py
@@ -16,27 +16,38 @@ from pyseir.rt import plotting, utils
 rt_log = structlog.get_logger(__name__)
 
 
-def run_rt_for_fips(fips, figure_collector=None):
+def run_rt_for_fips(
+    fips, include_deaths=False, include_testing_correction=True, figure_collector=None
+):
     """Entry Point for Infer Rt"""
 
     # Generate the Data Packet to Pass to RtInferenceEngine
     input_df = _generate_input_data(
-        fips, include_testing_correction=True, figure_collector=figure_collector
+        fips=fips,
+        include_testing_correction=include_testing_correction,
+        include_deaths=include_deaths,
+        figure_collector=figure_collector,
     )
+    if input_df.dropna().empty:
+        rt_log.warning(event="Infer Rt Skipped. No Data Passed Filter Requirements:", fips=fips)
+    else:
+        # Save a reference to instantiated engine (eventually I want to pull out the figure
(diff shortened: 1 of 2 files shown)


pyseir/rt/infer_rt.py
```diff
@@ -16,37 +16,46 @@
...
-    # and saving so that I don't have to pass a display_name and fips into the class
-    engine = RtInferenceEngine(
-        data=input_df, display_name=_get_display_name(fips), fips=fips, include_deaths=False
+        fips=fips,
+        include_testing_correction=include_testing_correction,
+        include_deaths=include_deaths,
+        figure_collector=figure_collector,
     )
+    if input_df.dropna().empty:
+        rt_log.warning(event="Infer Rt Skipped. No Data Passed Filter Requirements:", fips=fips)
```
ghop02, line 32: I think it's clearer here if you return at this point, then everything below doesn't ahve to be indented

pyseir/rt/infer_rt.py
```diff
@@ -64,14 +73,15 @@ def _generate_input_data(
...
         figure_collector=figure_collector,
         display_name=fips,
+        log=rt_log.new(fips=fips),
     )
     return df
 
 
 def filter_and_smooth_input_data(
-    df: [pd.DataFrame], display_name, include_deaths=False, figure_collector=None
+    df: [pd.DataFrame], display_name, include_deaths, figure_collector, log
```
ghop02, line 82: can you add types to the rest of the variables?  Also, is df `pd.DataFrame` or truly a list?

pyseir/rt/infer_rt.py
BrettBoval: Done. I have a weak preference to use if/else logic with the return at the end when there is only one logic branching event, but it's clear we could add additional short-circuit events in here around how output_df returns.

pyseir/rt/infer_rt.py
BrettBoval: Done. And df is more generally a dict-like object with at least the keys "cases" and "deaths".

PR #567, opened 2020-07-23
Add summary generation

ghop02:
This adds in dataset summary generation along with the data update + adds the summary file to the qa folder in the api build. 

It saves a summary csv of all fips with the following columns per variable:
has_value, min_date, max_date, max_value, min_value, latest_value, num_observations, largest_delta, largest_delta_date.

I decided to save it alongside the timeseries.csv files so that we can easily use the same tooling to comparing the timeseries files at different points in time.

The summary report adds about 2 minutes on to the data update time. There's probably an easier way to do a parallel pandas apply, which would speed up things considerably - will look into that.

This does not yet have data provenance.

### DIFF
--- a/libs/qa/dataset_summary.py
+++ b/libs/qa/dataset_summary.py
@@ -2,6 +2,7 @@ from typing import Optional, Tuple
 
 import pydantic
 import pandas as pd
+import numpy as np
 from covidactnow.datapublic.common_fields import CommonFields
 from libs.datasets.dataset_utils import AggregationLevel
 from libs.datasets.timeseries import TimeseriesDataset
--- a/libs/qa/dataset_summary.py
+++ b/libs/qa/dataset_summary.py
@@ -90,7 +90,7 @@ def find_missing_values_in_summary(
 
 
 def get_summaries(
-    sha1: str, sha2: str, level: AggregationLevel = None, fips: str = None
+    sha1: str, sha2: str, level: Optional[AggregationLevel] = None, fips: Optional[str] = None
 ) -> Tuple[TimeseriesSummary, TimeseriesSummary]:
     """Builds summaries comparing timeseries between two commit shas.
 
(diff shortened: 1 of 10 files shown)


libs/qa/dataset_summary.py
```diff
@@ -0,0 +1,139 @@
...
+        sha=sha1, timeseries=timeseries1, summary=summary1, fips=fips, level=level
+    )
+    sum2 = TimeseriesSummary(
+        sha=sha2, timeseries=timeseries2, summary=summary2, fips=fips, level=level
+    )
+
+    return sum1, sum2
+
+
+def get_changes(df1: pd.DataFrame, df2: pd.DataFrame) -> pd.DataFrame:
```
TomGoBravo, line 129: what is expected to be in df1 and df2? the original df or summaries?

libs/qa/dataset_summary.py
```diff
@@ -0,0 +1,139 @@
...
+) -> pd.DataFrame:
+    """Returns values that are in the left summary but not in right summary."""
+    summary_l_value_index = summary_l[summary_l.has_value].index
+    summary_r_missing = summary_r[summary_r.has_value == False]
+    newly_missing_r = summary_r_missing[summary_r_missing.index.isin(summary_l_value_index)]
+    return summary_l.loc[newly_missing_r.index]
+
+
+def get_summaries(
+    sha1: str, sha2: str, level: AggregationLevel = None, fips: str = None
```
TomGoBravo, line 93: add Optional to level and fips types

libs/qa/dataset_summary.py
```diff
@@ -0,0 +1,139 @@
...
+        [CommonFields.FIPS, CommonFields.DATE, VARIABLE_FIELD]
+    )
+    fips_variable_grouped = melted.groupby([CommonFields.FIPS, VARIABLE_FIELD])
+    return fips_variable_grouped["value"].apply(generate_field_summary).unstack()
+
+
+def find_missing_values_in_summary(
+    summary_l: pd.DataFrame, summary_r: pd.DataFrame
+) -> pd.DataFrame:
+    """Returns values that are in the left summary but not in right summary."""
```
TomGoBravo, line 85: this looks like it depends on the two dataframes having the same index, or can you be sure there is a summary_r.has_value False for every input that doesn't have a value?

libs/qa/dataset_summary.py
ghop02: Aha! they changed it on me! 

https://google.github.io/styleguide/pyguide.html#3195-nonetype
> Use explicit Optional instead of implicit Optional. Earlier versions of PEP 484 allowed a: Text = None to be interpretted as a: Optional[Text] = None, but that is no longer the preferred behavior.

libs/qa/dataset_summary.py
ghop02: Summaries! but really just any two data frames with matching indexes

PR #568, opened 2020-07-23
Add combined datasets provenance

TomGoBravo:
This PR addresses https://trello.com/c/BdTfcCvM/301-build-a-data-provenance-report

It builds a `DataFrame` containing a string source name for each field-fips-date. It does not save the provenance report to disk or groupby field-fips (dropping the date because the sources are expected to be consistent).

* Remove BY_TIMESERIES_POINT, there are no plans to use it
* Extract the main loop of `_build_dataframe` into `_merge_data` and `_merge_data_by_row`

### Please Check
- [ ] Are tests passing?


libs/datasets/combined_datasets.py
```diff
@@ -298,14 +300,14 @@ class Override(Enum):
     BY_TIMESERIES = 2
     # For each <fips, variable, date>, use the highest priority datasource that has a real
     # (not NaN) value.
-    BY_TIMESERIES_POINT = 3
+    BY_TIMESERIES_POINT = 3  # Deprecated
 
 
 def _build_dataframe(
```
ghop02, line 306: Maybe `_build_dataframe_with_provenance` be more applicable now

TomGoBravo: i went with slightly shorter `_build_data_and_provenance`

TomGoBravo:
> This all looks good to me!
> 
> What do you see as the next steps for keeping provenance around?

I have another branch I started that adds provenance to the dataset object and will change how it is saved to also write the provenance.

PR #569, opened 2020-07-24
Fix run.sh

ghop02:
Didn't make the qa directory :/, this actually creates it so the copy will succeed


TomGoBravo:
LGTM, thank you for fixing it.
Managing state of disk is annoying. let's try to do less of it.

PR #571, opened 2020-07-24
Fix the Florida Negative Negative_Tests Edge Case

BrettBoval:
Some counties in Florida have erroneously reported test data that was not caught by our test correction filters. This tightens up filter restrictions as a first step towards a more robust pipeline.

https://trello.com/c/88PFNVhy/325-florida-infection-rate-wild-ride

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/pyseir/load_data.py
+++ b/pyseir/load_data.py
@@ -639,7 +639,8 @@ def load_new_test_data_by_fips(fips, t0, smoothing_tau=5, correction_threshold=5
     df = df.loc[
         (df[CommonFields.POSITIVE_TESTS].notnull())
         & (df[CommonFields.NEGATIVE_TESTS].notnull())
-        & ((df[CommonFields.POSITIVE_TESTS] + df[CommonFields.NEGATIVE_TESTS]) > 0),
+        & (df[CommonFields.NEGATIVE_TESTS] > 0)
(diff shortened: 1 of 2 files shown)


pyseir/load_data.py
```diff
@@ -639,7 +639,8 @@ def load_new_test_data_by_fips(fips, t0, smoothing_tau=5, correction_threshold=5
     df = df.loc[
         (df[CommonFields.POSITIVE_TESTS].notnull())
         & (df[CommonFields.NEGATIVE_TESTS].notnull())
-        & ((df[CommonFields.POSITIVE_TESTS] + df[CommonFields.NEGATIVE_TESTS]) > 0),
+        & (df[CommonFields.NEGATIVE_TESTS] > 0)
+        & (df[CommonFields.POSITIVE_TESTS] > 0),
```
mikelehen, line 643: FWIW, positive tests == 0 is normal in some counties that have contained covid (or haven't been hit yet).

But maybe it's okay not to do any test correction in that case.

BrettBoval: The naming convention is confusing. Here CommonFields.POSITIVE_TESTS is cumulative. This is a very permissive filter.

PR #572, opened 2020-07-24
Puerto rico WIP

eleanoravril:
Current issues are that the ICU bed count/capacity seems to be negative — trying to figure that out — and running generate api with the pr_output/web_ui input doesn't write files as expected. Files for MA are included in ma_output for comparison purposes.

### Please Check
- [ ] Are tests passing?


libs/datasets/latest_values_dataset.py
```diff
@@ -83,6 +87,34 @@ def build_from_data_source(cls, source):
 
         return cls.from_source(source, fill_missing_state=source.FILL_MISSING_STATE_LEVEL_DATA)
 
+    @classmethod
+    def _aggregate_puerto_rico_data(cls, data):
+        _log.info("aggregating puerto rico")
+        pr_data = data[data[CommonFields.FIPS].isin(custom_aggregations.ALL_PR_FIPS)]
```
ghop02, line 93: I think instead of defining all of the fips codes manually, you can just do something like the following:
```
is_pr_county = data[CommonFields.FIPS].str.match("72[0-9][0-9][0-9]")
pr_data = pr_data.loc[is_pr_county]
```

libs/datasets/latest_values_dataset.py
```diff
@@ -83,6 +87,34 @@ def build_from_data_source(cls, source):
...
+            ).sum() / licensed_beds.sum()
+            # should be updating here - in custom_aggregations file?
+            # data["72001"][CommonFields.MAX_BED_COUNT] = weighted_all_bed_occupancy
+        if CommonFields.ICU_BEDS and CommonFields.ICU_TYPICAL_OCCUPANCY_RATE in pr_data.columns:
+            icu_beds = pr_data[CommonFields.ICU_BEDS]
+            occupancy_rates = pr_data[CommonFields.ICU_TYPICAL_OCCUPANCY_RATE]
+            weighted_icu_occupancy = (licensed_beds * occupancy_rates).sum() / licensed_beds.sum()
+            # should be updating here - in custom_aggregations file?
+            # data["72001"][CommonFields.ICU_BEDS] = 3333
+
```
ghop02, line 115:
```
is_pr_state = data[CommonFields.FIPS] == "72"
data.loc[is_pr_state, ComonFields. ALL_BED_TYPICAL_OCCUPANCY_RATE] = weighted_all_bed_occupancy
data.loc[is_pr_state, ComonFields. ICU_TYPICAL_OCCUPANCY_RATE] = weighted_icu_occupancy
```

Should do the trick at the end!

ghop02: Or rather, you could put those updates in the if statements above

libs/datasets/sources/fips_population.csv
```diff
@@ -3141,81 +3141,81 @@ WY,Teton County,56039,23464.0
...
-PR,Toa Baja Municipio,72137,
-PR,Trujillo Alto Municipio,72139,
-PR,Utuado Municipio,72141,
-PR,Vega Alta Municipio,72143,
-PR,Vega Baja Municipio,72145,
-PR,Vieques Municipio,72147,
-PR,Villalba Municipio,72149,
-PR,Yabucoa Municipio,72151,
-PR,Yauco Municipio,72153,
+PR,Adjuntas Municipio,72001,17363.0
```
ghop02, line 3144: Can you add this to https://github.com/covid-projections/covid-data-public/blob/master/data/misc/fips_population.csv as well?

ghop02: We should probably remove this file and just rely on covid-data-public as well, but no need to do that in this PR

libs/datasets/beds.py
```diff
@@ -102,6 +103,46 @@ def from_source(cls, source: "DataSource", fill_missing_state=True):
 
         return cls(data)
 
+    @classmethod
+    def _aggregate_puerto_rico_data(cls, data):
```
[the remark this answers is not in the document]
eleanoravril: Ah, yup, forgot to remove — it isn't used from what I can tell.

libs/datasets/latest_values_dataset.py
```diff
@@ -83,6 +85,34 @@ def build_from_data_source(cls, source):
 
         return cls.from_source(source, fill_missing_state=source.FILL_MISSING_STATE_LEVEL_DATA)
 
+    @classmethod
+    def _aggregate_puerto_rico_data(cls, data):
```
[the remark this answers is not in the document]
eleanoravril: Will do!

libs/datasets/sources/fips_population.csv
eleanoravril: Ok, will add there!

pyseir/load_data.py
```diff
@@ -594,6 +594,8 @@ def load_whitelist():
     path = pyseir.utils.get_run_artifact_path(
         fips=PLACEHOLDER_FIPS, artifact=pyseir.utils.RunArtifact.WHITELIST_RESULT
     )
+    log.info("path")
+    log.info(path)
```
ghop02, line 598: remove this log

PR #577, opened 2020-07-27
Add autosigma notebook

BrettBoval:

### Please Check
- [ ] Are tests passing?


BrettBoval:
@ghop02 let's discuss if we want things like this to have binder support. 

Also, I had to bump some requirements to make the jupyter lab interactive widgets work. I think they are significant improvements and we should support them, but haven't done it yet. Finally time for Poetry?

BrettBoval:
Specifically I know there is a mismatch between the newest jupyter lab and our currently pinned version of ipympl.

PR #579, opened 2020-07-28
Customizing Binder Environment + Adding a generic dataset spelunking notebook

ghop02:
I ended up merging in the cds data branch into this for some local testing, but otherwise wanted to get this out there.

A couple changes here: 
 * Added a folder `binder/` which contains files necessary to creating a working binder environment
 * Added a notebook `notebooks/Data Source Comparison.ipynb` which allows for some basic data spelunking among different datasets
 * Removed the environment.yaml file for conda installations as the requirements were quite out of date. 

Thoughts:
 * having the separate binder folder is helpful for containing the binder environment files - however, it means that requirements.txt is duplicated.  There are a couple solutions that I could think of to this problem (unfortunately symlinking didn't seem to work):
   1. Add a comment in the root requirements.txt file to copy on any changes
   2. One of the reasons why I originally did this was because we install the covid data public package via a link to a local copy.  I ended up changing the covid-data-public to point to git rather than locally.   Now that requirements.txt actually works without having covid-data-public installed, I could move all the files in `binder/` to the root of the directory.  This seems okay, but it's a bit messier.

I still don't have the workflow file for building binder automatically.  My thinking is that I would like it to just run on master commits - its a bit heavy for running on every push imo


ghop02:
https://hub.gke.mybinder.org/user/covid-projectio-ovid-data-model-iwid8o4z/notebooks/notebooks/Data%20Source%20Comparison.ipynb

Here's the notebook if anyone wants to play with it

notebooks/Data Source Comparison.ipynb
```diff
@@ -0,0 +1,329 @@
+{
```
TomGoBravo, line 1: I see a lot of familiar code in here. Did you use a tool to help you build this file?

ghop02: Oh this is a jupyter notebook! I'd like to move more this code out of the notebook - planning on doing some of that when I write the next notebook.

PR #580, opened 2020-07-29
Add Patch to Apply Rt Aggregate Levels for Arbitrary Fips Collections

BrettBoval:
This PR addresses New Orleans request for collections of county fips to present a single infer rt value. 

### Please Check
- [ ] Are tests passing?


pyseir/rt/patches.py
```diff
@@ -0,0 +1,46 @@
+import pandas as pd
+import numpy as np
+
+from libs.datasets.combined_datasets import load_us_latest_dataset
+import pyseir.utils
+
+population_dict = load_us_latest_dataset().data.set_index("fips")["population"].to_dict()
```
ghop02, line 7: I'd prefer not to load this here - in general I like to keep imports real lightweight.  when you're reading from disk during a file import, I find it make it more difficult to test + slows down initialization. 

couple options:
 - just run this code in the `patch aggregate rt results` function.  I don't think it should be too expensive.
 - Just call `population = combined_datasets.get_us_latest_for_fips(fips)[CommonFields.POPULATION]` inside of load_and_append_population - should also be speedy

pyseir/rt/patches.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def f(x):
+        """Helper function to apply to weighted arithmetic means"""
+        d = dict()
+        d["Rt_MAP_composite"] = np.average(x["Rt_MAP_composite"], weights=x["population"])
+
+        # TODO: Decide whether these errors should be changed to being combined in quadrature
+        # http://ipl.physics.harvard.edu/wp-uploads/2013/03/PS3_Error_Propagation_sp13.pdf instead
+        # of the population weighted arithmetic mean
+        d["Rt_ci95_composite"] = np.average(x["Rt_ci95_composite"], weights=x["population"])
```
ghop02, line 42:
```
weighted_data = {
  'Rt_MAP_composite': ...,
  # comment
  'Rt_ci95_composite': ...
}
```
may be a bit more concise

pyseir/rt/utils.py
```diff
@@ -234,3 +243,27 @@ def align_time_series(series_a, series_b):
...
+    ----------
+    fips: str
+        State or County FIPS code.
+
+    Returns
+    -------
+    results: pd.DataFrame
+        DataFrame containing the R_t inferences.
+    """
+    if fips in NEW_ORLEANS_FIPS:
```
ghop02, line 262: I think it would be helpful to add an easy toggle for this if possible so that we can turn on/off NO fips easily.

possible an environment variable?
maybe
```
DISABLE_NEW_ORLEANS_COMBINED_RT = os.getenv('DISABLE_NEW_ORLEANS_COMBINED_RT', False)
```

then you could do:
```
if not DISABLE_NEW_ORLEANS_COMBINED_RT and fips in NEW_ORLEANS_FIPS:
```

pyseir/rt/patches.py
BrettBoval: I'm going to leave it as is. The separate function logic is to apply it to a groupby object that depends on multiple columns of a grouped object. I agree that it's a little hard to reason about.

pyseir/rt/utils.py
BrettBoval: We should decide how many of these aggregations we are going to do. It's not clear to me if we should have for every bespoke aggregation we do, or if a single bespoke toggle is enough. I'll punt on this decision for this PR.

pyseir/rt/patches.py
BrettBoval: Fair point -> implemented the second one.

pyseir/rt/patches.py
BrettBoval: But I'll change the tmp dict creation. I still need the f(x) -> pd.Series for each groupby object.

PR #582, opened 2020-07-29
Adding in logic to remove CT case backfill

ghop02:
Removes CT case backfill from the NYTimes dataset.  

Cases are removed weighted on number of cases that day, see: https://trello.com/c/rPiM6UF4/333-adjust-ct-case-data-based-on-reported-backfill-vs-current-cases

### DIFF
--- a/libs/datasets/sources/nytimes_dataset.py
+++ b/libs/datasets/sources/nytimes_dataset.py
@@ -23,6 +23,17 @@ ESTIMATED_REMOVED_CT_CASES_COUNTY = {
 }
 
 
+def _remove_ct_backfill_cases(data: pd.DataFrame) -> pd.DataFrame:
+    # July 24th was the day there was a backfill of cases
+    is_on_or_after_july_24 = data[CommonFields.DATE] >= "2020-07-24"
+
+    # Remove backfilled tests from cumulatives on and after July 24th.
+    for fips, count in ESTIMATED_REMOVED_CT_CASES_COUNTY.items():
+        is_fips_data_after_jul_24 = is_on_or_after_july_24 & (data[CommonFields.FIPS] == fips)
+        data.loc[is_fips_data_after_jul_24, CommonFields.CASES] -= count
+    return data
+
+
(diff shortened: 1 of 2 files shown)


libs/datasets/sources/nytimes_dataset.py
```diff
@@ -48,6 +63,22 @@ def local(cls) -> "NYTimesDataset":
         data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
         return cls(data_root / cls.DATA_FOLDER / cls.COUNTIES_DATA_FILE)
 
+    @classmethod
```
TomGoBravo, line 66: how about a plain function instead of a cls method? seems simplier.

libs/datasets/sources/nytimes_dataset.py
```diff
@@ -48,6 +63,22 @@ def local(cls) -> "NYTimesDataset":
...
+        is_county = data[cls.Fields.AGGREGATE_LEVEL] == "county"
+        is_ct = data[cls.Fields.STATE] == "CT"
+        # July 24th was the day there was a backfill of cases
+        is_on_or_after_july_24 = data[cls.Fields.DATE] >= "2020-07-24"
+        is_ct_county_data_after_jul_24 = is_county & is_ct & is_on_or_after_july_24
+
+        # Remove backfilled tests from cumulatives on and after July 24th.
+        for fips, count in ESTIMATED_REMOVED_CT_CASES_COUNTY.items():
+            is_ct_fips_data_after_jul_24 = is_ct_county_data_after_jul_24 & (
+                data[CommonFields.FIPS] == fips
```
TomGoBravo, line 77: how about matching on FIPS and DATE only? I'd like to reduce our dependency on other fields so we can more easily support other region types. I like project Li locationID.

ghop02: Oh do you mean filtering out the ct filter? that makes sense to me

TomGoBravo: Yes, you got it.

PR #583, opened 2020-07-29
adding notebook helpers + more flexibility around covid_data_public variable + Compare Snapshot notebook

ghop02:


BrettBoval:
Can we enforce a no_spaces_in_notebook_names rule please?

libs/datasets/dataset_utils.py
```diff
@@ -9,12 +9,17 @@
...
-if os.getenv("COVID_DATA_PUBLIC"):
-    LOCAL_PUBLIC_DATA_PATH = pathlib.Path(os.getenv("COVID_DATA_PUBLIC"))
-else:
-    LOCAL_PUBLIC_DATA_PATH = (
-        pathlib.Path(__file__).parent.parent / ".." / ".." / "covid-data-public"
-    )
+
+def _get_public_data_path():
+    """Sets global path to covid-data-public directory."""
+    if os.getenv("COVID_DATA_PUBLIC"):
```
BrettBoval, line 15: Is there a list of environmental variables that we attempt to access? I didn't realize that we had this as an option and have been doing some hacky jupyter lab %cd stuff to make it explicit in the notebook what working directory we need.

ghop02: We don't have a comprehensive place yet.  I would really like to use something like [pydantic settings](https://pydantic-docs.helpmanual.io/usage/settings/) eventually to keep all settings easily in one spot

ghop02: Yeah that's definitely an option.  

I have found that those end up getting out of date (definitely preventable) and having a single source of truth for constants is helpful (even if you're not using environment variables)

PR #587, opened 2020-07-30
Make backfilling easier + remove new ct backfilled cases

ghop02:


libs/datasets/sources/nytimes_dataset.py
```diff
@@ -1,36 +1,72 @@
...
+
+    Args:
+        data: Input Data.
+        date: Date of backfill.
+        backfilled_cases: Number of backfilled cases.
+        state_fips: FIPS code for state.
+
+    Returns: Dictionary of estimated fips -> backfilled cases.
+    """
+    is_county_level = data[CommonFields.AGGREGATE_LEVEL] == AggregationLevel.COUNTY
```
TomGoBravo, line 37: rm unused

libs/datasets/sources/nytimes_dataset.py
```diff
@@ -1,36 +1,72 @@
...
+
+    fields = [CommonFields.DATE, CommonFields.FIPS, CommonFields.CASES]
+    cases = (
+        data.loc[is_state & is_not_unknown, fields]
+        .set_index([CommonFields.FIPS, CommonFields.DATE])
+        .sort_index()
+    )
+    cases = cases.diff().reset_index(level=1)
+    cases_on_date = cases[cases.date == date]["cases"]
+    return (cases_on_date / cases_on_date.sum() * backfilled_cases).round().to_dict()
```
TomGoBravo, line 49: do you think it is possible for the new sum of cases in the counties to be more than 1 from the old sum of cases - backfilled if the rounding gets unlucky and a bunch of counties have a new case count that gets rounded in the same direction? Don't need to fix but maybe add a comment if it seems possible to you.

ghop02: It's definitely possible (especially for a state with more counties).  I'll add a comment!

PR #591, opened 2020-07-31
Remove nytimes test + simplify nytimes class

ghop02:
Uses the timeseries-common file now in covid-data-public


ghop02:
So I double checked again, and it looks like there are a couple small changes to a couple days of state data in TX + a small change in NJ.  I feel fine merging it and kicking off a build + maybe raising the issue with NYTimes

PR #593, opened 2020-08-02
Add wide dates timeseries output

TomGoBravo:
##  This PR

* Adds a new csv output that has a column for each date. This file is 1/5 the size of the current `timeseries.csv` and includes the provenance and summary data.
* Reverts back to sorted timeseries-summary.csv, as disabled by https://github.com/covid-projections/covid-data-model/pull/606 because we want the file sorted for easier diffing.

I'm not totally sure how I'm going to use `timeseries-wide-dates.csv` :-/ but want to try having it on-hand for debugging data problems.

### Please Check
- [ ] Are tests passing?


libs/datasets/timeseries.py
```diff
@@ -96,6 +97,51 @@ def latest_values(self) -> pd.DataFrame:
             .reset_index(drop=True)
         )
 
+    def get_date_columns(self) -> pd.DataFrame:
```
ghop02, line 100: Can you add a docstring to this?

libs/datasets/timeseries.py
```diff
@@ -12,6 +12,7 @@
 from libs.datasets.common_fields import CommonIndexFields
 from libs.datasets.common_fields import CommonFields
 from libs.datasets.dataset_utils import AggregationLevel
+from libs.qa.dataset_summary_gen import generate_field_summary
```
ghop02, line 15: would prefer to import the module ;)

libs/datasets/timeseries.py
TomGoBravo: done

libs/datasets/timeseries.py
TomGoBravo: done

PR #594, opened 2020-08-03
Pyseir location cleanup + start of removing county_metadata

ghop02:
A couple things this PR accomplishes:
 * removes all state level data access functions in load_data - everything is accessed through fips.
 * Start removing most of the county_metadata references.  County metadata is another source that goes outside of our traditional data pipeline.  Removing the references helps streamline some of the pyseir code.  That source has always been very opaque to me.
 * One of the confusing things is that we use the full state name in a lot of locations when we should be using fips or the ISO 2 abbreviated name.  I start to try to make it more explicit where we do that in this PR.  I'd like to take another pass through in a separate PR to finish that cleanup.
 * It seems we only really use county metadata in the ensemble_runner for age distribution.  Do we actually use that? it would be great if we could find another way to save that data.

### DIFF
--- a/pyseir/inference/initial_conditions_fitter.py
+++ b/pyseir/inference/initial_conditions_fitter.py
@@ -229,6 +229,7 @@ def _fit_fips(fips, generate_report=False):
         return {"model_params": None, "t0_date": None, "reduced_chi2": None}
 
 
+# Chris: Can we delete??? Not used anywhere and I don't think start dates matter...
 def generate_start_times_for_state(state, generate_report=False):
     """
     Generate imputed start dates for each county.
--- a/libs/datasets/__init__.py
+++ b/libs/datasets/__init__.py
@@ -11,3 +11,4 @@ from libs.datasets.sources.fips_population import FIPSPopulation
 from libs.datasets.sources.nha_hospitalization import NevadaHospitalAssociationData
 
 from libs.datasets.common_fields import CommonFields
+from libs.datasets.dataset_utils import AggregationLevel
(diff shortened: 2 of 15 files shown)


pyseir/inference/initial_conditions_fitter.py
```diff
@@ -229,6 +229,7 @@ def _fit_fips(fips, generate_report=False):
         return {"model_params": None, "t0_date": None, "reduced_chi2": None}
 
 
+# Chris: Can we delete??? Not used anywhere and I don't think start dates matter...
```
BrettBoval, line 232: Remove this whole file (InitialConditionsFitter class) and the cli calls associated with this. It isn't used anymore.

BrettBoval:
> * It seems we only really use county metadata in the ensemble_runner for age distribution.  Do we actually use that? it would be great if we could find another way to save that data

Which line are you referencing with this?

ghop02:
> > * It seems we only really use county metadata in the ensemble_runner for age distribution.  Do we actually use that? it would be great if we could find another way to save that data
> 
> Which line are you referencing with this?

https://github.com/covid-projections/covid-data-model/blob/master/pyseir/ensembles/ensemble_runner.py#L247-L254

ghop02:
@BrettBoval I'm going to leave deleting the other code for another PR if that's alright!

BrettBoval:
> > > * It seems we only really use county metadata in the ensemble_runner for age distribution.  Do we actually use that? it would be great if we could find another way to save that data
> > 
> > 
> > Which line are you referencing with this?
> 
> https://github.com/covid-projections/covid-data-model/blob/master/pyseir/ensembles/ensemble_runner.py#L247-L254

I don't know of anywhere where we use that data.

BrettBoval:
> @BrettBoval I'm going to leave deleting the other code for another PR if that's alright!

That's fine with me.

PR #600, opened 2020-08-06
Rename cmdc -> covid county data

ghop02:
Uses the new covid-county-data files saved in covid-data-public

### DIFF
--- a/libs/datasets/sources/covid_county_data.py
+++ b/libs/datasets/sources/covid_county_data.py
@@ -1,6 +1,7 @@
 import pandas as pd
 
 from covidactnow.datapublic.common_fields import CommonFields
+from covidactnow.datapublic import common_df
 from libs.datasets import data_source
 from libs.datasets import dataset_utils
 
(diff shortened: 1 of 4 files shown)


libs/datasets/sources/covid_county_data.py
```diff
@@ -1,15 +1,16 @@
...
 from libs.datasets import data_source
 from libs.datasets import dataset_utils
 
 
-class CmdcDataSource(data_source.DataSource):
-    DATA_PATH = "data/cases-cmdc/timeseries-common.csv"
-    SOURCE_NAME = "CMDC"
+class CovidCountyDataDataSource(data_source.DataSource):
+    DATA_PATH = "data/cases-covid-county-data/timeseries-common.csv"
+    SOURCE_NAME = "CovidCountyData (Valorum)"
```
TomGoBravo, line 11: CCD or Valorum. So far SOURCE_NAME has always looked more like a short key containing only a-zA-Z_ which are easily recognizable by people on the team but not really fit for external display. Lets stick with that convention. If we need source names for easy external use lets do it for all sources at once.

PR #601, opened 2020-08-06
Refactor for wide dates csv

TomGoBravo:
### This PR
* Changes `update_data_public_head` to always get data as arguments, making dependency explicit. Also I want `cli.data.update` to have the data so it can write the wide date csv in my next PR.
* Factor out `fips_index_geo_data` so it can be reused
* in `timeseries_test.py` give every row a unique FIPS value
* move `generate_field_summary` so it can be imported by `TimeseriesDataset` without creating an import loop.

### Please Check
- [ ] Are tests passing?


cli/data.py
```diff
@@ -34,10 +34,16 @@ def _save_field_summary(timeseries_dataset: TimeseriesDataset, output_path: path
...
 
-    dataset = timeseries_pointer.load_dataset()
-    _save_field_summary(dataset, path_prefix / summary_filename)
+    latest_dataset = combined_datasets.build_us_latest_with_all_fields()
+    timeseries_dataset = combined_datasets.build_us_timeseries_with_all_fields()
+    _, timeseries_pointer = combined_dataset_utils.update_data_public_head(
+        path_prefix, latest_dataset, timeseries_dataset
+    )
+
+    if summary_filename:
```
ghop02, line 44: There will always be a summary filename, correct?

libs/datasets/combined_datasets.py
```diff
@@ -293,23 +289,7 @@ def _build_data_and_provenance(
...
-    all_identifiers = pd.concat(
-        df.reset_index().loc[
-            :, [CommonFields.FIPS] + list(df.columns.intersection(preserve_columns))
-        ]
-        for df in datasource_dataframes.values()
-    ).drop_duplicates()
-    # Make a DataFrame with a unique FIPS index. If multiple rows are found with the same FIPS then there
-    # are rows in the input data sources that have different values for county name, state etc.
-    fips_indexed = all_identifiers.set_index(CommonFields.FIPS, verify_integrity=True)
+    fips_indexed = fips_index_geo_data(pd.concat(datasource_dataframes.values()))
```
ghop02, line 292: Can you call this `dataset_utils.fips_index_geo_data`?

cli/data.py
TomGoBravo: Not with `./run.py data update --summary-filename=''`. I wanted a way to disable it to see if the wide-dates output is faster.

libs/datasets/combined_datasets.py
TomGoBravo: Done in 5c7c0f0

PR #603, opened 2020-08-10
Top level metrics

benholtzmansmith:
Adds top level metrics for case density and test positivity

### Please Check
- [ ] Are tests passing?


libs/functions/generate_api.py
```diff
@@ -156,9 +160,19 @@ def generate_region_timeseries(
             for row in model_output.data.to_dict(orient="records")
         ]
 
+    metrics_timeseries = []
+    region_level_metrics = calculate_top_level_metrics_for_timeseries(
```
ghop02, line 164: Can you pass this in as an argument (moving the calculation to `api_pipeline.py` i believe? I'd like to keep most of the burden of calculations out of the `generate_api` code.

api/can_api_definition.py
```diff
@@ -59,6 +61,20 @@ class ResourceUtilization(base_model.APIBaseModel):
     )
 
 
+class Metrics(base_model.APIBaseModel):
+    testPositivity: Optional[int] = pydantic.Field(
+        ..., description="The percentage of people who test positive."
```
ghop02, line 66: This is stored as a ratio not a percent, right?

libs/top_level_metrics.py
```diff
@@ -0,0 +1,115 @@
...
+
+    pos_tests_cumulative = timeseries.data[CommonFields.POSITIVE_TESTS]
+    pos_tests_daily = pos_tests_cumulative.diff()
+
+    case_density = calculate_case_density(cases=pos_tests_daily, population=population)
+    test_positivity = calculate_test_positivity(
+        pos_cases=pos_tests_daily, neg_tests=neg_tests_daily
+    )
+
+    return pd.DataFrame({"caseDensity": case_density, "testPositivity": test_positivity}).replace(
```
ghop02, line 36: Can you include the date index with this result? Also, I generally try to include fips in any output data frames as well, makes it much easier to join and combine multiple regions if desired

api/can_api_definition.py
```diff
@@ -59,6 +61,20 @@ class ResourceUtilization(base_model.APIBaseModel):
     )
 
 
+class Metrics(base_model.APIBaseModel):
```
ghop02, line 64: Can you add a class level docstring? If you pull in master, there are a few more examples.  The docstring ends up showing up in [the documentation](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md#resourceusageprojection)

api/can_api_definition.py
ghop02: Should we note that this is a 7d rolling average*

libs/functions/generate_api.py
benholtzmansmith: That seems a little weird to me given that here is where we also calculate `actuals_timeseries` which gets returned in `RegionSummaryWithTimeseries`. Makes sense to me have those two values evaluated in the same place.

ghop02: Ah, the actuals calculations does do that one calculation, but that is more of a relic that we will hopefully remove (that field has been deprecated for a while now).  

My thinking is that i'd generally like this file to be simple mappings to help make testing easier/separate how the data is persisted from how we're generating the data.

For example, I could see wanting to perform the metrics calculations in bulk to speed it up.  If we pass in just the metrics df for one fips here as an argument, I think it gives us more flexibility to do that sort of change outside of this code.

libs/functions/generate_api.py
```diff
@@ -156,9 +160,19 @@ def generate_region_timeseries(
...
+        timeseries, region_summary
+    ).to_dict(orient="records")
+    for metric_row in region_level_metrics:
+        MetricsTimeseriesRow(**metric_row, date=row[CommonFields.DATE])
```
ghop02, line 168: don't think this is actually getting added to the timeseries :) 

I think it would be helpful to add a quick test making sure that the metrics data is being added to the output - 
[here](https://github.com/covid-projections/covid-data-model/blob/master/test/libs/generate_api_test.py) is where the existing tests are

benholtzmansmith: Doh, I think it got removed in a previous change. I will that test.

libs/top_level_metrics.py
```diff
@@ -0,0 +1,129 @@
...
+    pos_tests_daily = pos_tests_cumulative.diff()
+    cases_daily = cases_cumulative.diff()
+
+    case_density = calculate_case_density(cases=cases_daily, population=population)
+    test_positivity = calculate_test_positivity(
+        pos_cases=pos_tests_daily, neg_tests=neg_tests_daily
+    )
+
+    top_level_metrics_data = {
+        "caseDensity": case_density,
```
ghop02, line 41: I think it would be good to make these constants.  It's a bit of a trick that we're using to say "the key is the same name as the output of the API definition".  Definitely possible that we could rename it.  One suggestion I have is to make a simple class (or enum, similar to [CommonFields](https://github.com/covid-projections/covid-data-public/blob/master/covidactnow/datapublic/common_fields.py#L25)) that has the names of the metrics columns as keys.  

that way this could be `MetricsFields.CASE_DENSITY` and this is a bit less tightly coupled to the API output name

libs/top_level_metrics.py
```diff
@@ -0,0 +1,129 @@
...
+# to get all fips codes for that state
+def calculate_metrics_for_counties_in_state(state: str):
+    latest = combined_datasets.load_us_latest_dataset()
+    state_latest_values = latest.county.get_subset(state=state)
+    for fips in state_latest_values.all_fips:
+        yield calculate_top_level_metrics_for_fips(fips)
+
+
+def smooth_with_rolling_average(
+    series: pd.Series, window: int = 7, includeTrailingZeros: bool = True
```
ghop02, line 103: `include_trailing_zeros`

libs/top_level_metrics.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def calculate_case_density(
+    cases: pd.Series, population: int, smooth: int = 7, normalize_by: int = 100000
+) -> pd.Series:
+    """
+    Calculates normalized cases density.
+
+    Args:
+        cases: Number of cases in a given fips.
```
ghop02, line 57: Number of daily cases? (probably good to note since cases are often cumulative)

libs/pipelines/api_pipeline.py
```diff
@@ -67,10 +72,34 @@ def run_on_all_fips_for_intervention(
...
+    else:
+        latest = None
+    return metrics_for_fips, latest
+
+
 def build_timeseries_for_fips(
     intervention, us_latest, us_timeseries, model_output_dir, fips
 ) -> Optional[RegionSummaryWithTimeseries]:
     fips_latest = us_latest.get_record_for_fips(fips)
+    metrics_timeseries, metrics_latest = generate_metrics_and_latest_for_fips(fips)
```
ghop02, line 102: since you have the timeseries and latest datasets for a given fips in this function, I might recommend calling the version of the function `calculate_top_level_metrics_for_timeseries` so that you're not filtering by fips twice.

PR #604, opened 2020-08-11
Update API readme documentation

ghop02:
Updates [readme](https://github.com/covid-projections/covid-data-model/blob/update-documentation/api/README.V1.md) to include api schemas in the output.

Most of the logic is in libs/update_readme_schemas.py.  It's definitely very hacky, but parses the JSON Schema and generates sensible readme output.

I generate the readme from a template replacing a specific template key (`###api_descriptions###`).


ghop02:
I ended up going with making the subheadings the same title as that in the table.  Down to make more changes in the future

PR #609, opened 2020-08-12
Share loaded data sources between latest and timeseries combined data

TomGoBravo:
## This PR

* Moves code that changes a `DataSource` into a `DatasetBase` from combined_datasets, latest_values_dataset.py, timeseries.py to data_source.py.
* Load data sources before calling _build_combined_dataset_from_sources so that the loaded values can be used in multiple calls. Because `DataSource.latest` sometimes calls `DataSource.timeseries` the cache saves a bunch of processing time, reducing time as measured by `/usr/bin/time -v ./run.py data update` from 6 to 5 minutes.
* Make sure geo data columns in the wide-data DataFrame are created in a deterministic manner and sorted to match existing data.

## Tested

`./run.py data update` re-generates the existing contents of `data/` exactly.

### DIFF
--- a/libs/datasets/data_source.py
+++ b/libs/datasets/data_source.py
@@ -69,8 +69,22 @@ class DataSource(object):
 
     @lru_cache(None)
     def timeseries(self) -> TimeseriesDataset:
-        return TimeseriesDataset.build_from_data_source(self)
+        """Build TimeseriesDataset from this data source."""
+        if set(self.INDEX_FIELD_MAP.keys()) != set(TimeseriesDataset.INDEX_FIELDS):
+            raise ValueError("Index fields must match")
+
+        return TimeseriesDataset.from_source(
+            self, fill_missing_state=self.FILL_MISSING_STATE_LEVEL_DATA
+        )
 
     @lru_cache(None)
     def latest_values(self) -> LatestValuesDataset:
-        return LatestValuesDataset.build_from_data_source(self)
+        if set(self.INDEX_FIELD_MAP.keys()) == set(TimeseriesDataset.INDEX_FIELDS):
+            return LatestValuesDataset(self.timeseries().latest_values())
+
+        if set(self.INDEX_FIELD_MAP.keys()) != set(LatestValuesDataset.INDEX_FIELDS):
+            raise ValueError("Index fields must match")
+
+        return LatestValuesDataset.from_source(
+            self, fill_missing_state=self.FILL_MISSING_STATE_LEVEL_DATA
+        )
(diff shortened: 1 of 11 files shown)


libs/datasets/data_source.py
```diff
@@ -60,14 +60,31 @@ def local(cls) -> "DataSource":
...
-        return TimeseriesDataset.build_from_data_source(self)
+        """Build TimeseriesDataset from this data source."""
+        if set(self.INDEX_FIELD_MAP.keys()) != set(TimeseriesDataset.INDEX_FIELDS):
+            raise ValueError("Index fields must match")
+
+        return TimeseriesDataset.from_source(
+            self, fill_missing_state=self.FILL_MISSING_STATE_LEVEL_DATA
+        )
+
+    @lru_cache(None)
```
ghop02, line 80: I guess it doesn't matter too much, but I'd lean towards dropping the lru_cache here. We tried to rely on it more when it was trickier to query certain sets of data, but now that's not as much of an issue

libs/datasets/dataset_base.py
```diff
@@ -48,3 +43,14 @@ def to_csv(self, path: pathlib.Path):
         if self.provenance is not None:
             provenance_path = str(path).replace(".csv", "-provenance.csv")
             self.provenance.sort_index().to_csv(provenance_path)
+
+    def indexed_df(self) -> pd.DataFrame:
```
ghop02, line 47: Could you add a docstring here? 

Also, I might propose making this a property:
```
@property
def indexed_data(self):
    return ...

```

Might give a bit more symmetry accessing: 
`self.data` and `self.indexed_data`

libs/datasets/data_source.py
TomGoBravo: just on latest_values or timeseries too? I'm guessing it makes a big difference for timeseries and I like the consistency.

libs/datasets/dataset_base.py
TomGoBravo: added docstring
Symmetry is good but I also like making cheap reads simple object attributes and more expensive ones methods.

libs/datasets/data_source.py
TomGoBravo: With both LRU removed run time increases from 300 -> 384 seconds.

libs/datasets/dataset_base.py
ghop02: how about `indexed_data` rather than `indexed_df`? We don't really refer to df anywhere - so data feels a bit more consistent

TomGoBravo: Done in 502a9ec

PR #616, opened 2020-08-15
synthesize testing data for valorum

TomGoBravo:
## This PR

https://trello.com/c/mIImJgh5/282-consume-plumb-test-positivity-data-for-louisiana-counties-from-valorum

* Adds `CovidCountyDataDataSource.synthesize_test_metrics` that generates testing data where they can be calculated from other metrics and keeps method used to generate it in provenance.
* Adds optional `DataSource.provenance`, with same structure as `TimeseriesDataset.provenance`
* Any DataSource with provenance (currently only Valorum/CovidCountyDataDataSource) is written to disk in a wide-dates.csv
* Factors out `wide_dates_df.write_csv`
* Extends `run.py utils csv-diff` to support the left side being a git rev, ie `run.py utils csv-diff master data/timeseries.csv`
* Sorts csv-diff output of timeseries by amount of diff because nobody cares about metrics that haven't changed.
* Changes ` _to_timeseries_rows` (turns a provenance DataFrame with a string for every date to a string per variable) to join all values with `;` instead of keeping only the first. And renamed it to `provenance_wide_metrics_to_series`
* Changes the contents of `build_us_with_all_fields` to be inlined at the one calling site.

## Tested

I ran `./run.py data update` and then `./run.py utils csv-diff master data/timeseries.csv` to find out how the combined data changed. There are 574 FIPS that gained a negative_tests metric. negative_tests and positive_tests changed for lots of FIPS. Here are the top 20:
```
                          diff  points_overlap  has_overlap
variable       fips                                        
negative_tests 36061  1.000000             205         True
positive_tests 48431  0.714286              14         True
               17031  0.408864             151         True
negative_tests 17031  0.374730              37         True
positive_tests 48451  0.207374              14         True
               48135  0.189428              14         True
               48093  0.141138              14         True
               48355  0.126038              14         True
               48149  0.104994              14         True
               48465  0.094098              14         True
               01121  0.058504             157         True
               01009  0.045378             155         True
               01039  0.045270             154         True
               48449  0.041463              14         True
               17091  0.036001             151         True
               01099  0.033007             149         True
               01019  0.029950             156         True
               01053  0.026612             152         True
               01115  0.024237             150         True
               01057  0.023739             150         True
```

36061 is NYC! Looking at `grep -P -H '\A36061,negative_tests' data-af258b1/timeseries-wide-dates.csv data/timeseries-wide-dates.csv` where af258b1 is master we had all `0` for negative_tests from Valorum. I don't know why it was 0 and looked at the covid-data-public repo `csvgrep -c fips -r '(36047|36081|36005|36085|36061)' data/cases-covid-county-data/timeseries-common.csv | csvcut -c fips,date,county,cases,deaths,recovered,positive_tests,negative_tests,total_tests | csvlook |less` but all the FIPS appear to be blank for negative_tests. This PR turns it into something that looks realistic.

48431 is a small county in rural Texas, which had positive_tests from CDS that was 0 for months, went up to 1 and back to 0 and back to 1. This PR changes to Valorum data that looks better.

17031 is Cook County / Chicago. Raised issue in `CAN // Valorum Scraping Tracker` google sheet.

48451 is Tom Green County, TX. Raised issue in `CAN // Valorum Scraping Tracker` google sheet.
Same for 48135 Ector County, TX


libs/qa/common_df_diff.py
```diff
@@ -82,8 +82,8 @@ def __str__(self):
 TS only in this file: {self.my_ts}
 TS points only in this file: {self.my_ts_points.groupby('date').size().to_dict()}
 """ + (
-            f"""TS diffs:\n{self.ts_diffs.sort_values('diff', ascending=False)}
-TS diffs by variable and has_overlap:\n{self.ts_diffs.groupby('variable has_overlap'.split()).mean()}
+            f"""TS diffs:\n{self.ts_diffs.sort_values('diff', ascending=False).head(20)}
+TS diffs by variable and has_overlap:\n{self.ts_diffs.groupby('variable has_overlap'.split()).mean().sort_values('diff', ascending=False).head(20)}
```
ghop02, line 86: Seems like it might be worth pulling this calculation out from the template string and into a variable

cli/utils.py
```diff
@@ -2,17 +2,20 @@
...
 
 import click
+import git
 import structlog
 
 from covidactnow.datapublic import common_df
 from covidactnow.datapublic.common_fields import CommonFields
 from libs import github_utils
-from libs.datasets import combined_datasets
+from libs.datasets import combined_datasets, dataset_utils
```
ghop02, line 14: Can you add this import on a separate line? recommended practice by the [gsg](https://google.github.io/styleguide/pyguide.html#313-imports-formatting)

libs/datasets/sources/covid_county_data.py
```diff
@@ -40,10 +44,88 @@ class CovidCountyDataDataSource(data_source.DataSource):
...
+            "negative_tests.isna() & positive_tests.notna() & total_tests.notna() & cases.notna()"
+        )
+        missing_pos = df.eval(
+            "negative_tests.notna() & positive_tests.isna() & total_tests.notna() & cases.notna()"
+        )
+
+        # Keep the same order of rows in `provenance` as `df` so that the same masks can be used when
+        # writing it.
+        provenance = df.loc[:, [CommonFields.FIPS, CommonFields.DATE]]
+        provenance[CommonFields.POSITIVE_TESTS] = "none"
```
ghop02, line 89: Might be nice to have the different provenance cases as a class:
```
class TestingProvenance:
    NONE = "none"
    TESTS_AND_CASES = "tests_and_cases"
    ...

```

libs/datasets/sources/covid_county_data.py
```diff
@@ -40,10 +44,88 @@ class CovidCountyDataDataSource(data_source.DataSource):
...
+        provenance[CommonFields.NEGATIVE_TESTS].mask(missing_neg, "missing_neg", inplace=True)
+
+        df[CommonFields.POSITIVE_TESTS].mask(
+            missing_pos,
+            df[CommonFields.TOTAL_TESTS] - df[CommonFields.NEGATIVE_TESTS],
+            inplace=True,
+        )
+        provenance[CommonFields.POSITIVE_TESTS].mask(missing_pos, "missing_pos", inplace=True)
+
+        provenance_series = libs.datasets.combined_datasets._to_timeseries_rows(
```
ghop02, line 118: could you make this function public if you're going to call it from outside of combined_datasets?

ghop02: also, without looking at the code, I'm not totally sure what this does.  What is a timeseries row in this context?

libs/datasets/sources/covid_county_data.py
```diff
@@ -40,10 +44,88 @@ class CovidCountyDataDataSource(data_source.DataSource):
         }
     }
 
+    @classmethod
+    def synthesize_test_metrics(cls, data: pd.DataFrame) -> Tuple[pd.DataFrame, pd.Series]:
```
ghop02, line 48: A bit of a larger question, but should this belong in covid data public or covid data model? After reading through this a bit more, I would expect it to be in covid-data-public - mostly in the spirit of keeping the data source wrappers pretty lightweight.

libs/datasets/combined_datasets.py
```diff
@@ -111,7 +111,9 @@
 )
 
 
-def build_us_with_all_fields() -> Tuple[TimeseriesDataset, LatestValuesDataset]:
+def build_us_with_all_fields() -> Tuple[
```
ghop02, line 114: The meaning of this function feels like it's drifted - this is a bit verbose, but maybe this could be named something along the lines of  `load_data_sources_and_build_us_combined_datasets()`?

TomGoBravo: Agreed that it has drifted and not so great as a function now. I took a different approach with 5ac9081 and copied it to the one place calling it.

libs/datasets/sources/covid_county_data.py
TomGoBravo: made public and renamed

libs/datasets/sources/covid_county_data.py
TomGoBravo: I'm not at all sure we want to stick with this, for now think of these values as more like logging comments that ride along with the data.

libs/datasets/sources/covid_county_data.py
TomGoBravo: With the idea that maybe we want all of cases, total tests, pos tests, neg tests in the model I think it make sense to keep this transformation here. I see what happens in covid-data-public as pretty bare-bones transformation to the common format with very little business logic that we may be subject to experimentation. stuff in this repo may need more careful consideration. the line isn't between the two isn't very sharp to me right now so there is a bit of 'ship whatever is easiest'

PR #620, opened 2020-08-18
add pipeline.Region and use it in WebUIDataAdaptorV1.map_fips. rm some unused code

TomGoBravo:
## This PR
https://trello.com/c/nYgDEl3M/385-ability-to-query-raw-data-by-msa

* adds `pipeline.Region` and some wrappers for it that hold data for a region. Use them in WebUIDataAdaptorV1.map_fips.
*  deletes `get_compartment_value_on_date` which is only used in `notebooks/compare_ca_county_hospitalizations.ipynb`
* deletes some other unused code


libs/pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+        df = pd.read_json(output_file, dtype={"fips": "str"})
+        if len(self.fips) == 2:
+            return df.iloc[0].to_dict()
+        else:
+            return df.set_index("fips").loc[self.fips].to_dict()
+
+    def get_population(self) -> int:
+        """Gets the population for this region."""
+        population_data = FIPSPopulation.local().population()
+        return population_data.get_record_for_fips(self.fips)[CommonFields.POPULATION]
```
ghop02, line 55: I would prefer to get this from the us_latest I think - that's what we do in most other places.  the reliance on fips population was a relic I think

libs/pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+        output_filename = pyseir.utils.get_run_artifact_path(
+            self.fips, pyseir.utils.RunArtifact.ENSEMBLE_RESULT
+        )
+        if not os.path.exists(output_filename):
+            return None
+
+        with open(output_filename) as f:
+            return json.load(f)
+
+    def load_Rt_result(self) -> Optional[pd.DataFrame]:
```
ghop02, line 68: would prefer `load_rt_results` to move toward standard python naming conventions

libs/pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __repr__(self) -> str:
+        return f"Region(fips={self.fips})"
+
+    def get_us_latest(self):
+        """Gets latest values for a given state or county fips code."""
+        us_latest = load_us_latest_dataset()
+        return us_latest.get_record_for_fips(self.fips)
+
+    def load_inference_result(self):
```
ghop02, line 36: I would push back against having `load_{inference_result,rt_result,ensemble_result}` in this class.  Those output files are pure intermediaries and I think it muddies what this Region object represents a bit.  It feels more natural to say "i'd like the population for this region" - less so to say "i'd like the output of a previous step in the pyseir code".

libs/pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+            Dictionary of fit result information.
+        """
+        output_file = get_run_artifact_path(self.fips, RunArtifact.MLE_FIT_RESULT)
+        df = pd.read_json(output_file, dtype={"fips": "str"})
+        if len(self.fips) == 2:
+            return df.iloc[0].to_dict()
+        else:
+            return df.set_index("fips").loc[self.fips].to_dict()
+
+    def get_population(self) -> int:
```
ghop02, line 52: I do realize it's loading data, but this feels much cleaner as a property `population` rather than `get_population`.  Maybe the region object could load the `us_latest` on initialization or store the first call to `get_us_latest`?

libs/pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+
+        path = pyseir.utils.get_run_artifact_path(
+            self.fips, pyseir.utils.RunArtifact.RT_INFERENCE_RESULT
+        )
+        if not os.path.exists(path):
+            return None
+        return pd.read_json(path)
+
+    @staticmethod
+    def from_fips(fips: str):
```
ghop02, line 88: Thoughts on making this a classmethod? I think I would be surprised if I subclassed `Region` and it returned theh super class

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -42,65 +38,30 @@ def __init__(
...
-
-    def map_fips(self, fips: str) -> None:
+    def map_fips(self, region: libs.pipeline.Region) -> None:
         """
         For a given fips code, for either a county or state, generate the CAN UI output format.
 
         Parameters
         ----------
-        fips: str
-            FIPS code to map.
```
ghop02, line 87: Can you add region to the docstring?

libs/pipeline.py
ghop02: having a module like `pyseir_pipeline_intermediaries` where `load_inference_result` could live feels a bit more clear.

libs/pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+from typing import Optional
+
+import pandas as pd
+from pydantic import BaseModel
+
+import pyseir
+from covidactnow.datapublic.common_fields import CommonFields
+from libs.datasets import FIPSPopulation
+from libs.datasets.combined_datasets import load_us_latest_dataset, _log
```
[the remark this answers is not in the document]
TomGoBravo: Fixed both of these.

libs/pipeline.py
TomGoBravo: Done

libs/pipeline.py
TomGoBravo: Done

libs/pipeline.py
TomGoBravo: I created three classes and while they could be made related by inheritance I don't think that helps us so I didn't do it. Now `Region` is a pure identifier and there are other classes for passing around data with a `Region`. Good enough to ship? @ghop02

libs/pipeline.py
TomGoBravo: Done. I think my C++ and Java muscle memory is just more comfortable with methods than direct attribute access.

libs/pipeline.py
```diff
@@ -0,0 +1,130 @@
...
+        """Gets latest values for a given state or county fips code."""
+        us_latest = combined_datasets.load_us_latest_dataset()
+        return us_latest.get_record_for_fips(self.region.fips)
+
+    def get_population(self) -> int:
+        """Gets the population for this region."""
+        return self.get_us_latest()[CommonFields.POPULATION]
+
+
+class RegionalApiInput(BaseModel):
```
ghop02, line 61: What is the intent of Api Input?  just worried that API is a bit of an overloaded term

libs/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+    """Identifies a geographical area and wraps access to any related data read by the WebUIDataAdaptorV1."""
+
+    region: Region
+
+    # Don't access this directly. It'd be private (_combined_data) if pydantic supported it or we
```
ghop02, line 66: Do you think we gain a lot of value from using a pydantic base model over a normal python class??

libs/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @property
+    def population(self):
+        return self.combined_data.population
+
+    @property
+    def fips(self) -> str:
+        return self.region.fips
+
+    def get_us_latest(self):
```
ghop02, line 84: Feels like it would be cleaner to pass the RegionalCombinedData as a separate object? but don't care too much.

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -42,44 +38,29 @@ def __init__(
...
-        """
-        Get the population for a region.
-
-        Parameters
-        ----------
-        fips: str
-            2 digits for a state or 5 digits for a county.
-        """
-        return self.population_data.get_record_for_fips(fips)[CommonFields.POPULATION]
+    def map_fips(self, region: pipeline.RegionalApiInput) -> None:
```
ghop02, line 43: `regional_inputs` `regional_data`?

libs/pipeline.py
TomGoBravo: Inheritance? :-P
https://www.youtube.com/watch?v=3MNVP9-hglc

libs/pipeline.py
TomGoBravo: Changing to the builtin (starting 3.7) `@dataclass(frozen=True)` which supports sunder `_private` attribute names. Much nicer.

libs/pipeline.py
TomGoBravo: Changed to `RegionalWebUIInput`

pyseir/deployment/webui_data_adaptor_v1.py
TomGoBravo: Done

libs/pipeline.py
TomGoBravo: You mean split the identifier thingy (`Region`) from the data holding thingies (`RegionalCombinedData` and `RegionalWebUIInput`)? Yes, could be cleaner but don't /think/ I'm painting us into a corner and just want to shipit.

pyseir/deployment/webui_data_adaptor_v1.py
TomGoBravo: Done

PR #622, opened 2020-08-19
splitting out tests + adding proper interpolation for stalled cumulatives

ghop02:


api/README.V1.md
```diff
@@ -235,6 +235,23 @@ Utilization of hospital resources.
...
 
+### Metrics
+Calculated metrics data based on known actuals.
+
+| name           | type   | description                                                                       |
+|----------------|--------|-----------------------------------------------------------------------------------|
+| testPositivity | number | Ratio of people who test positive calculated using a 7 day rolling average.       |
+| caseDensity    | number | The number of cases per 100k population calculated using a 7 day rolling average. |
+
+### MetricsTimeseriesRow
```
benholtzmansmith, line 246: Minor: The fact that both of these classes essentially have the same documentation kind of indicates to me that we probably don't need both.

api/schemas/AggregateRegionSummary.json
```diff
@@ -174,6 +174,27 @@
         "ICUBeds"
       ]
     },
+    "Metrics": {
```
benholtzmansmith, line 177: How do the API schemas get used?

api/README.V1.md
ghop02: Yeah I go back and forth there.  I think the cost of having it is pretty small, but definitely a bit awkward

api/schemas/AggregateRegionSummary.json
ghop02: They're now used a couple ways.

1. The frontend code actually consumes them to generate typed classes.
2. we use it now to autogenerate the readme documentation

PR #624, opened 2020-08-20
Ensure that NaNs accross all fields stay NaN

ghop02:
By setting min_count to 1, it requires 1 non-na number to be present to produce a value.


ghop02:
[New York County, NY has all zeros for current_icu and current_hospitalized](https://trello.com/c/D2dj7LcR/393-new-york-county-ny-has-all-zeros-for-currenticu-and-currenthospitalized)

PR #626, opened 2020-08-20
Tweaking averages to more accurately match the website

ghop02:
A couple more small tweaks to fix how data is calculated when data backslides

### DIFF
--- a/libs/pipelines/api_pipeline.py
+++ b/libs/pipelines/api_pipeline.py
@@ -84,10 +84,8 @@ def generate_metrics_and_latest_for_fips(
     Returns:
         Tuple of MetricsTimeseriesRows for all days and the latest.
     """
-    metrics_timeseries = (
-        calculate_top_level_metrics_for_timeseries(timeseries, latest)
-        .reset_index()
-        .to_dict(orient="records")
+    metrics_timeseries = calculate_top_level_metrics_for_timeseries(timeseries, latest).to_dict(
+        orient="records"
     )
     metrics_for_fips = [MetricsTimeseriesRow(**metric_row) for metric_row in metrics_timeseries]
     if len(metrics_for_fips):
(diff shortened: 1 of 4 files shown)


libs/pipelines/api_pipeline.py
```diff
@@ -84,10 +84,8 @@ def generate_metrics_and_latest_for_fips(
     Returns:
         Tuple of MetricsTimeseriesRows for all days and the latest.
     """
-    metrics_timeseries = (
-        calculate_top_level_metrics_for_timeseries(timeseries, latest)
-        .reset_index()
-        .to_dict(orient="records")
+    metrics_timeseries = calculate_top_level_metrics_for_timeseries(timeseries, latest).to_dict(
+        orient="records"
```
TomGoBravo, line 88: what happened to reset_index?

libs/top_level_metrics.py
```diff
@@ -46,12 +47,11 @@ def calculate_top_level_metrics_for_timeseries(timeseries: TimeseriesDataset, la
...
     top_level_metrics_data = {
         MetricsFields.CASE_DENSITY: case_density,
+        CommonFields.DATE: data[CommonFields.DATE],
         MetricsFields.TEST_POSITIVITY: test_positivity,
         CommonFields.FIPS: fips,
-        CommonFields.DATE: data[CommonFields.DATE],
     }
-
-    return pd.DataFrame(top_level_metrics_data).replace({np.nan: None}).set_index(CommonFields.DATE)
+    return pd.DataFrame(top_level_metrics_data, index=test_positivity.index).replace({np.nan: None})
```
TomGoBravo, line 54: add comment about why np.nan is replaced with None. my hunch is that forces the column to be changed from float to object, which seems like it'd hurt performance.

libs/pipelines/api_pipeline.py
ghop02: I removed setting the index in the df that was returned

PR #629, opened 2020-08-24
ensemble_runner.RegionalInput and more use of Region

TomGoBravo:
## This PR

* Adds `ensemble_runner.RegionalInput`, wrapping access to data used by the ensemble_runner code
* Moves `Region` further up the stack in `pyseir/cli.py` so we're closer to using it between stages of the pipeline. In this process, removed some `_run...` methods that added a list region parameter where the call in the main pipeline has only one region.
* Adds some more constructors making it easier to make `RegionalInput` objects from an existing `Region` instead of FIPS
* Deletes unused `generate_empirical_distancing_policy_by_state`
* Adds an exception when RunMode.DEFAULT is used, because it seems to be not used and isn't tested.


pyseir/cli.py
```diff
@@ -53,40 +54,37 @@ def _run_infer_rt(states: List[str], states_only=False):
...
-
-
-def _run_ensembles(states, ensemble_kwargs=dict(), states_only=False):
-    for state in states:
-        ensemble_runner.run_state(state, ensemble_kwargs=ensemble_kwargs, states_only=states_only)
-
-
 def _map_outputs(
-    states, output_interval_days=1, states_only=False, output_dir=None, run_mode="default"
+    states: List[pipeline.Region],
```
ghop02, line 58: Calling this `states` feels a bit misleading now? `state_regions` maybe? Understandable if that doesn't feel correct.

pyseir/cli.py
```diff
@@ -53,40 +54,37 @@ def _run_infer_rt(states: List[str], states_only=False):
...
-        web_ui_mapper.generate_state(state, whitelisted_county_fips=[], states_only=states_only)
+        state_input = pipeline.RegionalWebUIInput.from_region(state)
+        web_ui_mapper.generate_state(
+            state_input, whitelisted_county_fips=[], states_only=states_only
+        )
 
 
 def _state_only_pipeline(
-    state, run_mode=DEFAULT_RUN_MODE, output_interval_days=1, output_dir=None,
+    state: pipeline.Region, run_mode=DEFAULT_RUN_MODE, output_interval_days=1, output_dir=None,
```
ghop02, line 75: I'm a bit hesitant to intermix `state` as a string vs `state` as a region.  It's currently a consistent pattern that all states are 2 letter abbreviations.  I'd prefer for it to be explicit between those two until everything is consistent.  Thoughts?

ghop02: My preference might be to name the variable `region` and then assert that it's a state level region at the beginning of this function.

pyseir/ensembles/ensemble_runner.py
```diff
@@ -1,24 +1,28 @@
...
 from pyseir.parameters.parameter_ensemble_generator import ParameterEnsembleGenerator
 import pyseir.models.suppression_policies as sp
 from pyseir.utils import get_run_artifact_path, RunArtifact, RunMode
-from pyseir.inference import fit_results
-from libs.datasets import AggregationLevel
 from libs.datasets import combined_datasets
 
 
-_logger = logging.getLogger(__name__)
+_log = structlog.get_logger()
```
ghop02, line 25: would prefer to keep the `__name__` so that the logger is not the root logger? (i believe that's the case, though could definitely be wrong)

pyseir/inference/model_fitter.py
```diff
@@ -736,7 +739,7 @@ def build_county_list(state: str) -> List[str]:
     return all_fips
 
 
-def run_state(state, states_only=False):
+def run_state(state: pipeline.Region, states_only=False):
```
ghop02, line 742: similar potential `state` -> `region` comment

pyseir/cli.py
```diff
@@ -145,7 +143,8 @@ def _build_all_for_states(
             output_interval_days=output_interval_days,
             output_dir=output_dir,
         )
-        p.map(states_only_func, states)
+        states_regions = map(pipeline.Region.from_state, states)
```
ghop02, line 146: small, but I generally prefer a list comprehension to map #pythonic

pyseir/ensembles/ensemble_runner.py
```diff
@@ -58,7 +112,7 @@ class EnsembleRunner:
 
     def __init__(
         self,
-        fips,
+        regional_input: RegionalInput,
```
ghop02, line 115: Not 100% sure where to put this comment, but I'm not sure how I feel about `region` being strictly accessed through `regional_input`.  My thinking is that keeping the two separate might be cleaner: `region: Region, input_data: RegionalInput`. Having to get fips and state through nested calls like `regional_input.region.fips` feels a bit of a code smell where as having `region` as its own input to the function may be more clean?  I feel like region could definitely still be on the `RegionalInput` class (maybe as `_region`), but maybe we surfacing things from like `regional_input.fips` might be more clean?

pyseir/ensembles/ensemble_runner.py
TomGoBravo: I think this isn't needed because we're using `structlog.stdlib.LoggerFactory()` (via common_init.py) that takes care of getting a name from the stack.

pyseir/cli.py
TomGoBravo: I was thinking that in places where the type is represented in Python types (here `region: Region`) I can immediately start exercising some liberty with the var name. But I don't care much and asserts work too, so added one and renamed.

pyseir/cli.py
TomGoBravo: Done

pyseir/inference/model_fitter.py
TomGoBravo: Done

pyseir/cli.py
TomGoBravo: Done. The simplicity of map is too tempting when I have a function and a list and don't need to do anything but put them together.

pyseir/ensembles/ensemble_runner.py
TomGoBravo: I'm not a fan of passing around two objects in parallel but added RegionalInput.fips (as a property!) and state_name to reduce the r_i.region.X calls, which I agree are a little ugly.

TomGoBravo:
> No test changes? :P
> 
> but looks good to me!

Then obviously I didn't break anything :-)

PR #630, opened 2020-08-24
Adding contact tracing to API output

ghop02:
This adds in the contact tracing metric to the API output. 

Don't think there's anything too fancy here - next up will be rt

### DIFF
--- a/api/can_api_definition.py
+++ b/api/can_api_definition.py
@@ -72,12 +72,20 @@ class Metrics(base_model.APIBaseModel):
 
     testPositivity: Optional[float] = pydantic.Field(
         ...,
-        description="Ratio of people who test positive calculated using a 7 day rolling average.",
+        description="Ratio of people who test positive calculated using a 7-day rolling average.",
     )
 
     caseDensity: Optional[float] = pydantic.Field(
         ...,
-        description="The number of cases per 100k population calculated using a 7 day rolling average.",
+        description="The number of cases per 100k population calculated using a 7-day rolling average.",
+    )
+
+    contactTracerCapacity: Optional[float] = pydantic.Field(
+        ...,
+        description=(
+            "Ratio of currently hired tracers to estimated "
+            "tracers needed based on 7-day daily case average."
+        ),
     )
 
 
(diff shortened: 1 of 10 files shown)


api/can_api_definition.py
```diff
@@ -72,12 +72,20 @@ class Metrics(base_model.APIBaseModel):
...
+        description="Ratio of people who test positive calculated using a 7-day rolling average.",
     )
 
     caseDensity: Optional[float] = pydantic.Field(
         ...,
-        description="The number of cases per 100k population calculated using a 7 day rolling average.",
+        description="The number of cases per 100k population calculated using a 7-day rolling average.",
+    )
+
+    contactTracerCapacity: Optional[float] = pydantic.Field(
```
TomGoBravo, line 83: Nit on naming though the ship may have sailed.... capacity makes me think the value is an absolute count rather than ratio. contactTracerCaseRatio?

libs/top_level_metrics.py
```diff
@@ -7,9 +7,15 @@
 from libs import series_utils
 
 
+# We will assume roughly 5 tracers are needed to trace a case within 48h.
+# The range we give here could be between 5-15 contact tracers per case.
```
TomGoBravo, line 11: Do we have a citation somewhere in our code for the origin of the 5-15 range?

libs/top_level_metrics.py
```diff
@@ -87,21 +98,40 @@ def calculate_test_positivity(
...
 
+def calculate_contact_tracers(
+    cases: pd.Series,
+    contact_tracers: pd.Series,
+    contact_tracers_per_case: int = CONTACT_TRACERS_PER_CASE,
+) -> pd.Series:
+    """Calculates ratio of hired tracers to estimated tracers needed based on daily cases.
+
+    Args:
+        cases: Cumulative cases.
```
TomGoBravo, line 122: Do these Series have datetime index? no index but already aligned?

ghop02: unfortunately they don't have a datetime index.  I went back and forth on wether to make it so or not.  the ultimate reason I held off was because it was a bit more work to make the functions in `series_util` (for example smoothing) to work the same on data w/ a datetime index and data without. but definitely up for changing it in the future.

ghop02: but correct, they are already aligned

TomGoBravo: cool. my comment was more a request to document the existing state then change the structure in this PR.

libs/top_level_metrics.py
ghop02: Didn't see one in the javascript code :/

PR #631, opened 2020-08-24
Remove unused imports

ghop02:
Something that's been bothering me for a while - just adding a lint rule to enforce no unused imports.  

This helps keep files a bit more clean and consistent. It can be misleading to know what dependencies a module actually has if a bunch of the imports aren't being used.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -4,3 +4,11 @@
 # is as it pydantic.BaseModel definitely exists.
 ignored-modules=pydantic
 ignore=src
+
+
+[MESSAGES CONTROL]
+disable=all
+
+enable=E,
+       F,
+       unused-import
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,8 +7,11 @@ ignore=src
 
 
 [MESSAGES CONTROL]
+
+# The following options disable all errors and only enable ones we want to track.
 disable=all
 
+# To enable linting for more warnings, add to this list here.
 enable=E,
        F,
        unused-import
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ unittest:
 	pytest -n 2 test/
 
 lint:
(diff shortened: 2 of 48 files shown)


ghop02:
> Thank you for automating this, I'd been doing it hap-haphazardly. I'm concerned that a blanket `disable=all` by default is a downgrade but don't see the default at https://docs.pylint.org/en/1.6.0/features.html . Is it https://github.com/PyCQA/pylint/blob/master/pylintrc ?

My understanding is that our previous automated linting was relying on the "pytest --pylint -m pylint --pylint-error-types=EF ."  to only check for errors and failures.  

I'm not sure I totally understand what you mean by:  "but don't see the default at <file>"? I think I'm trying to emulate what they're describing [here](https://github.com/PyCQA/pylint/blob/master/pylintrc#L49-L52).  i.e. disable everything, but enable a few warnings (we can add more in as we feel comfortable)

TomGoBravo:
> > Thank you for automating this, I'd been doing it hap-haphazardly. I'm concerned that a blanket `disable=all` by default is a downgrade but don't see the default at https://docs.pylint.org/en/1.6.0/features.html . Is it https://github.com/PyCQA/pylint/blob/master/pylintrc ?
> 
> My understanding is that our previous automated linting was relying on the "pytest --pylint -m pylint --pylint-error-types=EF ." to only check for errors and failures.
> 
> I'm not sure I totally understand what you mean by: "but don't see the default at "? I think I'm trying to emulate what they're describing [here](https://github.com/PyCQA/pylint/blob/master/pylintrc#L49-L52). i.e. disable everything, but enable a few warnings (we can add more in as we feel comfortable)

Sounds good to me. I hadn't seen the comments you highlighted in `master/pylintrc`. Perhaps add a comment in our pylintrc with link to that file.

PR #632, opened 2020-08-25
add TimeseriesDataset.get_columns_and_date_subset

TomGoBravo:

This PR moves combined_datasets.get_timeseries_for_fips columns and min_range_with_some_value logic to TimeseriesDataset.get_columns_and_date_subset so it can be run on data not read with combined_datasets

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -175,6 +175,16 @@ class TimeseriesDataset(dataset_base.DatasetBase):
         )
         return self.__class__(self.data.loc[row_binary_array, :])
 
+    def get_columns_and_date_subset(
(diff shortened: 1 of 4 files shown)


libs/datasets/timeseries.py
```diff
@@ -175,6 +175,16 @@ def get_subset(
         )
         return self.__class__(self.data.loc[row_binary_array, :])
 
+    def get_columns_and_date_subset(
```
ghop02, line 178: Thoughts on just adding these additional args to `get_subset`? They seem useful

TomGoBravo: I thought about it but decided to not overload get_subset more and keep the existing flow as much as possible until I have more confidence that I can quickly work out that i've broken something.

PR #633, opened 2020-08-25
Fix pyseir load hospitalization bug

ghop02:


ghop02:
Merging to unblock the build - happy to fix any comments tomorrow

PR #636, opened 2020-08-25
Add infection rate

ghop02:
This adds in the infection rate and infection rate confidence interval into the metrics output.

A bit of annoying logic to pull the rt from 7 days in the past, but came out alright.

### DIFF
--- a/test/libs/top_level_metrics_test.py
+++ b/test/libs/top_level_metrics_test.py
@@ -1,9 +1,22 @@
 import io
+
+import pytest
 import numpy as np
 import pandas as pd
 from covidactnow.datapublic.common_fields import CommonFields
+from covidactnow.datapublic import common_df
 from libs import top_level_metrics
 from libs.datasets.timeseries import TimeseriesDataset
+from libs.datasets.sources.can_pyseir_location_output import CANPyseirLocationOutput
+
+
+def _build_metrics_df(content: str) -> pd.DataFrame:
+    header = (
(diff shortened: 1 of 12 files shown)


test/libs/top_level_metrics_test.py
```diff
@@ -70,13 +83,59 @@ def test_top_level_metrics_basic():
         CommonFields.POPULATION: 100_000,
         CommonFields.FIPS: "36",
     }
-    results = top_level_metrics.calculate_top_level_metrics_for_timeseries(timeseries, latest)
+    results = top_level_metrics.calculate_metrics_for_timeseries(timeseries, latest, None)
+
+    header = (
```
TomGoBravo, line 88: why not use  _build_metrics_df ?

test/libs/top_level_metrics_test.py
```diff
@@ -70,13 +83,59 @@ def test_top_level_metrics_basic():
...
+    data = common_df.read_csv(data, set_index=False)
+    model_output = CANPyseirLocationOutput(data)
+
+    latest = {
+        CommonFields.POPULATION: 100_000,
+        CommonFields.FIPS: "36",
+    }
+    results = top_level_metrics.calculate_metrics_for_timeseries(timeseries, latest, model_output)
+    header = (
+        "date,fips,caseDensity,testPositivityRatio,contactTracerCapacityRatio,"
```
TomGoBravo, line 130: why not use _build_metrics_df ?

PR #638, opened 2020-08-25
Separate Generation of Infection Rate Metric from PySEIR

BrettBoval:
https://trello.com/c/5KduQwfy/401-create-infer-rt-results-table-and-persist-to-snapshot-artifact

### Please Check
- [ ] Are tests passing?


pyseir/cli.py
```diff
@@ -48,6 +49,26 @@ def _generate_whitelist():
     gen.generate_whitelist()
 
 
+def _generate_infection_rate_metric(regions: List[pipeline.Region] = None):
+    """
+    Apply infer_rt.run_rt_for_fips for each region in regions and return a combined results table
+    with a unique row for each region+date combination.
+    """
+    if regions is None:
```
ghop02, line 57: `if not regions`

pyseir/cli.py
```diff
@@ -48,6 +49,26 @@ def _generate_whitelist():
...
+        regions = combined_datasets.load_us_latest_dataset().all_fips
+
+    with Pool(maxtasksperchild=10) as p:
+        results = p.map(infer_rt.run_rt_for_fips, regions)
+
+    def label_data(label, data):
+        """Modify the returned dataframe to meet requirements"""
+        data = data.reset_index(drop=False)  # Move date column from index to columns
+        data["fips"] = label  # Add region column id
```
ghop02, line 66: Can you add this in the `run_infer_rt_for_fips` code? I'd like this to be as thin of a wrapper as possible

pyseir/cli.py
```diff
@@ -48,6 +49,26 @@ def _generate_whitelist():
...
+        return data
+
+    rt_df = pd.concat([label_data(region, data) for region, data in zip(regions, results)])
```
ghop02, line 69: does it handle none appropriately?

pyseir/cli.py
```diff
@@ -19,6 +19,7 @@
 from pyseir.deployment.webui_data_adaptor_v1 import WebUIDataAdaptorV1
 from libs.datasets import combined_datasets
 from pyseir.inference.whitelist_generator import WhitelistGenerator
+from pyseir.utils import get_summary_artifact_path, SummaryArtifact
```
ghop02, line 22: I would either import this as `import pyseir.utils` or `from pyseir import utils`

ghop02: Also, can you make these two separate lines? (soon to be a lint rule 🪂 )

pyseir/cli.py
```diff
@@ -48,6 +49,26 @@ def _generate_whitelist():
     gen.generate_whitelist()
 
 
+def _generate_infection_rate_metric(regions: List[pipeline.Region] = None):
```
ghop02, line 52: Add a return type here

pyseir/cli.py
```diff
@@ -144,14 +165,23 @@ def _build_all_for_states(
         root.info("Only executing for states. returning.")
         return
 
+    # Separating the Infection Rate Calculations
+    infection_rate_metric_df = _generate_infection_rate_metric()
```
ghop02, line 169: I think this should take in the input regions - otherwise you won't be able to run on subsets locally

pyseir/cli.py
```diff
@@ -206,12 +236,13 @@ def generate_whitelist():
 @click.option(
     "--state", help="State to generate files for. If no state is given, all states are computed."
 )
-@click.option("--states-only", default=False, is_flag=True, type=bool, help="Only model states")
+@click.option("--states-only", default=False, is_flag=True, type=bool, help="Deprecated")
```
ghop02, line 239: :( - fwiw i use this a lot and find it pretty helpful!

pyseir/utils.py
```diff
@@ -53,6 +53,31 @@ class RunArtifact(Enum):
...
+    output_dir: str or NoneType
+        Output directory to obtain the path for.
+
+    Returns
+    -------
+    path: str
+        Location of the artifact.
+    """
+    output_dir = output_dir or OUTPUT_DIR
+    return os.path.join(output_dir, "pyseir", artifact.value)
```
ghop02, line 78: you should use pathlib!

ghop02, line 78: Where does pyseir come from?  maybe after this is merged, I'll take a stab at refactoring output paths

ghop02:
Also, could you add a test for generate infection rate metric? Happy to help talk through what that would look like!

pyseir/cli.py
BrettBoval: It's deprecated in the sense that this flag never actually changed anything. Calling run_infer_rt from the CLI never propagated down to running counties for the Infection Rate calculations. Right now build_all calls the cli method for states, but calls the direct method for fips. So calling it with the flag is the same as calling it without the flag.

pyseir/utils.py
BrettBoval: Agreed, but all the other code in PySEIR uses os.path and I wasn't ready to sign up for changing all of it.

BrettBoval: This path logic is hidden in the body of the call.

I'd like you to grab the Infection Rate and the ICU from the new tables, and then refactoring output paths will be removing paths. 🤞

BrettBoval:
> Also, could you add a test for generate infection rate metric? Happy to help talk through what that would look like!

Yeah. How about we put something on the calendar tomorrow morning? Ideally before 10:30am EST if you can.

pyseir/cli.py
BrettBoval: It's a moot point because I removed that outside this function after talking with Tom, but I thought `if X is None` was how to check for None -> https://www.python.org/dev/peps/pep-0008/#programming-recommendations. Has that been updated somewhere else?

pyseir/cli.py
BrettBoval: Good point. I'll have it return an empty dataframe if results evaluates false.

pyseir/cli.py
BrettBoval: done

pyseir/cli.py
BrettBoval: I went with `import pyseir.utils`, but that makes some very long function names. But I agree that pyseir.utils.X is needed to know where it comes from.

pyseir/cli.py
BrettBoval: I wanted three things here: (1) if regions is None run for all fips, (2) if empty (evaluates to False) then return empty Dataframe, and (3) if not empty run for each item in the list.

pyseir/cli.py
BrettBoval: done

BrettBoval:
> Also, could you add a test for generate infection rate metric? Happy to help talk through what that would look like!

Just added a couple use cases (including the empty/full combo you mentioned).

BrettBoval:
@TomGoBravo PTAL at the merge conflict for pyseir/cli.py. Feel free to change things up, especially around how to presist the infection metric table to the snapshot archive.

test/infer_rt_test.py
```diff
@@ -193,3 +194,41 @@ def test_smoothing_and_causality():
...
+    df = cli._generate_infection_rate_metric(regions=REGIONS)
+    assert df.empty
+
+
+def test_generate_infection_rate_metric_one_empty():
+    FIPS = [
+        "51017",  # Bath County VA Almost No Cases. Will be filtered out under any thresholds.
+        "51153",  # Prince William VA Lots of Cases
+    ]
+    REGIONS = [infer_rt.RegionalInput.from_fips(region) for region in FIPS]
```
ghop02, line 211: nit: generally all caps are indicate global constants [heres a great guide](https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations), can you make these lowercase?

test/infer_rt_test.py
```diff
@@ -193,3 +194,41 @@ def test_smoothing_and_causality():
         ),
         "test_smoothing_and_causality",
     )
+
+
+def test_generate_infection_rate_metric_no_region_given():
+    FIPS = []
+    REGIONS = [infer_rt.RegionalInput.from_fips(region) for region in FIPS]
+    df = cli._generate_infection_rate_metric(regions=REGIONS)
```
ghop02, line 202: i'd prefer not to call this with the variable name, since it's a positional/non-optional parameter.  
i.e. `cli._generate_infection_rate_metric(REGIONS)`.  main reason for preferring this is 2 parts:
 - makes it clear which arguments are actually optional
 - lets you rename the positional argument without having to update everywhere

pyseir/rt/infer_rt.py
```diff
@@ -91,19 +94,31 @@ def _generate_input_data(
...
+            observed_new_cases,
+            observed_new_deaths,
+        ) = load_data.calculate_new_case_data_by_region(
+            regional_input.get_timeseries(),
+            t0=InferRtConstants.REF_DATE,
+            include_testing_correction=include_testing_correction,
+        )
+    except AssertionError as e:
+        rt_log.exception(
+            event=f"An AssertionError was raised in the loading of {regional_input.display_name}"
```
ghop02, line 117: I'm not 100% sure how structlog handles this in sentry, but in general it's good to leave variables out of the main message so that all errors of a similar origin are grouped together

test/infer_rt_test.py
BrettBoval: done

pyseir/cli.py
```diff
@@ -50,6 +50,20 @@ def _generate_whitelist():
     gen.generate_whitelist()
 
 
+def _generate_infection_rate_metric(regions: List[pipeline.Region]) -> pd.DataFrame:
```
TomGoBravo, line 53: I think the type is meant to be `List[infer_rt.RegionalInput]`

pyseir/cli.py
```diff
@@ -146,30 +160,47 @@ def _build_all_for_states(
         root.info("Only executing for states. returning.")
         return
 
-    all_county_fips = build_counties_to_run_per_state(states, fips=fips)
+    # Get the Lists of Fips to Calculate
+    pyseir_target_fips = build_counties_to_run_per_state(states, fips=fips)
+
+    infer_rt_target_fips = [fips] if fips else combined_datasets.load_us_latest_dataset().all_fips
```
TomGoBravo, line 166: why go back to combined_datasets to get all fips instead of using pyseir_target_fips + states_regions ?

pyseir/rt/infer_rt.py
```diff
@@ -91,19 +94,32 @@ def _generate_input_data(
...
+        (
+            times,
+            observed_new_cases,
+            observed_new_deaths,
+        ) = load_data.calculate_new_case_data_by_region(
+            regional_input.get_timeseries(),
+            t0=InferRtConstants.REF_DATE,
+            include_testing_correction=include_testing_correction,
+        )
+    except AssertionError as e:
```
TomGoBravo, line 115: Oh, sorry you needed this. Let me know what fix you'd like from https://covidactnow.slack.com/archives/C0110QQAPJA/p1598547382041700?thread_ts=1598483626.040100&cid=C0110QQAPJA

pyseir/rt/infer_rt.py
```diff
@@ -570,7 +586,16 @@ def infer_all(self, plot=True, shift_deaths=0):
             df_raw[timeseries_type.value] = timeseries_raw
 
             df = pd.DataFrame()
-            dates, posteriors, start_idx = self.get_posteriors(timeseries_type)
+            try:
+                dates, posteriors, start_idx = self.get_posteriors(timeseries_type)
+            except Exception as e:
+                structlog.getLogger().exception(
```
TomGoBravo, line 592: why not rt_log?

pyseir/rt/infer_rt.py
```diff
@@ -570,7 +586,16 @@ def infer_all(self, plot=True, shift_deaths=0):
...
+                    event="Posterior Calculation Error",
+                    region=self.regional_input.display_name,
+                    error=e,
```
TomGoBravo, line 595: If you need to explicitly pass e to `exception()` our logger is not configured correctly. https://github.com/hynek/structlog/issues/31#issuecomment-69319872

pyseir/rt/infer_rt.py
BrettBoval: done

pyseir/rt/infer_rt.py
BrettBoval: Right now I am going to filter out all fips that end in "999" upstream. So this additional context shouldn't be necessary going forward. I'm going to leave it in for right now, but plan to remove it when I fix the MA counties issues too.

pyseir/rt/infer_rt.py
```diff
@@ -91,19 +94,32 @@ def _generate_input_data(
...
+            regional_input.get_timeseries(),
+            t0=InferRtConstants.REF_DATE,
+            include_testing_correction=include_testing_correction,
+        )
+    except AssertionError as e:
+        rt_log.exception(
+            event="An AssertionError was raised in the loading of the data for the calculation of "
+            "the Infection Rate Metric",
+            region=regional_input.display_name,
+            error=e,
```
ghop02, line 120: Can you remove the error from here? I think it should get included automatically

pyseir/cli.py
BrettBoval: The inclusion criteria is different for pyseir than for the infection rate calculation. Essentially, the infection rate calculation can tolerate lower absolute numbers but is more sensitive to needing recent data. In the past, the infection whitelist was a subset of the pyseir SEIR whitelist. That was mostly historical convenience, and I am more explicitly separating that going forward.

pyseir/rt/infer_rt.py
BrettBoval: Fixed.

test/infer_rt_test.py
```diff
@@ -128,22 +129,22 @@ def test_med_scale_strong_growth_and_decay():
...
-            disable_deaths=True,
-            scale=5.0,
-            ratechange1=RateChange(0, 1.0),
-            ratechange2=RateChange(70, 1.2),
-        ),
-        "test_low_cases_weak_growth",
-    )
-    # check_standard_assertions(rt1, rt2, t_switch, rt)
-    # TODO is failing this test rt = .84 instead of rt1
+# @pytest.mark.slow
```
ghop02, line 132: Instead of commenting out this test, can you add a skip? https://docs.pytest.org/en/latest/skipping.html#skipping-test-functions

pyseir/rt/infer_rt.py
BrettBoval: fixed.

pyseir/rt/infer_rt.py
BrettBoval: fixed.

BrettBoval: Removed.

test/infer_rt_test.py
BrettBoval: Done. The note is vague because Alex had commented out the line in the test that was the actual assertion. So I don't know specifically why it's not passing, just that it the test was passing without any actual test happening.

pyseir/rt/infer_rt.py
BrettBoval: Done.

pyseir/rt/infer_rt.py
BrettBoval: Fixed.

pyseir/cli.py
BrettBoval: Yep. Fixed.

test/infer_rt_test.py
BrettBoval: Fixed.

pyseir/cli.py
BrettBoval: I changed the copy in the text to reflect that the flag is currently ignored and always defaults to states-only right now.

BrettBoval:
https://covidactnow.org/internal/compare?left=974&locations=2&metric=0&right=976&sort=3
compares 974 (master) and 976 (proposed refactor) on the same data and shows no changes to the outputted infection rate metrics for States and 100 top counties by population.

PR #639, opened 2020-08-25
add ensemble_runner.RegionalInput.from_model_fitter and use it for counties

TomGoBravo:

* In `ensemble_runner` and `webui_data_adaptor_v1` add constructor `Regionalnput.from_model_fitter` which copies data directly from a `ModelFitter` object, saving a round trip to disk.
* Remove `webui_data_adaptor_v1` `generate_state` which was only used for regions in one place.
* Remove `if __main__` in `webui_data_adaptor_v1` that looked like some kind of early test
* Remove `model_fitter.RegionalInput.load_inference_result` which wasn't used from other classes
* Remove `execute_model_for_fips` which added a branch that always went one way
* Move the `states_only=False` part of `model_fitter.run_state` to it's own function because it had a single caller and complicated using `Regionalnput.from_model_fitter`
* Adds use of a tempdir to `test_pyseir_end_to_end_idaho` so it doesn't pass when output from a previous run is still in `output/`
* Removes first of two writes of state data, at the end of `_state_only_pipeline` and inline the remaining caller of `cli._map_outputs`


pyseir/cli.py
```diff
@@ -50,42 +49,16 @@ def _generate_whitelist():
...
-        state_input = webui_data_adaptor_v1.RegionalInput.from_region(state)
-        web_ui_mapper.generate_state(
-            state_input, whitelisted_county_fips=[], states_only=states_only
-        )
-
-
 def _state_only_pipeline(
     region: pipeline.Region, run_mode=DEFAULT_RUN_MODE, output_interval_days=1, output_dir=None,
-):
+) -> model_fitter.ModelFitter:
```
ghop02, line 54: Can this return the infer rt, and ensembles results as well?

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -68,7 +72,9 @@ def load_inference_result(self) -> Mapping[str, Any]:
         : dict
             Dictionary of fit result information.
         """
-        return load_inference_result(self.region)
+        if self._mle_fit_result is not None:
+            return self._mle_fit_result
+        return pipeline.load_inference_result(self.region)
```
ghop02, line 77: will we ever expect this to be reached?

pyseir/cli.py
ghop02: It seems like those outputs could easily be passed into the WebUIInputRegion as well. but fine tackling that later

pyseir/inference/model_fitter.py
```diff
@@ -736,52 +718,47 @@ def build_county_list(state: str) -> List[str]:
...
-
-            # Serialize the model results.
-            for fips, fitter in zip(all_fips, fitters):
-                if fitter:
-                    with open(get_run_artifact_path(fips, RunArtifact.MLE_FIT_MODEL), "wb") as f:
-                        pickle.dump(fitter.mle_model, f)
+    return model_fitter
+
+
+def run_counties_of_state(region):
```
ghop02, line 745: types + a docstring would be handy here!

test/pyseir_end_to_end_test.py
```diff
@@ -12,19 +15,22 @@
...
-    path = pathlib.Path(path)
-    assert path.exists()
-    output = CANPyseirLocationOutput.load_from_path(path)
-    data = output.data
-    with_values = data[schema.RT_INDICATOR].dropna()
-    assert len(with_values) > 10
-    assert (with_values > 0).all()
-    assert (with_values < 6).all()
+
+    with temppathlib.TemporaryDirectory() as tmp_dir:
```
ghop02, line 19: I do find the [tmp_path](https://docs.pytest.org/en/stable/tmpdir.html) fixture handy. I know we probably have slightly different styles here :) I think I tend to think that the multiple indentation is more confusing than having pytest take care of it

test/pyseir_end_to_end_test.py
```diff
@@ -12,19 +15,22 @@
...
+        with unittest.mock.patch("pyseir.utils.OUTPUT_DIR", tmp_dir.path):
```
ghop02, line 20: This would be handy to make an environment variable I think as well

pyseir/cli.py
```diff
@@ -187,14 +152,8 @@ def _build_all_for_states(
...
-    for state in states:
-        region = pipeline.Region.from_state(state)
-        state_input = webui_data_adaptor_v1.RegionalInput.from_region(region)
-        web_ui_mapper.generate_state(
-            state_input,
-            whitelisted_county_fips=[k for k, v in all_county_fips.items() if v == state],
-            states_only=False,
-        )
+    for fitter in itertools.chain(state_fitters, county_fitters):
+        web_ui_mapper.map_fips(webui_data_adaptor_v1.RegionalInput.from_model_fitter(fitter))
```
ghop02, line 156: might be nice to rename map_fips without the map - it makes me thing we're running over multiple fips

ghop02, line 156: This doesn't look like it's in parallel.  we probably want to run this in a pool right?

pyseir/cli.py
TomGoBravo: Yes, my thought was get this merged with one output, then make another pass for the others.

pyseir/deployment/webui_data_adaptor_v1.py
TomGoBravo: Hopefully not long term. For now it is reached by code that uses the RegionalInput.from_fips constructor, which is called by the non build-all flows.

pyseir/inference/model_fitter.py
TomGoBravo: Done

test/pyseir_end_to_end_test.py
TomGoBravo: I'm torn between my dislike of the fixture magic driven by parameter names and having shorter lines of code :-P

test/pyseir_end_to_end_test.py
TomGoBravo: Perhaps, though I haven't been motivated for that so far...

pyseir/cli.py
TomGoBravo: Done, `write_region`

TomGoBravo: Done

PR #641, opened 2020-08-26
Switching github action black formatter to a version that pins black …

ghop02:
…version

The latest version of black causes a bunch of changes (that I'm not totally sure are correct).  This github action uses a pinned version of black so won't break going forward

### DIFF
--- a/.github/workflows/python_unit_tests.yml
+++ b/.github/workflows/python_unit_tests.yml
@@ -10,9 +10,13 @@ jobs:
     - name: Checkout covid-data-model
       uses: actions/checkout@v2
     - name: Run Black
-      uses: lgeiger/black-action@v1.0.1
+      uses: RojerGS/python-black-check@1.0.1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/python_unit_tests.yml
```diff
@@ -10,9 +10,13 @@ jobs:
     - name: Checkout covid-data-model
       uses: actions/checkout@v2
     - name: Run Black
-      uses: lgeiger/black-action@v1.0.1
+      uses: RojerGS/python-black-check@1.0.1
```
TomGoBravo, line 13: Why not run black in the same way that `make test` is run so it more closely replicates what happens on our local workstations?

ghop02: I do like that idea - the main reason was for speed, the environment needed is a bit heavy (though many ways around that problem) - definitely down to address in the future though

PR #643, opened 2020-08-26
Icu headroom

ghop02:
this is a first pass at calculating the icu headroom numbers on the backend.  I still need to properly output top level metrics in the api code, but wanted a first pass.

Still to do:
 - Handle missing data better (i.e. don't have logic checking for recency)
 - Fix up tests
 - Add better top level description to the api


libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+    def actual_current_icu_covid(self) -> Optional[pd.Series]:
+        actuals = self.data[CommonFields.CURRENT_ICU]
+        if not actuals.any():
+            return None
+
+        return actuals
+
+    @property
+    def estimated_current_icu_covid(self) -> Optional[pd.Series]:
+        estimated = self.data[schema.CURRENT_ICU]
```
mikelehen, line 58: FWIW `schema` seems overly generic to the point of being meaningless (what's the difference between `CommonFields.CURRENT_ICU` and `schema.CURRENT_ICU`? I have no idea.) If you need to shorten it, perhaps `model_schema`?

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+class ICUMetricData:
+    def __init__(self, data, latest_values, decomp_factor):
+        self.data = data
+        self.latest_values = latest_values
+        self.decomp_factor = decomp_factor
+
+    @property
+    def actual_current_icu_covid(self) -> Optional[pd.Series]:
+        actuals = self.data[CommonFields.CURRENT_ICU]
+        if not actuals.any():
```
mikelehen, line 51: FWIW I think the frontend looks for 1 weeks worth of data within the last 2 weeks.  Is that handled elsewhere? Or did you intentionally drop that (which isn't out of the question; it was kind of arbitrary)?

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+    @property
+    def actual_current_icu_total(self) -> Optional[pd.Series]:
+        actuals = self.data[CommonFields.CURRENT_ICU_TOTAL]
+        if not actuals.any():
+            return None
+
+        return actuals
+
+    @property
+    def total_icu_beds(self):
```
mikelehen, line 73: nit: Is `self.latest_values[CommonFields.ICU_BEDS]` a scalar value (as opposed to timeseries)?  So this may return a pd.Series or a scalar value?  Could you document that in the return type like you do for the above property definitions in this file?

In general, I'm curious how you're deciding when to document the return type...  I'm a fan of always documenting types but I have no idea what is idiomatic in python. 😁

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+
+    @property
+    def current_icu_non_covid_with_source(self):
+        if self.actual_current_icu_covid is not None and self.actual_current_icu_total is not None:
+            source = NonCovidPatientsMethod.ACTUAL
+            return (self.actual_current_icu_total - self.actual_current_icu_covid, source)
+
+        if self.actual_current_icu_total is not None:
+            source = NonCovidPatientsMethod.ESTIMATED_FROM_TOTAL_ICU_ACTUAL
+            return (self.actual_current_icu_total - self.estimated_current_icu_covid, source)
```
mikelehen, line 91: What does this do if estimated_current_icu_covid is None?  Hopefully just return None?  (just confirming we don't need to specifically handle it)

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Do decomp
+        non_covid_utilization = self.typical_usage_rate - self.decomp_factor
+        source = NonCovidPatientsMethod.ESTIMATED_FROM_TYPICAL_UTILIZATION
+
+        if isinstance(self.total_icu_beds, pd.Series):
+            decomp = non_covid_utilization * self.total_icu_beds
+            decomp.name = None
```
mikelehen, line 99: what is this for?

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+            return (decomp, source)
+
+        total_icu_beds = pd.Series([self.total_icu_beds] * len(self.data), index=self.data.index)
+        decomp = non_covid_utilization * total_icu_beds
```
mikelehen, line 103: 1. Would this code be simpler if you changed self.total_icu_beds to always be a series?  (just extend the scalar into a series if we only have the latest value)
2. "decomp" seems like a not-super-useful variable name.  Seems like it should be `non_covid_patients` or something.  I guess it could be `non_covid_patients_via_decomp` but seems unnecessary?

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+    If we know exactly how many beds are in use by non covid patients we can
+    calculate the number of available beds exactly.
+
+    otherwise available beds is calulated by looking at historical data and
+    assuming that we can add more capacity.
+
+    So, we need to know or guess how many icu beds are occupied by non covid patients
+    non covid patients:
+     - get from actuals
+     - estimate from typical utilization + decomp factor
```
mikelehen, line 134: Can you turn this into prose to match the rest of the comment?  I'm having trouble processing it.

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+
+    available beds can be calulated
```
mikelehen, line 136: Not sure what this is referring to.

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+                ----------------------
+           total beds - non covid patients
+                          ==
+                  covid icu patients
+                ---------------------
+    total beds - (typical non covid usage - decomp)
+                          or
+                  covid icu patients
+                ---------------------
+    total beds - (total covid patients - covid patients)
```
mikelehen, line 148: "total covid patients - covid patients" seems wrong.

In general, I'm not sure how helpful this comment is (trying to describe all the estimation calculations up front instead of next to the code that implements it)

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+    non covid patient source: Actual total patients |
+                              Estimated from actual total patients |
+                              Estimated from typical utilization
+    covid patient source:     Actuals | Estimates
```
mikelehen, line 154: I guess this is trying to describe the return value types (but it's kinda' loose... the exact casing / formatting doesn't match the actual types)?  Does python not make it easy / possible to do this through actual type definitions?  If not, should you use the actual enum names in this comment?

libs/icu_headroom_metric.py
```diff
@@ -0,0 +1,168 @@
...
+
+    """
+    non_covid_patients, non_covid_source = icu_data.current_icu_non_covid_with_source
+    current_icu_covid, covid_source = icu_data.current_icu_covid_with_source
+    metric = current_icu_covid / (icu_data.total_icu_beds - non_covid_patients)
+
+    print(icu_data.total_icu_beds)
+    print(non_covid_patients, non_covid_source)
```
mikelehen, line 162: Did you mean to leave these prints here?

libs/icu_headroom_metric.py
ghop02: ah yeah, that's on my list of things to clean up

libs/icu_headroom_metric.py
ghop02: ah yes `s/total covid/total icu/`

sorry I should've given a bit more of a heads up about this comment as I'm not sure if it's really that useful. it was more useful to me as I was working through the different options available.  I do think some context is useful, so i'll try to boil it down in a more approachable/concise manner

libs/icu_headroom_metric.py
ghop02: ah I'm generally very very pro return types in python! I just hadn't gone through and cleaned up the type annotations here. I do try to have return types as much as possible, but sometimes add them as a last step

libs/series_utils.py
```diff
@@ -64,3 +65,23 @@ def interpolate_stalled_values(series: pd.Series) -> pd.Series:
...
+        days_back: Number of days back to look
+        required_non_null_datapoints: Number of non-null data points required.
+
+    Returns: True if has recent data, otherwise false.
+    """
+    today = datetime.datetime.today().date()
+    start_date = today - timedelta(days=days_back)
+    recent_series = series[start_date:]
+    num_datapoints = recent_series.notnull().sum()
+    nonzero_points = (recent_series > 0).sum()
```
TomGoBravo, line 86: since you aren't comparing this to a count of required_nonzero_points ... s/sum/any/

PR #644, opened 2020-08-26
Switching to use a datetime index to make passing around data more lightweight

ghop02:
Wanted to get this in to make some of the testing easier/make it easier to pass around individual series rather than dataframes

### DIFF
--- a/libs/top_level_metrics.py
+++ b/libs/top_level_metrics.py
@@ -53,7 +53,7 @@ def calculate_metrics_for_timeseries(
     fips = latest[CommonFields.FIPS]
     population = latest[CommonFields.POPULATION]
 
-    data = timeseries.data
+    data = timeseries.data.set_index(CommonFields.DATE)
 
     infection_rate = np.nan
     infection_rate_ci90 = np.nan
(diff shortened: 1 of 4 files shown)


libs/top_level_metrics.py
```diff
@@ -53,7 +53,7 @@ def calculate_metrics_for_timeseries(
     fips = latest[CommonFields.FIPS]
```
TomGoBravo, line 53: line above can be `assert timeseries.has_one_region()`

libs/top_level_metrics.py
```diff
@@ -99,17 +94,15 @@ def calculate_metrics_for_timeseries(
         cumulative_cases, data[CommonFields.CONTACT_TRACERS_COUNT]
     )
     top_level_metrics_data = {
-        CommonFields.DATE: data[CommonFields.DATE],
         CommonFields.FIPS: fips,
         MetricsFields.CASE_DENSITY_RATIO: case_density,
```
TomGoBravo, line 98: what's the relationship between MetricsFields, can_model_output_schema and can_api_definition?

ghop02:
> Oh boy, there are a lot of corners in this code base. I hadn't looked at series_utils or top_level_metrics before.

Yup! If you have any suggestions, very open to making them better at some point.  hopefully adding date as an index will help in the future with them

libs/top_level_metrics.py
ghop02: can_model_output_schema is pretty independent from the others, it's just some very old code around what fields are in the webui output files. 

`MetricsFields` are loosely the metrics in the output data frame from `calculate_metrics_for_timeseries`.  The values of them match the keys of the `Metrics` object in can api definition.  

In general I'm still a bit unsure of how to deal with data frames like this.  A lot of time, i'd like similar functionality from the TimeseriesDataset. I definitely want to make sure we're not referring to columns by strings defined in various places. but I haven't found anything I love so far.  For instance, I thought about making a `class MetricsTimeseries(TimeseriesDataset)` but it didn't quite feel right.

PR #645, opened 2020-08-27
generate whitelist with groupby and other optimizations

TomGoBravo:
This PR is part of https://trello.com/c/nYgDEl3M/385-ability-to-query-raw-data-by-msa and got kinda big.

* `TimeseriesDataset` optimizations in `has_one_region` and `_remove_padded_nans` that trim a few seconds from generating the whitelist
* `pyseir.cli`, factor out ` _states_region_list` which is called for each entry point with a `state` parameter
* merge `cli.build_counties_to_run_per_state` and `model_fitter.build_county_list` to `regions_in_states`
* change `run_counties_of_state` to expect the counties list to be already filtered


pyseir/inference/model_fitter.py
```diff
@@ -703,12 +703,11 @@ def _persist_results_per_state(state_df):
             pickle.dump(county_series.mle_model, f)
 
 
-def build_county_list(state: str) -> List[str]:
+def build_county_list(state: str, df_whitelist: pd.DataFrame) -> List[str]:
```
ghop02, line 706: small nit: in one place this is called `df_whitelist` in others its `whitelist_df`

pyseir/inference/whitelist_generator.py
```diff
@@ -42,7 +43,7 @@ def __init__(
         self.nonzero_case_datapoints = nonzero_case_datapoints
         self.nonzero_death_datapoints = nonzero_death_datapoints
 
-    def generate_whitelist(self):
+    def generate_whitelist(self, timeseries: TimeseriesDataset) -> pd.DataFrame:
```
ghop02, line 46: How long does the new version take?

test/whitelist_generator_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+    assert to_dict(["fips"], df) == {
+        "97111": {"state": "ZZ", "county": "Bar County", "inference_ok": False},
+    }
+
+
+def test_inference_ok_with_5_days_cases_changed():
+    # 5 days with cases data isn't enough to make inference_ok, 6 days are
+    # needed so that there are 5 days with an *delta* relative to a previous day.
+    input_df = read_csv_and_index_fips_date(
```
ghop02, line 38: any reason not to just use common_df.read_csv?

pyseir/inference/whitelist_generator.py
TomGoBravo: 18 seconds in generate_whitelist :-/ Of that 10s is in get_columns_and_date_subset called by calculate_new_case_data_per_region

test/whitelist_generator_test.py
TomGoBravo: Not very strong ones... so the test doesn't need to call StringIO and consistency with read_csv_and_index_fips that doesn't parse dates.

pyseir/inference/model_fitter.py
TomGoBravo: Done

pyseir/inference/whitelist_generator.py
ghop02: hmm, could using `parallel_apply` from pandarallel speed things up?

TomGoBravo: That reduced the wall time from 20s to 13s.

pyseir/cli.py
```diff
@@ -51,7 +52,15 @@ def entry_point():
 def _generate_whitelist() -> pd.DataFrame:
     gen = WhitelistGenerator()
     all_us_timeseries = combined_datasets.load_us_timeseries_dataset()
-    return gen.generate_whitelist(all_us_timeseries).to_csv("output/whitelist.csv", index=False)
+    return gen.generate_whitelist(all_us_timeseries)
+
+
+def _states_region_list(state: Optional[str], default: List[str]) -> List[pipeline.Region]:
```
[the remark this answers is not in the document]
TomGoBravo: Yes, that is shorter and eliminates some repetitiveness but I heard a declaration that all functions high, low, big and small shall henceforth receive their inputs as parameters, not loading from global locations. :-P

PR #646, opened 2020-08-27
Setting the index to always be named date in top level metrics

ghop02:
caused this:

https://sentry.io/organizations/covidactnow/issues/1861469979/?project=5203044&query=is%3Aunresolved


libs/top_level_metrics.py
```diff
@@ -102,6 +102,7 @@ def calculate_metrics_for_timeseries(
         MetricsFields.INFECTION_RATE_CI90: infection_rate_ci90,
     }
     metrics = pd.DataFrame(top_level_metrics_data)
+    metrics.index.name = CommonFields.DATE
```
TomGoBravo, line 105: It looks like, at least at this point in the code, the index of `metrics` are just integers, not dates. So it looks add setting the index name.

PR #649, opened 2020-08-27
Add latest icu covid and non-covid patients to details

ghop02:
The website needs these values to show stats.  Helps us remove all of the logic from the webapp


libs/icu_headroom_metric.py
```diff
@@ -181,6 +181,9 @@ def calculate_icu_utilization_metric(
     metric = current_icu_covid / (icu_data.total_icu_beds - non_covid_patients)
 
     details = can_api_definition.ICUHeadroomMetricDetails(
-        currentIcuCovidMethod=covid_source, currentIcuNonCovidMethod=non_covid_source
+        currentIcuCovidMethod=covid_source,
+        currentIcuCovid=current_icu_covid[current_icu_covid.last_valid_index()],
+        currentIcuNonCovidMethod=non_covid_source,
+        currentIcuNonCovid=non_covid_patients[non_covid_patients.last_valid_index()],
```
mikelehen, line 187: FWIW- This has been a bigger issue with contact tracing, but sometimes the sub-series that we aggregate together (e.g. contact tracers and new cases / day) into the series that we then graph (tracers hired) have staggered last data points, and so the website copy can end up including a data point that's newer than what was actually used for the metric value.

It would be nice to fix that rather than carry the bug forward. 😁

ghop02: hmm yeah that's a good point.  I guess we would pick the last non null date in the metric - as a nan in the numerator or denominator will cause the result to be nan.

TomGoBravo (approved):

libs/top_level_metrics.py
```diff
@@ -128,6 +128,32 @@ def calculate_metrics_for_timeseries(
     return metrics, metric_summary
 
 
+def _calculate_smoothed_daily_cases(cases: pd.Series, smooth: int = 7):
+
+    cases = cases.copy()
+
+    # filling gaps in the series to preserve the diff of the first
+    # value after the gap. For example, a cumulative series of [10, None, 30] would result
+    # in a diff of [10, None, None], dropping any signal from the last value.
```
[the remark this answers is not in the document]
ghop02: I removed the comment, but I think this is something to think about - I don't think this case logic is quite right as is, but is closer.  Since I'm not relying on the case values on the frontend just yet, I'd like to merge this and iterate on a future PR.

PR #652, opened 2020-08-30
Run each county in a process

TomGoBravo:

This PR
* factors out a new abstraction: `StatePipeline` and `SubStateRegionPipeline`. They are responsible for running the pipelines and storing the outputs. This is motivated by the need to have the old `_state_only_pipeline` return model_fitter and infer_rt data. It this point we don't need a common subclass so I haven't made one.
* changes how counties (aka sub-state regions) are run. Instead of each stage being run in parallel across all regions, then the next stage, now all counties are run in parallel down the entire pipeline. Hopefully this will be faster due to less data copying and less blocking.
* removes duplicate infer_rt of states, called out by Brett's "Note: double calculates the state fips"
* removes more intermediate files
* extract SubStateRegionPipelineInput construction to a static build_all method
* change ensemble_runner.RegionalInput to get state data as a parameter instead of from disk and remove from_fips which no longer works for substate/county regions
* rm model_fitter.run_counties_of_state that wasn't used in the main pipeline run.


pyseir/cli.py
```diff
@@ -74,21 +62,97 @@ def _states_region_list(state: Optional[str], default: List[str]) -> List[pipeli
...
-def _state_only_pipeline(
-    region: pipeline.Region, run_mode=DEFAULT_RUN_MODE, output_interval_days=1, output_dir=None,
-) -> model_fitter.ModelFitter:
-    assert region.is_state()
+@dataclass
+class StatePipeline:
+    """Runs the pipeline for one state and stores the output."""
+
+    region: pipeline.Region
+    infer_df: pd.DataFrame
```
ghop02, line 70: I know this is the existing name, but maybe something like `inference_result_df` might be more clear

pyseir/cli.py
```diff
@@ -74,21 +62,97 @@ def _states_region_list(state: Optional[str], default: List[str]) -> List[pipeli
...
+
+@dataclass
+class SubStateRegionPipelineInput:
+    region: pipeline.Region
+    run_fitter: bool
+    state_fitter: model_fitter.ModelFitter
+    run_mode: pyseir.utils.RunMode
+
+    @staticmethod
+    def build_all(
```
ghop02, line 92: Maybe an opportunity to change this name to something more specific?

pyseir/cli.py
```diff
@@ -74,21 +62,97 @@ def _states_region_list(state: Optional[str], default: List[str]) -> List[pipeli
...
+        fips: Optional[str],
```
ghop02, line 93: This feels like it should be a keyword arg?

pyseir/cli.py
```diff
@@ -74,21 +62,97 @@ def _states_region_list(state: Optional[str], default: List[str]) -> List[pipeli
...
+        whitelist_df = _generate_whitelist()
+        all_county_regions = set(
+            whitelist.regions_in_states(
+                list(state_fitter_map.keys()), fips=fips, whitelist_df=whitelist_df
+            )
+        )
+        pipeline_inputs = [
+            SubStateRegionPipelineInput(
+                region=region,
+                run_fitter=(region in all_county_regions),
```
ghop02, line 121: all_county_regions is a bit unclear to me.   Could you add a bit of context to it?

pyseir/cli.py
```diff
@@ -74,21 +62,97 @@ def _states_region_list(state: Optional[str], default: List[str]) -> List[pipeli
...
+@dataclass
+class StatePipeline:
+    """Runs the pipeline for one state and stores the output."""
+
+    region: pipeline.Region
+    infer_df: pd.DataFrame
+    fitter: model_fitter.ModelFitter
+
+    @staticmethod
+    def run(region: pipeline.Region, run_mode: pyseir.utils.RunMode) -> "StatePipeline":
```
BrettBoval, line 74: We can remove the RunMode enum-like object. That optionality is never exercised in the snapshot generation.

pyseir/cli.py
BrettBoval: Maybe drop the `inference` reference and use `infection_rate_result_df` instead. Make it more clear that it's supporting that metric?

pyseir/cli.py
```diff
@@ -74,21 +62,97 @@ def _states_region_list(state: Optional[str], default: List[str]) -> List[pipeli
...
+            )
+        )
+        pipeline_inputs = [
+            SubStateRegionPipelineInput(
+                region=region,
+                run_fitter=(region in all_county_regions),
+                state_fitter=state_fitter_map.get(region.get_state_region()),
+                run_mode=run_mode,
+            )
+            for region in (infer_rt_regions | all_county_regions)
```
BrettBoval, line 125: I don't think the union of these sets is the logic we want here. For example, the pipeline should route the regions representing MA counties to the SEIR_model module, but not to the infection_rate module.

TomGoBravo: There was a bug in my PR when you looked at it: `SubStatePipeline.run` wasn't respecting `SubStateRegionPipelineInput.run_fitter`. It is also possible the code that sets run_fitter (just above where you added this comment) isn't clear.

pyseir/cli.py
TomGoBravo: Done

pyseir/cli.py
TomGoBravo: Done

pyseir/cli.py
TomGoBravo: I'm hoping that being a static method of `SubStateRegionPipelineInput` is enough of a hint that it builds a bunch of SubStateRegionPipelineInput objects. I'm following a general pattern of a dataclass with immutable attributes, some static methods that act as constructors and some other methods that may wrap access to some attributes.

pyseir/cli.py
TomGoBravo: Added a comment. I'm trying to create a minimal PR that does this pipeline refactor... and it isn't very small.

pyseir/cli.py
TomGoBravo: Merged in https://github.com/covid-projections/covid-data-model/pull/654 in hopes of making this PR smaller.

TomGoBravo:
Converted to draft until New Orleans is fixed. See https://covidactnow.slack.com/archives/C0110QQAPJA/p1598920106072200

pyseir/cli.py
```diff
@@ -52,114 +56,190 @@ def _generate_whitelist() -> pd.DataFrame:
...
+    def population(self) -> float:
+        return self._combined_data.get_us_latest()[CommonFields.POPULATION]
 
-    with Pool(maxtasksperchild=10) as p:
-        results = pd.concat(p.map(infer_rt.run_rt, regions))
 
-    if results.empty:
-        return pd.DataFrame()
+def _patch_substatepipeline_nola_infection_rate(
+    input_pipelines: List[SubStatePipeline],
```
ghop02, line 178: nit: input is a bit misleading - makes me immediately think that it's the inputs into the pipelines (i later relized that's not the case, but was my first reaction)

pyseir/cli.py
```diff
@@ -52,114 +56,190 @@ def _generate_whitelist() -> pd.DataFrame:
...
+) -> List[SubStatePipeline]:
+    """Returns a new list of pipeline objects with New Orleans infection rate patched."""
+    pipeline_map = {p.region.fips: p for p in input_pipelines}
 
-    need_patch = set(results.fips) & set(NEW_ORLEANS_FIPS)
+    input_fips = set(pipeline_map.keys())
+    need_patch = input_fips & set(NEW_ORLEANS_FIPS)
```
ghop02, line 184: super small, `fips_to_patch` is more clear to me reading through

pyseir/cli.py
```diff
@@ -52,114 +56,190 @@ def _generate_whitelist() -> pd.DataFrame:
...
     if need_patch:
         logging.info("Applying New Orleans Patch")
         if len(need_patch) != len(NEW_ORLEANS_FIPS):
             logging.warning(
```
ghop02, line 188: looks like we're using the logging module here, can you change this to use `root`

pyseir/cli.py
```diff
@@ -52,114 +56,190 @@ def _generate_whitelist() -> pd.DataFrame:
...
     # do everything for just states in parallel
     with Pool(maxtasksperchild=1) as p:
-        states_only_func = partial(
-            _state_only_pipeline, output_interval_days=output_interval_days, output_dir=output_dir,
-        )
         states_regions = [pipeline.Region.from_state(s) for s in states]
-        state_fitters: List[model_fitter.ModelFitter] = p.map(states_only_func, states_regions)
-        state_fitter_map = {f.region: f for f in state_fitters}
+        state_pipelines: List[StatePipeline] = p.map(StatePipeline.run, states_regions)
+        state_fitter_map = {p.region: p.fitter for p in state_pipelines}
```
ghop02, line 226: I know they're technically not in the same scope, but might be worth changing the pool above to be named `pool` rather than `p` to disambiguate

pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -70,20 +71,11 @@ def load_inference_result(self) -> Mapping[str, Any]:
         : dict
             Dictionary of fit result information.
         """
-        if self._mle_fit_result is not None:
-            return self._mle_fit_result
-        return pipeline.load_inference_result(self.region)
+        return self._mle_fit_result
 
     def load_ensemble_results(self) -> Optional[dict]:
```
ghop02, line 76: should we change these to just be read only properties, now that they're not loading: 
```
@property
def ensemble_results(self):
    return self._ensemble_results
```

test/infer_rt_test.py
```diff
@@ -258,6 +252,33 @@ def test_generate_infection_rate_with_nans():
...
     regions = [infer_rt.RegionalInput.from_fips(region) for region in FIPS]
-    df = cli._generate_infection_rate_metric(regions)
+    df = pd.concat(infer_rt.run_rt(input) for input in regions)
     returned_fips = df.fips.unique()
     assert "25001" in returned_fips
+
+
+@pytest.mark.slow
+def test_patch_substatepipeline_nola_infection_rate():
+    FIPS = [
```
ghop02, line 262: would prefer `FIPS` -> `all_fips` since generally all caps are global constants

ghop02 (approved):
A handful of small comments, but overall looks good!

pyseir/cli.py
```diff
@@ -52,114 +56,190 @@ def _generate_whitelist() -> pd.DataFrame:
...
-    with Pool(maxtasksperchild=10) as p:
-        results = pd.concat(p.map(infer_rt.run_rt, regions))
 
-    if results.empty:
-        return pd.DataFrame()
+def _patch_substatepipeline_nola_infection_rate(
+    input_pipelines: List[SubStatePipeline],
+) -> List[SubStatePipeline]:
+    """Returns a new list of pipeline objects with New Orleans infection rate patched."""
+    pipeline_map = {p.region.fips: p for p in input_pipelines}
```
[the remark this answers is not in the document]
TomGoBravo: Done

pyseir/cli.py
TomGoBravo: changed to `pipelines`

pyseir/cli.py
TomGoBravo: Done

pyseir/cli.py
TomGoBravo: Oops. Done

pyseir/cli.py
TomGoBravo: Done

test/infer_rt_test.py
TomGoBravo: Done

pyseir/deployment/webui_data_adaptor_v1.py
TomGoBravo: Renamed a bunch of these. I'm still not sold on the benefit of making a function call look like an attribute access.

pyseir/cli.py
```diff
@@ -52,114 +56,190 @@ def _generate_whitelist() -> pd.DataFrame:
...
+
+    @staticmethod
+    def run(region: pipeline.Region) -> "StatePipeline":
+        assert region.is_state()
+        infer_df = infer_rt.run_rt(infer_rt.RegionalInput.from_region(region))
+        fitter_input = model_fitter.RegionalInput.from_state_region(region)
+        fitter = model_fitter.ModelFitter.run_for_region(fitter_input)
+        ensembles_input = ensemble_runner.RegionalInput.for_state(fitter)
+        ensemble = ensemble_runner.make_and_run(ensembles_input)
+        return StatePipeline(region=region, infer_df=infer_df, fitter=fitter, ensemble=ensemble)
```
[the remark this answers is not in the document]
TomGoBravo: Yes, an ...Input and ...Output does make sense. On the other hand the pattern in model_fitter, ensemble_runner etc is a RegionalInput class and some other class that combines the runner and output holder.

PR #657, opened 2020-09-01
move NEW_ORLEANS_FIPS to RT write

TomGoBravo:

* moves NEW_ORLEANS_FIPS to happen in _generate_infection_rate_metric when RT_INFERENCE_RESULT is written.
* adds some more `slow` marks on slow tests.

### Please Check
- [ ] Are tests passing?


pyseir/cli.py
```diff
@@ -56,14 +57,37 @@ def _generate_infection_rate_metric(regions: List[infer_rt.RegionalInput]) -> pd
...
+    if results.empty:
         return pd.DataFrame()
 
+    need_patch = set(results.fips) & set(NEW_ORLEANS_FIPS)
+    if need_patch:
+        logging.info("Applying New Orleans Patch")
+        unpatched = results.loc[~results.fips.isin(need_patch), :]
+
+        # Aggregate the results created so far into one timeseries of metrics in a DataFrame
+        nola_rt_inferences = pyseir.rt.patches.patch_aggregate_rt_results(need_patch)
```
BrettBoval, line 75: I'm not sure this code captures the correct logic if you only pass a subset of the NEW_ORLEANS_FIPS.

The patch should still be the entire NEW_ORLEANS_FIPS set, even if only a subset of those fips are being returned.

`pyseir.rt.patches.patch_aggregate_rt_results(NEW_ORLEANS_FIPS)`

BrettBoval, line 75: Also, can we call in `nola_combined_infection_rate` or something that doesn't use the word inferences. I think we have an over-burdened definition for `inference`.

pyseir/rt/patches.py
```diff
@@ -39,8 +39,9 @@ def f(x):
...
             Rt_MAP_composite=np.average(x["Rt_MAP_composite"], weights=x["population"]),
             Rt_ci95_composite=np.average(x["Rt_ci95_composite"], weights=x["population"]),
+            date=x.name,
         )
 
-        return pd.Series(weighted_data, index=["Rt_MAP_composite", "Rt_ci95_composite"])
+        return pd.Series(weighted_data)
 
-    return combined_df.groupby(combined_df.index).apply(f)
+    return combined_df.groupby(CommonFields.DATE).apply(f)
```
BrettBoval, line 47: I've always found this type of solution of applying a function to a pandas groubpy object hard to reason about. Especially because what is passed to the mapped function changes based on using aggregate vs. apply. So if you have a way to do this that's more easy to reason and read without starting up an interpreter, please let me know.

test/infer_rt_test.py
```diff
@@ -226,6 +229,20 @@ def test_generate_infection_rate_metric_two_aggregate_levels():
     assert "06075" in returned_fips
 
 
+@pytest.mark.slow
+def test_generate_infection_rate_new_orleans_patch():
```
BrettBoval, line 233: It might be worth having a test that makes sure that the patch is calculated for the full set even if only requesting a subset of region.

Although if that involves mocking up some dataclasses that we don't already have structure around, then I'm happy with just adding a warning comment in the patch code noting the intent.

pyseir/cli.py
TomGoBravo: For sure, that was a missing comment. Added as a warning because it is not expected to happen during prod runs.

TomGoBravo: Done

pyseir/rt/patches.py
TomGoBravo: There is https://stackoverflow.com/a/31521177/341400
```
wm = lambda x: np.average(x, weights=df.loc[x.index, "adjusted_lots"])
df.groupby(["contract", "month", "year", "buys"]).agg(adjusted_lots=("adjusted_lots", "sum"),  
                                                      price_weighted_mean=("price", wm))
```
but I find using df.loc in the lambda less clear than groupby apply reading the group.

test/infer_rt_test.py
TomGoBravo: I don't think it is worth complicating the system right now to support a case that isn't expected in production. It'd be nice to run one fips and produce the same value but for now just trying to retire all these intermediate files.

PR #659, opened 2020-09-02
Dont skip output if timeseries is empty

ghop02:
This will add in a few more regions where we might not have data but clicking on them will trigger an error.  Specifically this will add all of the mariana island "counties"

### DIFF
--- a/libs/pipelines/api_pipeline.py
+++ b/libs/pipelines/api_pipeline.py
@@ -13,6 +13,7 @@ from api.can_api_definition import (
     MetricsTimeseriesRow,
     RegionSummaryWithTimeseries,
 )
+from libs import us_state_abbrev
 from libs import dataset_deployer
 from libs import top_level_metrics
 from libs.datasets import CommonFields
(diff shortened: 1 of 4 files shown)


libs/pipelines/api_pipeline.py
```diff
@@ -27,6 +27,10 @@
 PROD_BUCKET = "data.covidactnow.org"
 
 
+def _is_unknown_county(fips: str) -> bool:
+    return len(fips) == 5 and fips.endswith("999")
```
TomGoBravo, line 31: Please move to us_state_abbrev, next to `ABBREV_US_UNKNOWN_COUNTY_FIPS`.

test/libs/api_pipeline_test.py
```diff
@@ -67,3 +67,33 @@ def test_build_api_output_for_intervention(nyc_fips, nyc_model_output_path, tmp_
...
+
+
+def test_output_no_timeseries_rows(nyc_fips, tmp_path):
+    us_latest = combined_datasets.load_us_latest_dataset()
+    us_timeseries = combined_datasets.load_us_timeseries_dataset()
+    latest = us_latest.get_subset(fips=nyc_fips)
+    timeseries = us_timeseries.get_subset(fips=nyc_fips)
+
+    # Clearing out all rows, testing that a region with no rows still has an API output.
+    timeseries.data = timeseries.data.loc[timeseries.data.fips.isna()]
```
TomGoBravo, line 91: What does nyc_fips do? In other words how about `us_timeseries.data.loc[us_timeseries.data.fips.isna()]`?

PR #660, opened 2020-09-02
Fixing JHU failing tests

ghop02:
Recovered was still causing a problem because all values are zero.  There's a more elegant approach, but since we don't actually use them, i'm excluding it as a column


ghop02:
the one failing test is due to incorrect covid county data, going to merge this so the next update data source is clean

PR #663, opened 2020-09-03
separating out running the pipelines from saving to disk

ghop02:
This separates running the pipelines from writing the output of the pipelines.  I was using the `--states-only` command and I realized that it wasn't writing the pipeline output.


TomGoBravo (approved):
Nice!

PR #664, opened 2020-09-03
Dont fill NAs with zeros in WebUI output

ghop02:
The origin of this PR is I was seeing zeros for the latest estimated current icu count in the API.  I looked into it, and saw that the estimated icu series (taken from pyseir webui output) has 0s after the icu calc data ended rather than NAs.  We were treating this as a value and reporting that "current icu covid patients" were 0. 

We were filling nan's with 0s, so it was not possible to tell where the estimated icu value ended when computing the top level metrics.  

I feel pretty okay with this - I think we filled with zeros initially because of the weird output format we were writing. this should help us disambiguate between values with 0 and missing values

### DIFF
--- a/pyseir/deployment/webui_data_adaptor_v1.py
+++ b/pyseir/deployment/webui_data_adaptor_v1.py
@@ -284,7 +284,6 @@ class WebUIDataAdaptorV1:
                 (output_dates >= datetime(month=3, day=3, year=2020))
                 & (output_dates < datetime.today() + timedelta(days=90))
             ]
-            output_model = output_model.fillna(0)
 
             # Fill in results for the Rt indicator.
             rt_results = regional_input.inferred_infection_rate()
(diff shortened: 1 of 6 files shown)


pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -329,14 +328,16 @@ def write_region(self, regional_input: RegionalInput) -> None:
                     schema.FIPS,
                 )
             ]
+            # Casing floats to ints and then replacing filled in zeros with NaN values instead of
+            # propagating zeros.
+            na_int_columns = output_model.loc[:, int_columns].isna()
             output_model.loc[:, int_columns] = output_model[int_columns].fillna(0).astype(int)
+            output_model[na_int_columns] = np.nan
```
TomGoBravo, line 335: this looks like you are using a data frame (with int_columns columns) for bool indexing. Does that work? I don't see it document at https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#boolean-indexing

TomGoBravo (approved):
How'd you test this change?

ghop02:
> How'd you test this change?

Just added a test in the pyseir end to end test

PR #665, opened 2020-09-03
Add function that captures exceptions rather than throwing to pool

ghop02:


pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -109,6 +109,12 @@ def __init__(
         self.include_imputed = include_imputed
         self.output_dir = output_dir
 
+    def write_region_safely(self, regional_input: RegionalInput) -> None:
+        try:
+            self.write_region(regional_input)
+        except Exception:
+            log.exception("Failed to write WebUI output")
```
BrettBoval, line 116: Can you add the context of the regional input to the log so that we can easily pinpoint the fips?

BrettBoval (approved):

PR #668, opened 2020-09-09
Adding icu beds total to icu headroom metadata

ghop02:


api/README.V1.md
```diff
@@ -244,6 +244,7 @@ Details about how the ICU Headroom Metric was calculated.
 | currentIcuCovidMethod    |         | Method used to determine number of current ICU patients with covid.    |
 | currentIcuNonCovid       | integer | Current number of covid patients in icu.                               |
 | currentIcuNonCovidMethod |         | Method used to determine number of current ICU patients without covid. |
+| latestIcuTotalCapacity   | integer | Latest ICU bed total capacity.                                         |
```
mikelehen, line 247: Why is this "latest" instead of "current"? (here and everywhere else...)

libs/icu_headroom_metric.py
```diff
@@ -100,7 +100,7 @@ def _latest_icu_beds(self):
     @property
     def total_icu_beds(self) -> Union[pd.Series, float]:
         timeseries = self._data[CommonFields.ICU_BEDS]
-        latest_icu_beds = self._latest_icu_beds
+        latest_icu_beds = self.latest_icu_beds
 
         if timeseries.any():
             return timeseries.fillna(latest_icu_beds)
```
mikelehen, line 106: FWIW- this latest_icu_beds / total_icu_beds code is kinda' confusing:

1. Why does this return a Series / float union?  Wouldn't it be cleaner to make it always return a series?
2. Why do we fill gaps with the latest value?  That seems kinda' non-sensical and could result in weird jumps, right?  If we really want to try to keep whatever gappy data we have, then we should probably interpolate over the gaps or something?
3. In cases where we have two conceptually similar methods that just differ in their return type (latest_icu_beds and total_icu_beds), it would be nice to name them to match, e.g. total_icu_beds_latest and total_icu_beds_series or something?  Idunno.

mikelehen (approved):
Looks fine minus some naming nits.

PR #672, opened 2020-09-10
Using Regional Input for api generation

ghop02:
Not quite finished yet (need to fix up tests), but this changes iteration for the api to be region based rather than fips based


cli/api.py
```diff
@@ -93,15 +94,18 @@ def generate_api(input_dir, output, summary_output, aggregation_level, state, fi
...
         _logger.info(f"Running intervention {intervention.name}")
-        all_timeseries = api_pipeline.run_on_all_fips_for_intervention(
-            us_latest, us_timeseries, intervention, input_dir
-        )
+        regional_inputs = [
+            api_pipeline.RegionalInput.from_region_and_intervention(region, intervention, input_dir)
+            for region in regions
+        ]
+        _logger.info(f"Loaded {len(regional_inputs)} regions.")
+        all_timeseries = api_pipeline.run_on_all_regional_inputs_for_intervention(regional_inputs)
```
TomGoBravo, line 108: Not sure the right time, but may be eventually nice to run all (region X interventions) at the same time. Or iterate through regions as the outside loop so this can be done as part of the pyseir per-region pipeline. certainly doesn't need to be part of this refactor.

TomGoBravo (commented):
overall, looking good

libs/pipelines/api_pipeline.py
```diff
@@ -29,37 +30,64 @@
...
+    @property
+    def fips(self) -> str:
+        return self.region.fips
+
+    @property
+    def state(self) -> str:
+        return self.latest[CommonFields.STATE]
+
+    @property
+    def latest(self):
```
TomGoBravo, line 51: pls add a return type

libs/pipelines/api_pipeline.py
```diff
@@ -29,37 +30,64 @@
...
+        return self._combined_data.latest
+
+    @property
+    def timeseries(self):
```
TomGoBravo, line 55: pls add a return type

test/libs/api_pipeline_test.py
```diff
@@ -69,31 +62,27 @@ def test_build_api_output_for_intervention(nyc_fips, nyc_model_output_path, tmp_
...
-
-    assert timeseries.empty
-
-    all_timeseries_api = api_pipeline.run_on_all_fips_for_intervention(
-        latest, timeseries, Intervention.OBSERVED_INTERVENTION, tmp_path
+    # Creating a new regional input with an empty timeseries dataset
+    timeseries = regional_input.timeseries
+    timeseries_data = timeseries.data.loc[timeseries.data.fips.isna()]
+    regional_data = combined_datasets.RegionalData(
+        regional_input._combined_data.region,
```
TomGoBravo, line 73: use `nyc_region` or `regional_input.region` instead of single underscore attribute

TomGoBravo (approved):

PR #673, opened 2020-09-11
replace more webui RegionalInput method access with attribute access

TomGoBravo:
replace more webui_data_adaptor_v1.RegionalInput method access with attribute access. See  https://google.github.io/styleguide/pyguide.html#213-properties and https://stackoverflow.com/a/1555169  for rationale.

It looks like I didn't merge https://github.com/covid-projections/covid-data-model/pull/671/files into https://github.com/covid-projections/covid-data-model/pull/656 but not sure why it tested green in github.

### DIFF
--- a/pyseir/deployment/webui_data_adaptor_v1.py
+++ b/pyseir/deployment/webui_data_adaptor_v1.py
@@ -1,14 +1,11 @@
 from dataclasses import dataclass
-from typing import Any
-from typing import Mapping
-from typing import Optional
+from typing import Any, Mapping, Optional
 
 import structlog
 from datetime import timedelta, datetime
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


pyseir/deployment/webui_data_adaptor_v1.py
```diff
@@ -69,7 +70,7 @@ def fips(self) -> str:
         return self.region.fips
 
     @property
-    def latest(self):
+    def latest(self) -> LatestValuesDataset:
```
ghop02, line 73: I think this is actually a dict

ghop02 (approved):

pyseir/deployment/webui_data_adaptor_v1.py
TomGoBravo: Good catch. Fixed.

PR #674, opened 2020-09-11
Added new NowcastingSEIRModel and supporting tests, enabled incremental run on old model

amacaula:
This is mostly new code implementing a new kind of SEIR Model and supporting tests

nowcast_model

Key features of the model
* Resolved differences between SEIR R(t) and Bettencourt determined R(t) so that more precise (no arbitrary step functions) Bettencourt R(t) and ML projections into future can be used to drive this model. In fact this is a requirement.
* Inclusion of explicit I compartment and process (covid testing) to move from I to C. Note this can be turned off (and most tests now do this) by not supplying a testing rate.
* Time varying effects on f_h and f_d, and evolution of, median age of new cases and hospitalizations including the delay where median age of hospitalized cases lags behind that of new cases
* The ability to restart this model from any point in time needing only initial compartment values and R(t) - both historical and predictions for the future
* A simple means to auto calibrate the model to the data ensuring continuous nC, H and nD values (from historical to predictions)

For more information see the [new modeling document](https://docs.google.com/document/d/1OvLZHy6z7QYQQzLySY2cBK6Myo5B19XybkyzgCT3tWk) 

### Please Check
- [x] Are tests passing?


pyseir/models/demographics.py
```diff
@@ -0,0 +1,247 @@
...
+
+    @staticmethod
+    def median_age_f(state=None):
+        """
+        Median age of covid cases for all of US and specific individual states
+        """
+        if state not in ["FL"]:  # median age for country as a whole
+            # From https://www.cdc.gov/coronavirus/2019-ncov/covid-data/covidview/index.html
+            data = {
+                68: 56.3,
```
tashwoods, line 187: Just flagging here that we may want to update these values as we discussed.

pyseir/models/historical_data.py
```diff
@@ -0,0 +1,220 @@
...
+    return (average_R, growth_ratio, adj_r_f)
+
+
+class ForecastData:
+    """
+    Look up forecast data to use for testing purposes.
+    """
+
+    raw = pd.read_csv(FORECAST_DATA_FILE, parse_dates=["date"])
+    data = raw[["fips", "date", "weekly_new_cases_0.5", "weekly_new_deaths_0.5",]]
```
tashwoods, line 193: what is the weekly_new_cases_0.5 mean?

pyseir/models/nowcast_seir_model.py
```diff
@@ -0,0 +1,942 @@
...
+
+    def get_calibration(self):
+        """
+        Get the current calibration which is only using one proportionality factor for each of
+        fh, fd
+        """
+        return (self.lr_fh[0], self.lr_fd[0])
+
+    def run_stationary(self, rt, median_age, t_over_x, x_is_new_cases=True):
+        """
```
tashwoods, line 181: Adding a doc string here with more context and the inputs and outputs would be great

tashwoods:
I think in general just removing unused bits and adding documentation and doc strings would be awesome : ).

pyseir/models/demographics.py
amacaula: Brett has found a bunch more data on the CDC site so I will generalize this better once that analysis is done - but that will be after this PR. Need this merged so integration work can proceed - this code is not yet used in production.

pyseir/models/historical_data.py
amacaula: Added comments explaining the .5 means median value from a set of projections

pyseir/models/nowcast_seir_model.py
amacaula: Added docstring here and in many other places

pyseir/models/historical_data.py
```diff
@@ -0,0 +1,230 @@
...
+import us
+from pyseir.rt.constants import InferRtConstants
+
+# Historical data file, prepared in each API snapshot, including key metrics and inferred Rt values
+# TODO update process to save this periodically in this locaiton
+# TODO find out why VI data appears to be broken in latest file
+HISTORICAL_DATA_FILE = "test/data/historical/merged_results.csv"
+
+# Forecast data
+FORECAST_DATA_FILE = "test/data/forecasts/timeseries-common.csv"
```
ghop02, line 14: @BrettBoval can we move these out of test?

BrettBoval (approved):
The default is to still run the current pyseir model. The intention is to split this PR into multiple blocks, with (1) adding the new code and tests the first block, and (2) then using that code to prototype scenario planners, and (3) having the default values from the new model show up in the API. Right now this is just part 1, which could still theoretically live in a separate branch, but it's been separate long enough that I think it's worth getting it into the modeling, even though it's not running during build-all right now.

PR #677, opened 2020-09-12
add OneRegionTimeseriesDataset

TomGoBravo:
This PR adds a class that is similar to TimeseriesDataset but holds a DataFrame that contains data for exactly one region. This makes it easier to follow how region identifiers/FIPS are passed around within DataFrames.

To mark the class as `@final` in Python 3.7 `typing-extensions` is added to `requirements.txt`


libs/datasets/timeseries.py
```diff
@@ -26,6 +31,50 @@ def __str__(self):
...
+
+    def get_subset(self, after=None, columns=tuple()):
+        rows_key = dataset_utils.make_rows_key(self.data, after=after,)
+        columns_key = list(columns) if columns else slice(None, None, None)
+        return OneRegionTimeseriesDataset(
+            self.data.loc[rows_key, columns_key].reset_index(drop=True)
+        )
+
+    def remove_padded_nans(self, columns: List[str]):
+        return OneRegionTimeseriesDataset(_remove_padded_nans(self.data, columns))
```
ghop02, line 75: How about `self.__class__` rather than calling the static class name (in case of subclassing)?

ghop02 (approved):

libs/datasets/timeseries.py
TomGoBravo: I tried but than we lose static type checking of the parameters used to make the class. I'll try marking it @final https://sobolevn.me/2018/07/real-python-contants

PR #678, opened 2020-09-14
mv pipeline.RegionalCombinedData to combined_datasets

TomGoBravo:

I ran into some gnarly import loops when using `Region` in more places. This refactor will hopefully break them and simplify a followup PR.

* mv pipeline.RegionalCombinedData to combined_datasets.RegionalData
* change access to the class to always use name `combined_datasets.RegionalData`


ghop02 (approved):
LGTM!

PR #680, opened 2020-09-14
Copy External Forecasts from covid-data-public to covid-data-model

BrettBoval:
Paired with https://github.com/covid-projections/covid-data-public/pull/123.

Makes a copy of the external forecasts (currently only the ensemble from ForecastHub) into the data directory of covid-data-model.

The move is currently a simple copy, since no other external forecast streams are currently being combined.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/cli/data.py
+++ b/cli/data.py
@@ -4,6 +4,7 @@ import logging
 import pathlib
 import os
 import json
+import shutil
 
 import click
 
@@ -21,6 +22,7 @@ from libs.datasets import dataset_utils
 from libs.datasets import combined_dataset_utils
 from libs.datasets import combined_datasets
 from libs.datasets.dataset_utils import AggregationLevel
+from libs.datasets.sources import forecast_hub
 from pyseir import DATA_DIR
 import pyseir.icu.utils
 from pyseir.icu import infer_icu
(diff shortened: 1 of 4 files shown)


cli/data.py
```diff
@@ -44,6 +46,17 @@ def _save_field_summary(timeseries_dataset: TimeseriesDataset, output_path: path
     _logger.info(f"Saved dataset summary to {output_path}")
 
 
+@main.command()
+@click.option("--filename", default="external_forecasts.csv")
+def update_forecasts(filename):
+    """Updates external forecasts to the current checked out covid data public commit"""
```
ghop02, line 52: Can you add a log line either at the beginning or end to signify that we're updating forecasts

libs/datasets/sources/forecast_hub.py
```diff
@@ -0,0 +1,24 @@
...
+
+    COMMON_FIELD_MAP = {
+        # Not Yet Implemented -> Currently only a move from covid-data-public to covid-data-model
+    }
+
+    @classmethod
+    def local(cls):
+        data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
+        input_path = data_root / cls.DATA_PATH
+        data = common_df.read_csv(input_path).reset_index()
```
ghop02, line 23: I think there's a command line arg you can pass now (maybe `set_index`) instead of having to reset the index here

ghop02 (approved):

cli/data.py
BrettBoval: Done.

PR #681, opened 2020-09-15
add MultiRegionTimeseriesDataset

TomGoBravo:
* Adds frozen final dataclass `MultiRegionTimeseriesDataset` which replaces `TimeseriesDataset` and reduces places where `data` is directly accessed, making it easier to change how regions are represented within the dataframe.
* Removes `combined_datasets.get_timeseries_for_fips`
* Removes `infer_t0.py`


libs/pipeline.py
```diff
@@ -10,6 +10,13 @@
 import us
 
 
+def fips_to_location_id(fips: str) -> str:
+    """Converts a FIPS code to a locationID"""
+    if len(fips) == 2:
+        iso1: us  # iso2:us-ca#fips:06087
```
ghop02, line 16: This feels incomplete?

libs/datasets/timeseries.py
```diff
@@ -292,6 +282,62 @@ def load_csv(cls, path_or_buf: Union[pathlib.Path, TextIO]):
         return cls(df)
 
 
+@final
+@dataclass(frozen=True)
+class MultiRegionTimeseriesDataset:
```
ghop02, line 287: The timeseries.py file is getting a bit crowded, thoughts on when we should separate the existing classes into multiple files?

ghop02 (approved):
seems fine to me right now!

libs/pipeline.py
TomGoBravo: oops, that's from another branch. fixed with 2a8fcdd

libs/datasets/timeseries.py
TomGoBravo: I'd like to replace the last of TimeseriesDataset with the more static and specific OneRegion/MultiRegionTimeseriesDataset, shrinking this file back down a lot.

PR #683, opened 2020-09-16
Use location_id multiregion combined data and add CBSA aggregator

TomGoBravo:

* Adds `CountyToCBSAAggregator` and test which is first attempt at aggregating counties to MSAs.
* Removes unused `libs.qa.dataset_summary.get_summaries`
* Adds `xlrd` to requirements.txt
* Changes combined_dataset to be saved as a MultiRegionTimeseriesDataset with location_id instead of TimeseriesDataset with FIPS.
* removes save_summary parameter `--level` because it isn't reason enough to expand the API of MultiRegionTimeseriesDataset 
* Adds dataset_base.SaveableDatasetInterface which sits under DatasetBase in the class hierarchy with stuff it has in common with MultiRegionTimeseriesDataset

## Tested

I ran `python ./run.py data update` and it regenerated an identical data/timeseries.csv. I then skimmed through `vimdiff data/timeseries.csv data/multiregion.csv` to check that the new data file looks the same except for adding a location_id column.


libs/pipeline.py
```diff
@@ -8,26 +8,49 @@
...
+        state_obj = us.states.lookup(fips, field="fips")
+        return f"iso1:us#iso2:us-{state_obj.abbr}"
+    elif len(fips) == 5:
+        state_obj = us.states.lookup(fips[0:2], field="fips")
+        return f"iso1:us#iso2:us-{state_obj.abbr}#fips:{fips}"
+    else:
+        raise ValueError("Bad fips")
+
+
+def cbsa_to_location_id(cbsa_code: str) -> str:
```
ghop02, line 26: Can you add a docstring to this, just giving a bit more context about the cbsa code

libs/pipeline.py
```diff
@@ -8,26 +8,49 @@
...
     def from_state(state: str) -> "Region":
         """Creates a Region object from a state abbreviation, name or 2 digit FIPS code."""
         state_obj = us.states.lookup(state)
         fips = state_obj.fips
-        return Region(fips=fips)
+        return Region.from_fips(fips)
+
+    @staticmethod
+    def from_cbsa_code(cbsa_code: str) -> "Region":
+        return Region(fips="", location_id=cbsa_to_location_id(cbsa_code))
```
ghop02, line 53: Does fips being optional break too many things?

libs/datasets/statistical_areas.py
```diff
@@ -0,0 +1,55 @@
...
+import pandas as pd
+
+from libs import pipeline
+from libs.datasets import timeseries
+from libs.datasets.timeseries import MultiRegionTimeseriesDataset
+from covidactnow.datapublic.common_fields import CommonFields
+from libs.datasets import dataset_utils
+
+
+CBSA_LIST_PATH = "data/census-msa/list1_2020.xls"
```
ghop02, line 12: it would be great if this was not an excel file + CBSA_COLUMN was maybe a common field?

libs/datasets/timeseries.py
```diff
@@ -297,14 +311,17 @@ class MultiRegionTimeseriesDataset:
     @staticmethod
     def from_csv(path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionTimeseriesDataset":
         df = common_df.read_csv(path_or_buf, set_index=False)
+        _add_location_id(df)
```
ghop02, line 314: Would it be possible to add this in at a single point? maybe in `from_source_data` or when we're constructing combined datasest?  I would love to be in a position where it was already included in the saved csv instead of having to mutate the loaded csv

libs/datasets/statistical_areas.py
```diff
@@ -0,0 +1,59 @@
...
+        # Make a copy to avoid modifying the input. DataFrame.assign is an alternative but the API
+        # doesn't work well here.
+        df = dataset_in.data.copy()
+        df[CBSA_COLUMN] = df[CommonFields.FIPS].map(self.county_map)
+
+        # TODO(tom): Put the title in the data.
+        df_cbsa = df.groupby([CBSA_COLUMN, CommonFields.DATE], as_index=False).sum()
+        df_cbsa[CommonFields.LOCATION_ID] = df_cbsa[CBSA_COLUMN].apply(pipeline.cbsa_to_location_id)
+
+        merged = pd.concat([df, df_cbsa]).drop(columns=CBSA_COLUMN)
```
ghop02, line 36: How would you feel about not recombining the data here - seems a bit simpler to me if the output of this function is a CSBA aggregated dataset, instead of being recombined with the input

libs/datasets/statistical_areas.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+CBSA_LIST_PATH = "data/census-msa/list1_2020.xls"
+
+
+CBSA_COLUMN = "CBSA"
+
+
+@dataclass
+class CountyAggregator:
```
ghop02, line 19: May be helpful to mention aggregating to CBSA either in the class name or the `aggregate` function

libs/datasets/statistical_areas.py
```diff
@@ -0,0 +1,59 @@
...
+    cbsa_title_map: Mapping[str, str]
+
+    def aggregate(self, dataset_in: MultiRegionTimeseriesDataset) -> MultiRegionTimeseriesDataset:
+        # Make a copy to avoid modifying the input. DataFrame.assign is an alternative but the API
+        # doesn't work well here.
+        df = dataset_in.data.copy()
+        df[CBSA_COLUMN] = df[CommonFields.FIPS].map(self.county_map)
+
+        # TODO(tom): Put the title in the data.
+        df_cbsa = df.groupby([CBSA_COLUMN, CommonFields.DATE], as_index=False).sum()
```
ghop02, line 33: Are you planning on adding the logic for metric aggregation here? i.e. we may not want to sum current hospitalizations if one area in the group doesn't have them

libs/datasets/statistical_areas.py
TomGoBravo: I don't think it is worth the complexity of introducing yet another intermediate representation to simplify what is currently 10 lines of code and unlikely to change for at least a few months.
I don't want access/dependencies on the CBSA_COLUMN to creep in all over our code so trying to keep it isolated to this module.

libs/pipeline.py
TomGoBravo: Made Optional.

libs/datasets/timeseries.py
TomGoBravo: I made the change to put location_id in the combined datasets. I really like the idea that the saved csv and in-memory representation are separate, mutating when loading is very inexpensive and as long as it is a 1-1 transformation between representations doesn't reduce our observalibity for debugging much. I was motivated by a desire to move location_id further upstream.

libs/datasets/statistical_areas.py
TomGoBravo: Yes, extended the TODO. For now just trying to get all the parts wired up.

libs/datasets/statistical_areas.py
TomGoBravo: Done

libs/datasets/statistical_areas.py
TomGoBravo: Done

libs/pipeline.py
TomGoBravo: Done.

libs/datasets/timeseries.py
```diff
@@ -282,29 +296,55 @@ def load_csv(cls, path_or_buf: Union[pathlib.Path, TextIO]):
...
+            raise ValueError("MultiRegionTimeseriesDataset.from_csv requires location_id column")
+        _add_fips_if_missing(df)
         return MultiRegionTimeseriesDataset(df)
 
     @staticmethod
     def from_timeseries(ts: TimeseriesDataset) -> "MultiRegionTimeseriesDataset":
-        return MultiRegionTimeseriesDataset(ts.data, provenance=ts.provenance)
+        df = ts.data.copy()
+        _add_location_id(df)
+        return MultiRegionTimeseriesDataset(df, provenance=ts.provenance)
```
ghop02, line 344: Should provenance be by location id too?

ghop02 (approved):
LGTM! Thanks for all of the work on this!

libs/datasets/timeseries.py
TomGoBravo: Good point. Added test and fixed.

PR #692, opened 2020-09-22
Clean up data dir build

ghop02:
This removes the api result folder after the build is finished


TomGoBravo (approved):
Please add comment about why cleanup matters. I thought this runs on a temporary virtual machine or docker instance or something like that.

PR #693, opened 2020-09-23
Change state output filename

ghop02:


libs/pipelines/api_v2_paths.py
```diff
@@ -64,9 +64,20 @@ def bulk_prediction_data(
         return self.root / f"{self.region_key}.{file_type.suffix}"
 
     def single_summary(self, region_summary: can_api_v2_definition.RegionSummary, file_type):
-        return self.region_subdir / f"{region_summary.fips}.{file_type.suffix}"
+        if self.level is AggregationLevel.STATE:
+            return self.region_subdir / f"{region_summary.state}.{file_type.suffix}"
+        if self.level is AggregationLevel.COUNTY:
```
TomGoBravo, line 69: nit: I prefer if ... elif .... else to make expected flow explicit.

TomGoBravo (approved):
I can't wait to reduce our dependency on using these files for debugging.

PR #695, opened 2020-09-23
combined_datasets raises when region not found in latest values data

TomGoBravo:
Excluding the possibility of a Region without latest values makes a future change easier.


libs/datasets/combined_datasets.py
```diff
@@ -302,6 +302,9 @@ def from_region(region: Region) -> "RegionalData":
         us_timeseries = load_us_timeseries_dataset()
         region_timeseries = us_timeseries.get_one_region(region)
 
+        if region_timeseries.empty or not region_latest:
+            raise KeyError(f"No combined data found for {region}")
```
ghop02, line 306: I think this might break some other parts of the pipeline - I think a lot of places where we iterate over all regions this will cause an exception adn break?

Also, I don't know if `KeyError` is quite appropriate, since generally it's when a key is not in a dictionary

TomGoBravo: In theory all the places that iterate over all regions get their list of regions from combined_datasets so this won't happen. :crossed_fingers: 

pandas uses `KeyError` when a label isn't found so this simplifies the next change, but easy to raise something else too.

ghop02: I think as it currently is, this will break the API:
 `./run.py api generate-api-v2 results/` (will take a couple minutes + not work I believe)

TomGoBravo: Indeed. scaled back to only raising when latest values isn't found

ghop02 (approved):

PR #696, opened 2020-09-23
Adding api auth

ghop02:
This is preliminary code involved for handling API auth.

Components:
* Uses two lambda functions: one to register emails and provide a key, another to check that the API key included in the request is valid.
* Stores email/api key pairs in a DynamoDB table
* Uses the serverless framework to deploy lambda functions.  lets us easily create different "stages" or envrionments so we can develop locally and not modify the prod api until ready. 

This hooks up to our API Gateway which is pretty manually put together.  I'd love to add that setup to the cloudformation resources, but will take a bit too much time right now.  

Since this code is deployed independently, I don't think it quite makes sense to have it integrate with the rest of the setup (i.e. share requirements, maybe live in libs).  Right now there's no testing for this code.  Open to suggestions.  Would love to hook this up to more of an automated deploy (deploy code to dev state, run an integration test, then deploy to prod) but don't have that hooked up right now. 

I'd love a review on the code itself and definitely open to discussing the more operational components

### DIFF
--- a/api/awsauth/api_key_repo.py
+++ b/api/awsauth/api_key_repo.py
@@ -1,0 +1,65 @@
+from typing import Optional, Dict, Any
+import os
+
+import dynamo_client
+
+
+API_KEY_TABLE_NAME = os.environ["API_KEYS_TABLE"]
+API_KEY_INDEX_NAME = "apiKeys"
+
+
+class APIKeyRepo:
+    """Logic for handling API Key functionality."""
+
+    @staticmethod
+    def add_api_key(email: str, api_key: str):
+        """Adds an email and api_key to database.
+
+        Args:
+            email: Email Address.
+            api_key: API Key to add.
+        """
+        # TODO: Client should only be initialized once
+        client = dynamo_client.DynamoDBClient()
+        obj = {"email": email, "api_key": api_key}
+        client.put_item(API_KEY_TABLE_NAME, obj)
+
+    @staticmethod
+    def get_api_key(email) -> Optional[str]:
+        """Fetches API Key for email, returning none if not found.
+
+        Args:
+            email: Email address.
+
+        Returns: API Key if exists, None if missing.
+        """
+        # TODO: Client should only be initialized once
+        client = dynamo_client.DynamoDBClient()
+
+        key = {"email": email}
+        api_key_item = client.get_item(API_KEY_TABLE_NAME, key)
+
+        if not api_key_item:
+            return None
+
+        return api_key_item["api_key"]
+
+    @staticmethod
+    def get_record_for_api_key(api_key) -> Optional[Dict[str, Any]]:
+        """Fetches record for API Key.
+
+        Args:
+            api_key: API Key to query.
+
+        Returns: Dict if exists, None if not found.
+        """
+        # TODO: Client should only be initialized once
+        client = dynamo_client.DynamoDBClient()
+        items = client.query_index(API_KEY_TABLE_NAME, API_KEY_INDEX_NAME, "api_key", api_key)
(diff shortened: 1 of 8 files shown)


api/awsauth/auth_app.py
```diff
@@ -0,0 +1,95 @@
...
+
+    api_key = _get_or_create_api_key(email)
+    body = {"api_key": api_key, "email": email}
+
+    response = {"statusCode": 200, "body": json.dumps(body)}
+
+    return response
+
+
+def _generate_accept_policy(user_record: dict, method_arn):
```
TomGoBravo, line 67: type of method_arn?

api/awsauth/dynamo_client.py
```diff
@@ -0,0 +1,53 @@
...
+    def put_item(self, table: str, item: Dict[str, Any]) -> None:
+        """Add item to table.
+
+        Args:
+            table: Table name.
+            item: Item to add in form {'<attribute_name>': <attribute_value>, ...}.
+
+        """
+        table = self._client.Table(table)
+        table.put_item(Item=item)
```
TomGoBravo, line 53: How about something like ConditionExpression='attribute_not_exists(email)'  to avoid duplicates if two requests are racing to create a new item?

api/awsauth/api_key_repo.py
```diff
@@ -0,0 +1,65 @@
...
+import os
+
+import dynamo_client
+
+
+API_KEY_TABLE_NAME = os.environ["API_KEYS_TABLE"]
+API_KEY_INDEX_NAME = "apiKeys"
+
+
+class APIKeyRepo:
```
TomGoBravo, line 11: I don't get the value of this layer of abstraction between the dynamo_client (written just for this app, entirely in our control) and auth_app.

api/awsauth/api_key_repo.py
```diff
@@ -0,0 +1,65 @@
...
+        Returns: Dict if exists, None if not found.
+        """
+        # TODO: Client should only be initialized once
+        client = dynamo_client.DynamoDBClient()
+        items = client.query_index(API_KEY_TABLE_NAME, API_KEY_INDEX_NAME, "api_key", api_key)
+        if not items:
+            return None
+
+        if len(items) > 1:
+            raise Exception("Multiple emails found for API key")
```
TomGoBravo, line 63: maybe API key is a better primary key than email... this case should never happen if the uuid works but also doesn't need to be fatal.

TomGoBravo (approved):
Is this it?!

api/awsauth/auth_app.py
```diff
@@ -0,0 +1,95 @@
...
+import json
+import logging
+
+from api_key_repo import APIKeyRepo
+
+
+_logger = logging.getLogger(__name__)
+
+
+EMAIL_REGEX = "^[a-z0-9]+[\._]?[a-z0-9]+[@]\w+[.]\w{2,3}$"
```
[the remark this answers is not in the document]
ghop02: ah good question - I just took it from https://www.geeksforgeeks.org/check-if-email-address-valid-or-not-in-python/

api/awsauth/api_key_repo.py
ghop02: I've always appreciated having the layer of abstraction to separate the implementation details of using dynamo db vs more abstract requests "I would like an api key". I've also found it easier to mock out responses at a not client level.  

IMO it gives a bit more room to make the distinction between the business logic + what's actually happening behind the scenes (+ if we were to ever switch up detail of the implementation, all of the logic is contained in the repo)

PR #701, opened 2020-09-24
Adding massachussetts counties back in

ghop02:
Mass counties now seem to work! Adding back in

### DIFF
--- a/pyseir/cli.py
+++ b/pyseir/cli.py
@@ -106,10 +106,7 @@ class SubStateRegionPipelineInput:
         else:  # Default to the full infection rate whitelist
             infer_rt_regions = {
                 combined_datasets.get_subset_regions(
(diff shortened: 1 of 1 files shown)


BrettBoval (approved):
These regions were originally filtered due to non-reporting of data for approx 8 days. These were originally returned as "0" values for incident new cases, which was causing a large error towards zero. That has been changed to reporting a nan value. The averaging kernel width is greater than the break in reporting so the smoothed value is contiguous, although it's still an open question how we should handle gaps this large (which should remain rare).

MA counties have resumed more regular reporting and that non-reporting anomaly is old enough to have passed out of the buffer used to calculate the most recent infection metric rates.

Therefore, it's worth re-activating the infection rate metric for MA counties, even with the caveat that mid-August historical infection rate values will be less reliable.

ghop02:
Thanks for the additional details!

PR #706, opened 2020-09-25
Introduce a new_cases column to the MultiRegionTimeseriesDataset

jabalsad:
Introduce a new_cases column to the MultiRegionTimeseriesDataset.

This change calculates a new_cases column from consecutive rows in the cases column. It also groups the dataset by location_id to avoid cross-counting different locations.


ghop02:
> _THIS IS A PROTOTYPE / NOT FINAL_
> 
> I'm opening this PR to facilitate a discussion around introducing a new cases column to the data sets. The current implementation is very rudimentary and just serves as a proof of concept.
> 
> I'd like to discuss the following questions:
> 
> 1. Is this the right place where we will be transforming the dataset? The implication is that only `MultiRegionTimeseriesDataset` will have this column included.

I think we should be doing the transformation outside of the class.  My thinking is that it should actually be an additiona line either [here](https://github.com/covid-projections/covid-data-model/blob/658e1267af68c8951d5260133f5fe72aaef20fb7/cli/data.py#L81) or [here](https://github.com/covid-projections/covid-data-model/blob/658e1267af68c8951d5260133f5fe72aaef20fb7/cli/data.py#L82)

maybe something like:
```python
timeseries_dataset: TimeseriesDataset = combined_datasets.build_from_sources(
    TimeseriesDataset, data_sources, ALL_TIMESERIES_FEATURE_DEFINITION, filter=US_STATES_FILTER
)
timeseries_dataset = add_new_cases(timeseries_dataset)
```


```python
multiregion_dataset = MultiRegionTimeseriesDataset.from_timeseries(timeseries_dataset)
multiregion_dataset = add_new_cases(multiregion_dataset)
```
but I'd also be interested in @TomGoBravo's thoughts there.

> 2. How general do we want the mechanism to be? For example, we could create a `Transform` class in the same way how Filters are currently implemented. Perhaps the `Transform` could be applied to any `DatasetBase`? This is certainly more work but more robust long term.

I don't have a strong opinion on the mechanism - my inclination is to start out simple and if we have 2 or 3 cases we can make a more robust solution.

> 3. How do we deal with edge cases? For example, when one row has a cases value of Nil then the resultant diff will also be zero (regardless of the fact that there may have been an increase on this day). Generally, this happens early on the first day in the timeseries where the cohort is reporting data.

That's a very good question.  

> 4. Another edge case worth considering, though I'm not entirely sure it is present in the data, is that in theory, the current implementation could calculate the difference between two rows from different cohorts (e.g. where the states, counties, or countries are not equal). The current implementation doesn't allow for that (I will need to do more research into panda to understand its limitations with regards to performing operations based on constraints). Due to the way the dataset is currently sorted, this doesn't seem to be a problem (but that's purely by chance).

Yeah this is a good point.  I think you'll want to apply a groupby on the location to calculate this by region.  You are right that it does work but is also purely by chance.  

There are a couple examples I can point you to.


> 5. What, if any, are the performance concerns here? I'm not entirely sure what performance boundaries we're operating under.

Don't think there are really any strong performance concerns! Ideally it's snappy - I would expect this to just take a couple of seconds - but unless it's taking more than a minute i wouldn't be too concerned

jabalsad:
Sorry one more question I forgot to ask in the original description:

6. Are we okay with reusing `CommonFields.NEW_CASES` or should we consider a new column name, e.g. `CommonFields.CALCULATED_NEW_CASES` or similar? I wasn't sure if that will cause problems if one of the upstream datasources start importing a column under the same name.

ghop02:
> Sorry one more question I forgot to ask in the original description:
> 
> 1. Are we okay with reusing `CommonFields.NEW_CASES` or should we consider a new column name, e.g. `CommonFields.CALCULATED_NEW_CASES` or similar? I wasn't sure if that will cause problems if one of the upstream datasources start importing a column under the same name.

`CommonFields.NEW_CASES` I think is good!

TomGoBravo:
> > _THIS IS A PROTOTYPE / NOT FINAL_
> > I'm opening this PR to facilitate a discussion around introducing a new cases column to the data sets. The current implementation is very rudimentary and just serves as a proof of concept.
> > I'd like to discuss the following questions:
> > 
> > 1. Is this the right place where we will be transforming the dataset? The implication is that only `MultiRegionTimeseriesDataset` will have this column included.
> 
> I think we should be doing the transformation outside of the class. My thinking is that it should actually be an additiona line either [here](https://github.com/covid-projections/covid-data-model/blob/658e1267af68c8951d5260133f5fe72aaef20fb7/cli/data.py#L81) or [here](https://github.com/covid-projections/covid-data-model/blob/658e1267af68c8951d5260133f5fe72aaef20fb7/cli/data.py#L82)
> 
> maybe something like:
> 
> ```python
> timeseries_dataset: TimeseriesDataset = combined_datasets.build_from_sources(
>     TimeseriesDataset, data_sources, ALL_TIMESERIES_FEATURE_DEFINITION, filter=US_STATES_FILTER
> )
> timeseries_dataset = add_new_cases(timeseries_dataset)
> ```
> 
> ```python
> multiregion_dataset = MultiRegionTimeseriesDataset.from_timeseries(timeseries_dataset)
> multiregion_dataset = add_new_cases(multiregion_dataset)
> ```
> 
> but I'd also be interested in @TomGoBravo's thoughts there.

A function or method with MultiRegionTimeseriesDataset parameter and MultiRegionTimeseriesDataset result makes sense to me.

> 
> > 1. How general do we want the mechanism to be? For example, we could create a `Transform` class in the same way how Filters are currently implemented. Perhaps the `Transform` could be applied to any `DatasetBase`? This is certainly more work but more robust long term.

`DatasetBase` is an abstraction that represents timeseries and non-timeseries data separately. I created MultiRegionTimeseriesDataset because I think these two types are instead often passed around together and treated separately. Use MultiRegionTimeseriesDataset, I'm trying to retire the others.

> I don't have a strong opinion on the mechanism - my inclination is to start out simple and if we have 2 or 3 cases we can make a more robust solution.
> 
> > 1. How do we deal with edge cases? For example, when one row has a cases value of Nil then the resultant diff will also be zero (regardless of the fact that there may have been an increase on this day). Generally, this happens early on the first day in the timeseries where the cohort is reporting data.
> 
> That's a very good question. I think there are a couple details to iron out. For this PR I think it would be best just to get some super simple code in place and then deal with a couple of these edge cases.

Agreed with Chris on getting everything wired up, then turn your focus to handling the corner cases well.

> > 1. Another edge case worth considering, though I'm not entirely sure it is present in the data, is that in theory, the current implementation could calculate the difference between two rows from different cohorts (e.g. where the states, counties, or countries are not equal). The current implementation doesn't allow for that (I will need to do more research into panda to understand its limitations with regards to performing operations based on constraints). Due to the way the dataset is currently sorted, this doesn't seem to be a problem (but that's purely by chance).
> 
> Yeah this is a good point. I think you'll want to apply a groupby on the location to calculate this by region. You are right that it does work but is also purely by chance.

cohorts, as in subsets of regions grouped by similarity in the data? Maybe not worth going into lots of detail on corner cases since focus right now is getting some data flowing.

Agreed that using `CommonFields.NEW_CASES` is fine. One day we'll have provenance information to differentiate synthesized and original NEW_CASES timeseries.

ghop02:
My only hesitation on a method on `MultiRegionTimeseriesDataset` is that the dataset doesn't explicitly contain case data.  it's more of a wrapper in that sense

TomGoBravo:
> My only hesitation on a method on `MultiRegionTimeseriesDataset` is that the dataset doesn't explicitly contain case data. it's more of a wrapper in that sense

Here's an example of the style of method I had in mind: https://github.com/covid-projections/covid-data-model/blob/master/libs/datasets/statistical_areas.py#L25

jabalsad:
A few lines from multiregion.csv. The `new_cases` column is bolded. Notice how it equals the difference in cases between previous row and current row.

01,2020-03-14,iso1:us#iso2:us-al,AL,USA,,state,12,0,**6**,,,6,22,,,,,,,,
01,2020-03-15,iso1:us#iso2:us-al,AL,USA,,state,23,0,**11**,,,12,28,,,,,,,,
01,2020-03-16,iso1:us#iso2:us-al,AL,USA,,state,29,0,**6**,,,28,28,,,,,,,,
01,2020-03-17,iso1:us#iso2:us-al,AL,USA,,state,39,0,**10**,,,36,28,,,,,,,,

libs/datasets/dataset_utils.py
```diff
@@ -142,6 +142,18 @@ def check_index_values_are_unique(data, index=None, duplicates_as_error=True):
...
 
 
+def decorate_new_cases(multiregion_timeseries):
+    """Adds a new_cases column to the dataset by calculating the daily diff in cases.
+
+    Args:
+        multiregion_timeseries: MultiRegionTimeseriesDataset
+
+    """
+    df = multiregion_timeseries.groupby_region()
```
TomGoBravo, line 152: nit: `df` is a some kind of special groupby object, not a normal data frame.

TomGoBravo (changes requested):
Looks nice and simple for wiring this up.

Please add a test. Here is a simple example: https://github.com/covid-projections/covid-data-model/blob/master/test/libs/datasets/timeseries_test.py#L242-L283

Then you can merge and move on to the much more difficult job of making this work with real data :-)

libs/datasets/dataset_utils.py
```diff
@@ -142,6 +142,18 @@ def check_index_values_are_unique(data, index=None, duplicates_as_error=True):
     return None
 
 
+def decorate_new_cases(multiregion_timeseries):
```
ghop02, line 145: nit: `decorate` feels a bit misleading to me only because of the presence of python decorators.  My first thought is that this function would be a decorator

jabalsad: Ah, old habit :-) Happy to call it something more obvious!

libs/datasets/dataset_utils.py
jabalsad: Understood.

jabalsad:
Added a basic unit test to verify new cases are calculated correctly.

Moved the function to be an instance method of the `MultiRegionTimeseriesDataset` class, though I'm not sure this was the best choice. I wasn't entirely comfortable putting it in `dataset_utils` because that seems to be a bit of a random bag of functionality all thrown together, and at the very least should only operate on dataframes, and it sounded like you wanted this functionality to be specific to MRTD classes.

If the new location is not appropriate, please feel free to correct and point me in a better direction :-)

TomGoBravo:
> Added a basic unit test to verify new cases are calculated correctly.
Test looks good, thank you.

> Moved the function to be an instance method of the `MultiRegionTimeseriesDataset` class, though I'm not sure this was the best choice. I wasn't entirely comfortable putting it in `dataset_utils` because that seems to be a bit of a random bag of functionality all thrown together, and at the very least should only operate on dataframes, and it sounded like you wanted this functionality to be specific to MRTD classes.

While neat little bags of code are great I'm okay with random bags when they are documented and tested. So far we've mostly kept 'business' logic out of the classes in timeseries.py. A function at the bottom of timeseries.py seems like an easy place to stick this.

For sure I'd like to keep `MultiRegionTimeseriesDataset` instances immutable, sorry for not noticing the mutation when I last looked at your code.

Please add types to your function `def decorate_new_cases(multiregion_timeseries: MultiRegionTimeseriesDataset) -> MultiRegionTimeseriesDataset:` instead of in the docstring. This increases the chances that pycharm, mypy etc helps us find problems.

jabalsad:
Moving the function to `dataset_utils` is problematic when having an interface that depends on `MultiRegionTimeseriesDataset`: there is a circular import dependency between timeseries.py and dataset_utils.py.

E.g.

```
ImportError while importing test module '/Users/jleroux/w/can/covid-data-model/test/dataset_utils_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test/dataset_utils_test.py:9: in <module>
libs/datasets/__init__.py:4: in <module>
    from libs.datasets.sources.jhu_dataset import JHUDataset
libs/datasets/sources/jhu_dataset.py:7: in <module>
libs/datasets/dataset_utils.py:10: in <module>
    from libs.datasets.timeseries import MultiRegionTimeseriesDataset
libs/datasets/timeseries.py:20: in <module>
    from libs.datasets import dataset_base
libs/datasets/dataset_base.py:9: in <module>
    from libs.datasets.dataset_utils import AggregationLevel, DatasetType
E   ImportError: cannot import name 'AggregationLevel' from 'libs.datasets.dataset_utils' (/Users/jleroux/w/can/covid-data-model/libs/datasets/dataset_utils.py)
```

jabalsad:
Sorry, I misunderstood your comment. I see now you implied that it can be a module function that lives outside of the class, but also contained in timeseries.py?

jabalsad:
Okay, so giving this some more thought, I'm not entirely sure having `MultiRegionTimeseriesDataset` in the interface, and requiring the function to make copies, is the way to go about doing this. The reason being that this class seems to be more complex than I thought (it separates out `timeseries_ts`, `latest_ts` and has provenance info). It also has no simple way of making copies, other than doing a deep copy of each of the individual attributes, and then creating a new class. At this point, the helper function that adds new cases has more logic around how to make a new copy of MRTD, than actual functionality for adding new cases.

Perhaps it makes more sense if the interface accepts a `pd.DataFrame` and returns the same instead?

TomGoBravo:
> Sorry, I misunderstood your comment. I see now you implied that it can be a module function that lives outside of the class, but also contained in timeseries.py?

Yes, that's what I had in mind. Hope you didn't get sucked into a nasty import loop vortex.

TomGoBravo:
> Okay, so giving this some more thought, I'm not entirely sure having `MultiRegionTimeseriesDataset` in the interface, and requiring the function to make copies, is the way to go about doing this. The reason being that this class seems to be more complex than I thought (it separates out `timeseries_ts`, `latest_ts` and has provenance info). It also has no simple way of making copies, other than doing a deep copy of each of the individual attributes, and then creating a new class. At this point, the helper function that adds new cases has more logic around how to make a new copy of MRTD, than actual functionality for adding new cases.
> 
> Perhaps it makes more sense if the interface accepts a `pd.DataFrame` and returns the same instead?

But we still need to get that returned DataFrame back into a MultiRegionTimeseriesDataset. I'm imaging making the copy is similar to the merge function:
```
ts_with_new_cases = mrtd.data.copy()
grouped_df = self.groupby_region()
ts_with_new_cases[CommonFields.NEW_CASES] = grouped_df[CommonFields.CASES].diff(1)
return MultiRegionTimeseriesDataset.from_dataframes(ts_with_new_cases, mrtd.latest.reset_index(), mrtd.provenace)
```

TomGoBravo (approved):
Looks good to me. Thank you for bearing with us.

jabalsad:
Do I just resolve the conflicts and then perform a merge?

TomGoBravo:
> Do I just resolve the conflicts and then perform a merge?

Yes. I generally update the PR desc, then use it again for the squash merge desc.

PR #707, opened 2020-09-29
Add api v2 schemas

ghop02:
Adds in V2 Schemas, used to generate types on the website


TomGoBravo (approved):
Somewhat related: https://github.com/covid-projections/covid-data-model/blob/f29fbde666a7627f08ad327be0cdfa77ca541c54/api/README.md could do with a pointer to the new docs.

PR #708, opened 2020-09-30
Get latest data from multiregion.csv

TomGoBravo:
## This PR

* strengthens `test_persist_and_load_dataset` so both latest and timeseries are compared after being loaded

### Please Check
- [ ] Are tests passing?


ghop02 (approved):
This works because we store the latest data in the multi-region csv without a date correct?

PR #709, opened 2020-09-30
drop dates without real values in the timeseries

TomGoBravo:
Tested by running `python ./run.py data filter-drop-na-dates data/multiregion.csv data/multiregion-drop-na-dates.csv` and looking at `vimdiff data/multiregion.csv data/multiregion-drop-na-dates.csv`. File size dropped from 84MB to 74MB; I was hoping for a bigger drop but code is written now.

### DIFF
--- a/cli/data.py
+++ b/cli/data.py
@@ -86,7 +86,7 @@ def update(summary_filename, wide_dates_filename, aggregate_to_msas: bool):
     )
     multiregion_dataset = MultiRegionTimeseriesDataset.from_timeseries_and_latest(
         timeseries_dataset, latest_dataset
-    )
+    ).drop_na_dates()
     if aggregate_to_msas:
         aggregator = statistical_areas.CountyToCBSAAggregator.from_local_public_data()
         multiregion_dataset = multiregion_dataset.merge(aggregator.aggregate(multiregion_dataset))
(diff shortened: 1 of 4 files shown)


ghop02 (commented):
I like this overall! I'm wondering, will this affect the daily new cases average in the metrics? We frontfill the columns with 0s, so when the first case hits, it's averaging over 6 zeros and 1 value.  If this change causes those front padded zeroes to go away, it might change the calculation.  

One way we could fix this is to make sure that we start the series at a given date in the top level metrics code, but open to suggestions.

It definitely feels too fragile now that this change would ultimately affect the metric calculations

TomGoBravo:
> I like this overall! I'm wondering, will this affect the daily new cases average in the metrics? We frontfill the columns with 0s, so when the first case hits, it's averaging over 6 zeros and 1 value. If this change causes those front padded zeroes to go away, it might change the calculation.
> 
> One way we could fix this is to make sure that we start the series at a given date in the top level metrics code, but open to suggestions.
> 
> It definitely feels too fragile now that this change would ultimately affect the metric calculations

Is there are easy way to check how this PR changes the daily new cases output?

ghop02:
> > I like this overall! I'm wondering, will this affect the daily new cases average in the metrics? We frontfill the columns with 0s, so when the first case hits, it's averaging over 6 zeros and 1 value. If this change causes those front padded zeroes to go away, it might change the calculation.
> > One way we could fix this is to make sure that we start the series at a given date in the top level metrics code, but open to suggestions.
> > It definitely feels too fragile now that this change would ultimately affect the metric calculations
> 
> Is there are easy way to check how this PR changes the daily new cases output?

The easiest would be to simply run a build on the branch (with the data updated).  The slightly more involved would just be to run the api for a couple places and look at the differentces

TomGoBravo:
No longer relevant. We changed to wide dates, which is waaaay smaller.

PR #710, opened 2020-09-30
Changing API Typing to be Required fields with nullable values

ghop02:
@mikelehen brought up a good point that it is strange that we have the fields as being possibly not included, since all fields are currently included right now.

I ended up going down a bit of a rabbit hole, but long story short is that pydantic doesn't actually support outputting a field that is typed: `Optional[int]` as being null or int.  This PR implements some manual tweaking to identify when there is a required field that can be either null or a value and modifying the resulting schema.

It will make the typing more clear on the website

### DIFF
--- a/libs/base_model.py
+++ b/libs/base_model.py
@@ -12,6 +12,7 @@ class APIBaseModel(pydantic.BaseModel):
     class Config:
         json_dumps = _nan_safe_json_dumps
 
+        @staticmethod
         def schema_extra(schema, model):
             # Updating json schema output to respect optional typed fields.
             # Without this code, the schema output doesn't signify that a value
--- a/libs/base_model.py
+++ b/libs/base_model.py
@@ -25,17 +25,9 @@ class APIBaseModel(pydantic.BaseModel):
                             {"type": existing_type},
                             {"type": "null"},
                         ]
-                        print(schema["properties"][field_name])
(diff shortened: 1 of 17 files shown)


libs/base_model.py
```diff
@@ -11,3 +11,34 @@ class APIBaseModel(pydantic.BaseModel):
...
+                        existing_field["anyOf"] = [
+                            {"type": existing_type},
+                            {"type": "null"},
+                        ]
+                    elif "$ref" in existing_field:
+                        existing_type = existing_field.pop("$ref")
+                        existing_field["anyOf"] = [
+                            {"$ref": existing_type},
+                            {"type": "null"},
+                        ]
```
TomGoBravo, line 44: is it worth doing anything if neither type or $ref are found, for example:
assert len({"type", "$ref"} & set(existing_field.keys())) == 1 ?

TomGoBravo (approved):

libs/base_model.py
ghop02: I think in this case we don't actually need to - the schema will include other keys and this I think is a subset of cases.  The "gotcha" would be if there are other keys that we would need to modify that indicate the type of the property, but I'm not aware of any (and certainly not any that we use right now), but could be wrong there.

PR #719, opened 2020-10-06
Add integration test and sentry

ghop02:
This adds sentry tracking to the lambda function + cleans up error messages some with proper responses


api/awsauth/awsauth/auth_app.py
```diff
@@ -81,15 +102,31 @@ def _generate_accept_policy(user_record: dict, method_arn):
     }
 
 
+def _generate_deny_policy(method_arn):
+    # The last part of the method arn does not give permission to the underlying resource.
+    # Replacing with a wildcard to give access to paths below.
+    *first, _ = method_arn.split("/")
```
TomGoBravo, line 108: `method_arn.split("/")[:-1]` to stay with mainstream syntax?

TomGoBravo (approved):

api/awsauth/awsauth/auth_app.py
ghop02: sure!

PR #720, opened 2020-10-07
Patch NJ Typical ICU as Quick Fix

BrettBoval:
This PR addresses issue https://trello.com/c/T15w5VLq/418-nj-icu-occupancy-rate-nonsensical.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/icu_headroom_metric.py
+++ b/libs/icu_headroom_metric.py
@@ -20,6 +20,12 @@ DEFAULT_ICU_DECOMP = 0.21
 NJ_CORRECTION = (
     0.176 - 0.57 + DEFAULT_ICU_DECOMP
 )  # https://trello.com/c/T15w5VLq/418-nj-icu-occupancy-rate-nonsensical
+# This is temporarily correcting for incorrect typical occupancy rate by
+# applying the shim in the decom override. It is negative, because we want
+# to say that the expect ICU occupancy is greater than the reported ICU occupancy
+# since CCM currently reports typical occupancy at 0.176 when they should be
+# reporting something closer to the historical 0.57 from the 2018 source data
+# they draw from.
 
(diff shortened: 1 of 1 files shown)


libs/icu_headroom_metric.py
```diff
@@ -17,6 +17,10 @@
 
 DEFAULT_ICU_DECOMP = 0.21
 
+NJ_CORRECTION = (
+    0.176 - 0.57 + DEFAULT_ICU_DECOMP
```
ghop02, line 21: I'm a bit confused by the subtraction here.  Could you update the comment to mention the mechanics of the subtraction? maybe something that explains the effect of this being negative

BrettBoval: Added a comment.

libs/icu_headroom_metric.py
```diff
@@ -20,6 +20,12 @@
 NJ_CORRECTION = (
     0.176 - 0.57 + DEFAULT_ICU_DECOMP
 )  # https://trello.com/c/T15w5VLq/418-nj-icu-occupancy-rate-nonsensical
```
ghop02, line 22: Very small nit, but when black reformats trailing comments on a single line to be a parens wrapped line, I prefer to put the comment above on its own line

ghop02 (approved):

libs/icu_headroom_metric.py
BrettBoval: Changed to have it with the rest of the comment below. Seems fine with me.

At the same time, I'm going to give you a little bit of a hard time for choosing black as our linter ("Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting") but then still having formatting opinions :) .

PR #721, opened 2020-10-07
Move data availability report to MULTI_REGION dataset.

mikelehen:
Note: I couldn't figure out the actual sheet that got written to, so I'm not 100% sure if this did reasonable things or not...  but it didn't crash.  🤷‍♂️


ghop02 (approved):
Thanks for doing this! I think it's [this sheet](https://docs.google.com/spreadsheets/d/1Z2wy8X9Xp1aWfXNYetPrr1FWNKN5Fraw3MVGMsHBFfc/edit#gid=1461571324), and looks like it's updating alright 👍

PR #722, opened 2020-10-07
Send welcome email on Register

ghop02:
This sends a welcome email on registration using AWS SES to deliver the email.

A couple points:
 * All email sending exceptions are handled gracefully and will not interrupt the registration flow
 * Keeping track of people we have sent emails to so that we can bulk backfill emails for people who have already signed up.
 * Email sending is optionally disabled through the `EMAILS_ENABLED` environment variable.


TomGoBravo (approved):
what happens to bounced emails?

ghop02:
> what happens to bounced emails?

There are events that get fired in aws for that.  we can put more error handling /monitoring in there i believe

PR #723, opened 2020-10-07
Update ICU State to County Distribution Weights Nov 2020

BrettBoval:
This PR updates the weights for State to County ICU using the past 30 days of data. To do that, I had to update the script to handle the new location_id key nomenclature.

For context, these re-weightings have been deliberately manual (instead of recalculated at runtime) to provide stability from the case where a single county reports an erroneous new case count (either due to backfill or fat finger). 

A more complex weighting system was considered (i.e. normalized temporal convolutions of previous 4 weeks cases for each date for each county in a state), but the explainability of the metric has been prioritized for now.

### Please Check
- [ ] Are tests passing?


pyseir/icu/utils.py
```diff
@@ -28,10 +28,13 @@ def calculate_case_based_weights() -> dict:
     last_month_cum_cases.name = "summed_cases"
 
     df = last_month_cum_cases.reset_index().dropna()
-    df["state_fips"] = df["fips"].str.slice(0, 2)  # Apply State Labels to Groupby
+    # Example location_id value = 'iso1:us#iso2:us-ak#fips:02013'
+    df["state_location_id"] = df["location_id"].str.split("#").str[1]  # 'iso2:us-ak'
```
TomGoBravo, line 32: What do you want to happen for MSAs, that don't have a single state?

pyseir/icu/utils.py
```diff
@@ -28,10 +28,13 @@ def calculate_case_based_weights() -> dict:
...
+    df["fips_location_id"] = df["location_id"].str.split("#").str[-1].str.split(":").str[-1]
```
TomGoBravo, line 33: What depends on this value? It looks like the column name in `output` changed from `fips`, what about the reader?

TomGoBravo, line 33: Is ` _add_fips_if_missing` at all handy? https://github.com/covid-projections/covid-data-model/blob/master/libs/datasets/timeseries.py#L328-L331

BrettBoval: Updated to use that helper function instead of rolling my own parser.

BrettBoval: Both "state_location_id" and "fips_location_id" are internal to this function. I switched to CommoFields.FIPS, but have to give more thought about generalizing this to move beyond the state-county mapping.

pyseir/icu/utils.py
BrettBoval: My first inclination is to sum up all the estimated ICU numbers from the county level. That breaks the hierarchy a little though. So you would disaggregate from states to counties and then reaggregate back to MSAs. There is an edge case where one state has actuals while the other has estimates that will result in a blended estimate, but I think that's the best we can do. Open to suggestions. Do you have any ideas?

ghop02 (approved):

PR #724, opened 2020-10-08
test positivity multiregion

TomGoBravo:
## This PR

This PR is like https://github.com/covid-projections/covid-data-public/pull/136 but in this repo, after combined datasets.

This is part of https://trello.com/c/6ZPv7k9T/477-revamp-test-positivity-calculation

* Adds module `test_positivity` that calculates and returns in memory all test positivity methods for all regions
* Adds `MultiRegionTimeseriesDataset.timeseries_long` and `get_regions_subset`
* Adds command `./run.py api generate-test-positivity --test-positivity-all-methods test-pos.csv` to dump all methods


cli/api.py
```diff
@@ -174,9 +182,15 @@ def generate_api_v2(model_output_dir, output, aggregation_level, state, fips):
         states=active_states,
     )
     _logger.info(f"Loading all regional inputs.")
+
+    regions_data = combined_datasets.load_us_timeseries_dataset().get_regions_subset(regions)
+    test_positivity_results = test_positivity.AllMethods.run(regions_data)
+    test_positivity_results.write(test_positivity_all_methods)
```
ghop02, line 188: If we're writing the test positivity output here, it won't get included as an artifact.  Maybe this file would be better as an input and be generated during the pyseir step?

libs/test_positivity.py
```diff
@@ -0,0 +1,142 @@
...
+    def calculate(self, delta_df: pd.DataFrame) -> pd.DataFrame:
+        assert delta_df.columns.names == [CommonFields.DATE]
+        assert delta_df.index.names == [PdFields.VARIABLE, CommonFields.LOCATION_ID]
+        # delta_df has the field name as the first level of the index. delta_df.loc[field, :] returns a
+        # DataFrame without the field label so operators such as `/` are calculated for each
+        # region/state and date.
+        return delta_df.loc[self.numerator, :] / delta_df.loc[self.denominator, :]
+
+
+TEST_POSITIVITY_METHODS = (
```
ghop02, line 31: These kinda feel like they could an enum

ghop02 (commented):
This looks good overall! I'd really prefer to save the generated csv somewhere else (just since it won't get included in any automated model builds). 

Definitely fine getting this in and then changing up how top level metrics generates, Would just be good to get a snapshot building so we can compare against existing sooner rather than later

libs/test_positivity.py
TomGoBravo: Sort of, but I think it'd make testing using a different set of methods harder.

cli/api.py
TomGoBravo: Generating them is so fast it is hardly worth saving :-P

TomGoBravo: It takes about 19 seconds to run all methods for all states+counties on my dinky laptop. Not great but not horrible.

PR #729, opened 2020-10-13
Updating to pydantic 1.6.1 and fixing enum handling

ghop02:
Updating pydantic to 1.6.1.  

The enum handling changed a bit in the latest version.  Also, it looks like the generated schemas in master for v2 were built with 1.6.1 so nothing changed for the new api.

This is definitely a hacky fix - not great code but we should be removing it soon when we yank out API v1.

Regenerated the schemas with `./run.py api update-schemas`


TomGoBravo (approved):
Please add command line used to regenerate docs in the PR desc.

PR #730, opened 2020-10-13
Adding api users updater

ghop02:
This adds in a cli command to update a google sheet with the latest API User activity.  

execution:
```
./run.py utils update-api-user-usage
```


libs/google_sheet_helpers.py
```diff
@@ -64,6 +64,42 @@ def create_or_replace_worksheet(
     return sheet.add_worksheet(worksheet_name, 100, 100)
 
 
+def create_or_clear_worksheet(sheet: gspread.Spreadsheet, worksheet_name: str) -> gspread.Worksheet:
```
TomGoBravo, line 67: s/sheet/spreadsheet/ since `sheet` is pretty overloaded

TomGoBravo (approved):
Oh boy, that looks like it was annoying to write.

PR #732, opened 2020-10-13
Add HHS data to combined dataset as highest priority.

mikelehen:


libs/datasets/sources/hhs_testing_dataset.py
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    COMMON_FIELD_MAP = {f: f for f in {CommonFields.NEGATIVE_TESTS, CommonFields.POSITIVE_TESTS,}}
+
+    @classmethod
+    def local(cls):
+        data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
+        input_path = data_root / cls.DATA_PATH
+        data = common_df.read_csv(input_path).reset_index()
+        return cls(cls._rename_to_common_fields(data))
```
TomGoBravo, line 27: renaming happened in the other repo. I expect `cls(data)` to work.

TomGoBravo (approved):
one small change and good to merge

libs/datasets/sources/hhs_testing_dataset.py
mikelehen: Whoops!  Too much copy-pasta.  Good catch, thanks!

PR #733, opened 2020-10-14
Use lambda edge

ghop02:
This reconfigures API Registration and Authentication to use cloudfront lambda@edge functions.  

A few notes:
 * Cloudfront doesn't have as flexible cloudwatch log configurations.  As a result, I couldn't find a super straightforward way to pass the email through to cloudfront logs.  To keep up our API User querying ability, I went the route of sending successful authentication requests to a firehose, which drops those records into S3.  The Viewer-Request should not be cached, so each request should log a record.  From there, we can create a small Athena table that will let us query in a way that lets us return similar aggregations
 * Lambda@edge functions don't support environment variables.  To get around this, I am using an .env file.  When we do automated deploys, we will be able to output a .env in the deploy process with the appropriate variables set. 
 * Somewhat related, but I added a global `Config` class that lets us initialize constants.  I've actually been meaning to do something like this for our model/api code, but I think it's a clean way of loading settings from environment variables/an environment file.


api/awsauth/awsauth/config.py
```diff
@@ -0,0 +1,28 @@
...
+    EMAILS_ENABLED: bool
+
+    class Config:
+        env_file = ".env"
+        env_file_encoding = "utf-8"
+
+
+class Config:
+
+    Constants: EnvConstants = None
```
TomGoBravo, line 24: Is there a case where we'd want to create a Config instance? If not, how about putting `constants: EnvConstants = None` at the module level?

TomGoBravo (approved):

api/awsauth/awsauth/auth_app.py
```diff
@@ -143,3 +174,65 @@ def check_api_key(event, context):
...
 
     return _generate_accept_policy(record, method_arn)
+
+
+def check_api_key_edge(event, context):
+    request = event["Records"][0]["cf"]["request"]
+
+    query_parameters = urllib.parse.parse_qs(request["querystring"])
+    # parse query parameter by taking first api key in query string arg.
+    api_key = None
```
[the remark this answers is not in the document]
ghop02: ooo that does look nice - the lambda edge deployments are pretty finnicky with deployment size requirements (I think the deployment size limit is 1MB) so i'll probably leave it out for now, but will consider in the future!

api/awsauth/.gitignore
```diff
@@ -16,6 +16,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+.env.local
```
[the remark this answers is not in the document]
ghop02: ah this was left over

api/awsauth/awsauth/config.py
ghop02: definitely fine with that!

ghop02: Ah so the reason that I actually don't do that is so that we can initialize the environment at runtime rather than import. Because otherwise, I believe that references to constants will get set at import time? though I'm now questioning if that's actually right

TomGoBravo: I'm imagining the module level variable is None until the init function is called, so the env is not read during import.

ghop02: Right - you would just have to be careful to not import `constants` directly (i.e. `from awsauth.config import constants`) or constants will be none.  However:
```
from awsauth import config
config.constants
```
would work alright

PR #736, opened 2020-10-14
Un revert infert rt nola patch

TomGoBravo:
## This PR

This unreverts https://github.com/covid-projections/covid-data-model/pull/735, in other words it is a second attempt at https://github.com/covid-projections/covid-data-model/pull/731 this time with fix for `KeyError <CommonFields.ICU_BEDS: 'icu_beds'>` reported at https://sentry.io/organizations/covidactnow/issues/1957780407/?project=5203044&referrer=slack

Changes `OneRegionTimeseriesDataset.data` to always contain location_id.

Changes `MultiRegionTimeseriesDataset.data` to *not* have FIPS, forcing callers that want it to use `data_with_fips`.

### Please Check
- [ ] Are tests passing?


libs/datasets/timeseries.py
```diff
@@ -113,14 +107,20 @@ def yield_records(self) -> Iterable[dict]:
 
     def get_subset(self, after=None, columns=tuple()):
         rows_key = dataset_utils.make_rows_key(self.data, after=after,)
-        columns_key = list(columns) if columns else slice(None, None, None)
+        if columns:
+            assert CommonFields.LOCATION_ID in columns
```
ghop02, line 111: I think if we have these asserts in the input data itself it makes it easier to reason about it elsewhere.  Why did you move it from the post init to here?

ghop02 (approved):
I think this is a step in the right direction and largely agrees with some of the work that I am doing to get MSAs into the API.

libs/datasets/timeseries.py
TomGoBravo: I bumped into a bug where LOCATION_ID wasn't in columns but you are correct that the check makes more sense in `__post_init__` so moved them there.

PR #737, opened 2020-10-15
test positivity provenance in API

TomGoBravo:
This PR follows up on https://github.com/covid-projections/covid-data-model/pull/728 to add provenance information.

* Populates provenance information for test positivity in the API output

## Tested

Ran a snapshot and looked at output such as https://data.covidactnow.org/snapshot/1283/v2/county/53033.timeseries.json where I see
```
metrics: {testPositivityRatio: null, testPositivityRatioDetails: {source: "other"},…}
```


libs/datasets/timeseries.py
```diff
@@ -105,12 +107,14 @@ def get_subset(self, after=None, columns=tuple()):
         rows_key = dataset_utils.make_rows_key(self.data, after=after,)
         columns_key = list(columns) if columns else slice(None, None, None)
         return OneRegionTimeseriesDataset(
-            self.data.loc[rows_key, columns_key].reset_index(drop=True), latest=self.latest,
+            self.data.loc[rows_key, columns_key].reset_index(drop=True),
+            latest=self.latest,
+            provenance=self.provenance,
```
mikelehen, line 112: `latest` and `provenance` are going to have columns outside of the `columns` subset specified, right?  But maybe we don't care?  Perhaps at least add a comment.

libs/datasets/timeseries.py
```diff
@@ -478,9 +482,17 @@ def from_combined_dataframe(
         return multiregion_timeseries
 
     @staticmethod
-    def from_csv(path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionTimeseriesDataset":
+    def from_csv(
+        path_or_buf: Union[pathlib.Path, TextIO], provenance: Optional[pd.Series] = None,
```
mikelehen, line 486: I was curious in what case we would be loading a CSV but have already-parsed provenance to go along with it...  It turns out we don't really.  But in tests, we load the provenance from a (string) CSV and then pass the _parsed_ version (as `pd.Series`) into from_csv() along with the _unparsed_ primary CSV (as just a `TextIO`).

This seems like a bit arbitrary / inconsistent.  Why not have `provenance` be `Optional[TextIO]` and always do the CSV parsing inside this method?  Or for full symmetry between the primary CSV and the provenance CSV, you could do `Optional[Union[pathlib.Path, TextIO]]` but that's overly flexible since we never need to pass in a Path right now.  🤷

api/can_api_definition.py
```diff
@@ -75,6 +76,15 @@ class CovidPatientsMethod(enum.Enum):
     ESTIMATED = "estimated"
 
 
+class TestPositivityRatioMethod(GetByValueMixin, enum.Enum):
+    """Method used to determine test positivity ratio."""
+
+    # XXX Fill out with actual method names
+    HHSTesting = "HHSTesting"
+    METHOD2 = "method2"
```
mikelehen, line 84: Please make sure we don't ship this in the API as-is. 😁

mikelehen (commented):
Left a few comments, but this seems reasonable to me.  I can review in more detail once this is ready to go (and merged with latest master so it doesn't include the diffs from the test positivity PR).

A couple high-level thoughts:
1. Provence is essentially just a per-location/column piece of metadata, right?  I wonder if we should be thinking about creating that abstraction instead (column metadata) instead of baking this specific use of column metadata (provenance) into our data structures and naming.
2. On that note, test_positivity already has kinda' 2 forms of provenance, right? a) what data source did the data come from ("X dashboard via covid-tracking"), and b) which calculation did we use to calculate the final metric value ("positive_tests_viral / total_tests_viral").  Do you have a sense of how we would represent that?  If we have a single provenance field per location/column, do we have to come up with some sort of string-encoding of all of that (a JSON blob or something)?

Also, just FYI- When I run this branch locally right now I get tons of "Unable to find TestPositivityRatioMethod location_id=iso1:us#iso2:us-sd#fips:46003" errors which I don't get in master.  But maybe that'll be fixed by a merge.

libs/datasets/timeseries.py
TomGoBravo: Yes, that sucked. Fixed with https://github.com/covid-projections/covid-data-model/pull/757

libs/datasets/timeseries.py
TomGoBravo: Maybe, but probably not soon.

TomGoBravo:
> Left a few comments, but this seems reasonable to me. I can review in more detail once this is ready to go (and merged with latest master so it doesn't include the diffs from the test positivity PR).
> 
> A couple high-level thoughts:
> 
> 1. Provence is essentially just a per-location/column piece of metadata, right?  I wonder if we should be thinking about creating that abstraction instead (column metadata) instead of baking this specific use of column metadata (provenance) into our data structures and naming.

Yes it is. While it is good to keep future direction in mind i don't want to build internal interfaces that are more general than what we need right now and/or take us in step along the path we know we want to follow.

> 2. On that note, test_positivity already has kinda' 2 forms of provenance, right? a) what data source did the data come from ("X dashboard via covid-tracking"), and b) which calculation did we use to calculate the final metric value ("positive_tests_viral / total_tests_viral").  Do you have a sense of how we would represent that?  If we have a single provenance field per location/column, do we have to come up with some sort of string-encoding of all of that (a JSON blob or something)?

Totally agree that provenance can easily have N sources and 1 method to combine them. JSON blob isn't a bad idea because it is easy to imagine that `source` could be broken down to dataset name and version (for example sha in the github world). That's beyond the scope of this PR.

> Also, just FYI- When I run this branch locally right now I get tons of "Unable to find TestPositivityRatioMethod location_id=iso1:us#iso2:us-sd#fips:46003" errors which I don't get in master. But maybe that'll be fixed by a merge.
Opps, I think this is fixed now.

api/can_api_v2_definition.py
```diff
@@ -61,6 +61,9 @@ class Metrics(base_model.APIBaseModel):
         ...,
         description="Ratio of people who test positive calculated using a 7-day rolling average.",
     )
+    testPositivityRatioMethod: Optional[
```
ghop02, line 64: My first thought is that this would be better served as `testPositivityRatioDetails` with the `method` being the only item inside.  

I think it'll make it more future proof.  If we ever add another metadata option (don't see one immediately, but wouldn't shock me) it will make it easier to make a non breaking change.  In that case, I would prefer to have it in its own object rather than cluttering the top level metrics space.  Additionally it copies an existing pattern of `icuHeadroomDetails`.

api/can_api_definition.py
```diff
@@ -75,6 +76,14 @@ class CovidPatientsMethod(enum.Enum):
     ESTIMATED = "estimated"
 
 
+class TestPositivityRatioMethod(GetByValueMixin, enum.Enum):
+    """Method used to determine test positivity ratio."""
+
+    HHSTesting = "HHSTesting"
+    VALORUM = "Valorum"
```
ghop02, line 83: Valorum feels pretty unhelpful for anyone who doesn't know the internals of our API data pipeline. I guess there's not really much of a better option right now

libs/top_level_metrics.py
```diff
@@ -100,14 +102,35 @@ def calculate_metrics_for_timeseries(
...
 
     return metrics, metric_summary
 
 
-def calculate_or_copy_test_positivity(data):
+def _lookup_test_positivity_method(
+    positive_tests_provenance: Optional[str], negative_tests_provenance: Optional[str], log
+) -> TestPositivityRatioMethod:
+    method = None
+    if positive_tests_provenance and positive_tests_provenance == negative_tests_provenance:
```
ghop02, line 116: Can you add a comment about when this wouldn't ever be true? after reading into it a bit more I can guess, but some context would be helpful

libs/top_level_metrics.py
```diff
@@ -121,9 +144,22 @@ def calculate_or_copy_test_positivity(data):
...
+        return (
+            calculate_test_positivity(cumulative_positive_tests, cumulative_negative_tests),
+            method,
+        )
     else:
-        return test_positivity
+        provenance = ts.provenance.get(CommonFields.TEST_POSITIVITY)
+        method = TestPositivityRatioMethod.get(provenance)
+        if method is None:
+            log.info("Unable to find TestPositivityRatioMethod", provenance=provenance)
```
ghop02, line 160: Info feels like a dangerous move here.  Do we ever expect this to happen?

libs/pipelines/api_v2_pipeline.py
```diff
@@ -120,6 +120,7 @@ def generate_metrics_and_latest(
     timeseries: OneRegionTimeseriesDataset,
     rt_data: Optional[OneRegionTimeseriesDataset],
     icu_data: Optional[OneRegionTimeseriesDataset],
+    log,
```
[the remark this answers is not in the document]
TomGoBravo: That's funny because I saw the implicit passing around of a bound logger as an anti-pattern. What we have right now is not very consistent :-( which is why I've been erring towards keeping everything explicit on the stock. Cleaning up logging to be more consistent seems like a fine idea but for now I'd like to avoid introducing yet another way to do logging in this repo.

libs/top_level_metrics.py
TomGoBravo: Do you mean you'd like this to be warning so we are less likely to overlook it breaking? SGTM.

libs/pipelines/api_v2_pipeline.py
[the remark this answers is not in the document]
ghop02: Heh, that's why I'm still not sold on structlog :) never really had a problem with the std logger before haha

api/can_api_v2_definition.py
```diff
@@ -129,6 +132,9 @@ class RiskLevels(base_model.APIBaseModel):
     testPositivityRatio: RiskLevel = pydantic.Field(
         ..., description="Test positivity ratio risk level."
     )
+    testPositivityRatioMethod: Optional[
+        can_api_definition.TestPositivityRatioMethod
+    ] = pydantic.Field(None)
```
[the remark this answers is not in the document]
mikelehen: FWIW- I'd advocate we don't add it to the API at all until we have specced out the desired API a bit better.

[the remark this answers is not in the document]
mikelehen: (e.g. as @ghop02 pointed out, "Valorum" isn't what we want to surface in the API)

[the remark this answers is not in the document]
ghop02: Do we have plans to use it in the short term? I may have been the person to push for this in the API (honestly not 100% sure), but would be good to figure out how to communicate this difference somewhere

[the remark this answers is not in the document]
mikelehen: Actually...  yes, maybe.  I would like to get CMS data in as a backfill this week, and the CMS data is a 14-day average that only has weekly datapoints, and may be up to 7-10 days lagged...  And we probably need to surface that on the site somehow.

I could do that in a hacky way (e.g. hardcode where we're using CMS data or "detect" it based on the cadence of data points)...  or we could do it propely and plumb it through with this provenance work.

The part that makes me nervous is:
* Exposing "Valorum" as a name to API consumers.
* "Valorum" vs "HHSTesting" isn't sufficient, since we also use Covid Tracking Project for ~3 states right now.
* As Chris mentioned, we may want to nest this under a "testPositivityRatioDetails" object so that we give ourselves flexibility to expand.

So it'd be nice if we could expose this as a "private" API until we are happy with it so we don't have to worry about breaking people..  though, in practice probably nobody will notice / use this data, so it may be okay for us to make breaking changes.

api/can_api_v2_definition.py
mikelehen: testPositivityRatioDetails =
```
{
  source: 'CMS'
}
```

libs/top_level_metrics.py
TomGoBravo: I probably made this function work when provenance is None before I had loading the provenance done correctly, then left it in place because it seemed better to log a warning than raise an exception. Would you prefer an exception?

libs/top_level_metrics.py
```diff
@@ -100,14 +102,35 @@ def calculate_metrics_for_timeseries(
...
-def calculate_or_copy_test_positivity(data):
+def _lookup_test_positivity_method(
+    positive_tests_provenance: Optional[str], negative_tests_provenance: Optional[str], log
+) -> TestPositivityRatioMethod:
+    method = None
+    if positive_tests_provenance and positive_tests_provenance == negative_tests_provenance:
+        method = TestPositivityRatioMethod.get(positive_tests_provenance)
+    if method is None:
+        log.info(
+            "Unable to find TestPositivityRatioMethod",
```
ghop02, line 120: After looking at the logs, looks like this is pretty noisy.  We should probably remove this log or maybe find a way to log total number of "other" results

libs/top_level_metrics.py
```diff
@@ -224,7 +260,10 @@ def calculate_latest_metrics(
     Returns: Metrics
     """
     data = data.set_index(CommonFields.DATE)
-    metrics = {}
+    metrics = {
+        "testPositivityRatioDetails": test_positivity_method,
+        "icuHeadroomDetails": icu_metric_details,
```
ghop02, line 265: 🙏

ghop02 (approved):
would like to clean the logging up but looks good!

libs/top_level_metrics.py
TomGoBravo: Cleaning this up will be much easier when test positivity calculation is extracted from top_level_metrics. Changed to debug for now.

PR #738, opened 2020-10-15
Small api fixes

ghop02:
This addresses a couple of issues that we were seeing in the API.

Also I added a registry that contains initialized connections.  This should drop the latency by only making a connection once per container rather than on every request.

### DIFF
--- a/api/awsauth/awsauth/the_registry.py
+++ b/api/awsauth/awsauth/the_registry.py
@@ -3,7 +3,6 @@ from awsauth.ses_client import SESClient
 from awsauth.firehose_client import FirehoseClient
 from awsauth.dynamo_client import DynamoDBClient
 from awsauth.ses_client import SESClient
-from awsauth.config import Config
 
 
 class RegistryProvider:
--- a/api/awsauth/awsauth/the_registry.py
+++ b/api/awsauth/awsauth/the_registry.py
@@ -1,0 +1,51 @@
+import inspect
+from awsauth.ses_client import SESClient
+from awsauth.firehose_client import FirehoseClient
+from awsauth.dynamo_client import DynamoDBClient
+from awsauth.ses_client import SESClient
+from awsauth.config import Config
+
+
+class RegistryProvider:
+    @property
+    def ses_client(self) -> SESClient:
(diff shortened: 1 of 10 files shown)


api/awsauth/awsauth/welcome_email.html
```diff
@@ -19,10 +19,9 @@
       reminder, a commercial license is required for any user who engages in
       commercial activities including selling goods or services. If this
       describes you, or if you’re interested in paid support, please
-      <a href="https://calendly.com/c/CBAXPVVOF7W6SWFN?month=2020-09"
-        >sign up for a 20 minute slot</a
-      >
-      to chat with our partnerships team.
+      <a href="mailto:tommacmullen@covidactnow.org">email our Partnerships Team</a>
```
TomGoBravo, line 22: Email address and link text don't match up. How about an email alias or mention Tom in the text?

api/awsauth/awsauth/the_registry.py
```diff
@@ -0,0 +1,50 @@
...
+        return DynamoDBClient()
+
+
+class Registry(object):
+    def __init__(self):
+        self._initialized_objects = {}
+
+    def initialize(self):
+        properties = [
+            prop for prop, obj in inspect.getmembers(self.__class__) if isinstance(obj, property)
```
TomGoBravo, line 28: This seems like a lot of logic where a cached_property on a method that creates the desired object would work. It isn't in 3.7 by default but there is https://pypi.org/project/backports.cached-property/

TomGoBravo (approved):

api/awsauth/awsauth/the_registry.py
ghop02: Yeah I think that's a good point.  Maybe if we switch to 3.8 or 3.9 we can just use cached property :).  I think I'll keep this way for now.

api/awsauth/awsauth/welcome_email.html
ghop02: 👍 I'm going to get some updated copy and will update soon - probably tomorrow

PR #739, opened 2020-10-15
Add MultiRegion iterator

TomGoBravo:
This is the subset of https://github.com/covid-projections/covid-data-model/pull/728 that is not expected to change the output.

* Rename `MultiRegionTimeseriesDataset.merge` to `append_regions`
* Add `MultiRegionTimeseriesDataset.join_columns` and `iter_one_regions`
* Change `api_v2_pipeline.RegionalInput` to get combined data as OneRegionTimeseriesDataset parameters instead of loading it from disk to a RegionalData
* Change how generate-api-v2 creates `api_v2_pipeline.RegionalInput` to use  `iter_one_regions` which makes a single groupby call for each MultiRegion instead of a query for each region. With this making the inputs for all regions takes ~10 seconds.

### DIFF
--- a/libs/pipelines/api_v2_pipeline.py
+++ b/libs/pipelines/api_v2_pipeline.py
@@ -23,7 +23,6 @@ from libs.datasets.timeseries import OneRegionTimeseriesDataset
 from libs.datasets.timeseries import MultiRegionTimeseriesDataset
 from libs.enums import Intervention
 from libs.functions import build_api_v2
-from libs.datasets import combined_datasets
 from libs.datasets import AggregationLevel
 
 logger = structlog.getLogger()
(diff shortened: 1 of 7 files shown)


libs/pipelines/api_v2_pipeline.py
```diff
@@ -52,15 +51,16 @@ def latest(self) -> Dict[str, Any]:
...
 
     @staticmethod
     def from_region_and_model_output(
         region: pipeline.Region,
+        combined_data: MultiRegionTimeseriesDataset,
         rt_data: MultiRegionTimeseriesDataset,
         icu_data: MultiRegionTimeseriesDataset,
     ) -> "RegionalInput":
-        combined_data = combined_datasets.RegionalData.from_region(region)
+        one_region_data = combined_data.get_one_region(region)
```
ghop02, line 63: hehe this is a todo for the future, and controversial syntax, but could be cool if the syntax was `combined_data[region]`

ghop02 (approved):

libs/pipelines/api_v2_pipeline.py
TomGoBravo: -1 for making slow operations hide behind simple syntax. I would be into making `MultiRegionTimeseriesDataset` objects iterable to yield fast access to all the regions.

PR #745, opened 2020-10-19
Filter to remove regions without population data

TomGoBravo:


ghop02 (approved):
LGTM.  Planning on running this? Also, a couple of those are because the covid care map data is using a legacy fips.  We should probably either change those entries or let covid care map know that the fips is out of date

TomGoBravo:
> LGTM. Planning on running this? Also, a couple of those are because the covid care map data is using a legacy fips. We should probably either change those entries or let covid care map know that the fips is out of date

Looking at the output of `./run.py data run-population-filter populated-filtered.csv` I see most of the regions without population are fake counties ending 999. The rest are territories other than PR `'iso1:us#iso2:us-vi', 'iso1:us#iso2:us-as', 'iso1:us#iso2:us-gu'`, subregions of Puerto Rico `'iso1:us#iso2:us-vi#fips:78030',  , 'iso1:us#iso2:us-vi#fips:78020', 'iso1:us#iso2:us-vi#fips:78010'` and 

`iso1:us#iso2:us-sd#fips:46113`
"South Dakota, 2015: Shannon County (FIPS 46113) is renamed to Oglala
Lakota County (FIPS 46102) as of May 1st, 2015." 

`'iso1:us#iso2:us-va#fips:51515'`
"Virginia, 2013: Bedford (independent) city (FIPS 51515) was changed to
town status and added to Bedford County (FIPS 51019) effective July 1st,
2013."

(text from https://www.ddorn.net/data/FIPS_County_Code_Changes.pdf)

TomGoBravo:
I created https://github.com/covidcaremap/covid19-healthsystemcapacity/issues/193

TomGoBravo:
I have a hunch that removing the fake counties ending 999 may break something downstream. What do you think @ghop02 ?

PR #747, opened 2020-10-21
Update new case logic to not throw away first case

ghop02:
This updates the new cases logic to not throw away the first case.  Small logic change before we can use it in a couple of other places.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -671,12 +671,29 @@ def _remove_padded_nans(df, columns):
     return df.reset_index(drop=True)
 
 
-def add_new_cases(mrts: MultiRegionTimeseriesDataset) -> MultiRegionTimeseriesDataset:
+def _diff_preserving_first_value(series):
+
+    series_diff = series.diff()
+    first_valid_index = series.first_valid_index()
+    if first_valid_index is None:
+        return series_diff
+
+    series_diff[first_valid_index] = series[series.first_valid_index()]
+    return series_diff
+
+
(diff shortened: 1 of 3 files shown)


libs/datasets/timeseries.py
```diff
@@ -671,12 +671,29 @@ def _remove_padded_nans(df, columns):
...
-        timeseries_df=df_copy, provenance=mrts.provenance
-    ).append_latest_df(mrts.latest_data.reset_index())
-    return new_mrts
+    df_copy = timeseries.data.copy()
+    grouped_df = timeseries.groupby_region()
+    # Calculating new cases using diff will remove the first detected value from the case series.
+    # We want to capture the first day a region reports a case. Since our data sources have
+    # been capturing cases in all states from the beginning of the pandemic, we are treating
+    # The first days as appropriate new case data.
+    df_copy[CommonFields.NEW_CASES] = grouped_df[CommonFields.CASES].apply(
```
TomGoBravo, line 693: Can we avoid the apply by adding `df_copy[CommonFields.NEW_CASES] = grouped_df[CommonFields.CASES].nth(0, dropna='any')` (or maybe use `first`) after the existing `df_copy[CommonFields.NEW_CASES] = grouped_df[CommonFields.CASES].diff(1)`  ?

TomGoBravo (approved):

libs/datasets/timeseries.py
ghop02: hmmm i gave that a shot and i didn't find anything that slotted it in as nicely.
 this was the easiset way I found to preserve, but agree i'd prefer two easier statements. Going to merge as is but happy to swap in another solution

PR #751, opened 2020-10-22
Adding ability to trigger api build from github action

ghop02:
Adds an option in the dataset updater to optionally trigger api build on dataset update.


mikelehen (approved):
LGTM.  Is there going to be a corresponding change to the .sh script we use to trigger this workflow?

FWIW- I think we should move towards workflow_dispatch instead of repository_dispatch and retiring our .sh scripts.  If you feel motivated to poke at that, I'd be psyched. 😁  https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

TomGoBravo (approved):

ghop02:
Ooo the workflow event is cool! I think a good one to start with would be actually be deploy_api.

PR #754, opened 2020-10-23
Add `locationId` to API V2 output

jabalsad:
This PR addresses issue https://trello.com/c/f25sZkDe/526-all-location-id-to-api-output

### Please Check
- [Y] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/master/api/schemas) updated?
- [N] Are the [api docs](https://github.com/covid-projections/covid-data-model/blob/master/api/README.V1.md) updated?
- [Y] Are tests passing?

### DIFF
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -158,6 +158,10 @@ class RegionSummary(base_model.APIBaseModel):
     lat: Optional[float] = pydantic.Field(
         ..., description="Latitude of point within the state or county"
     )
+    locationId: str = pydantic.Field(
+        ...,
+        description="Location ID as defined here: https://github.com/covidatlas/li/blob/master/docs/reports-v1.md#general-notes",
+    )
     long: Optional[float] = pydantic.Field(
         ..., description="Longitude of point within the state or county"
     )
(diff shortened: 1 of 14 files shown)


api/docs/open_api_schema.json
```diff
@@ -654,23 +654,6 @@
         },
         "description": "Known actuals data."
       },
-      "CovidPatientsMethod": {
```
jabalsad, line 657: I'm not sure why these entered the diff. I have the latest master checked out, but running schema generation includes these changes. Please advise?

api/can_api_v2_definition.py
```diff
@@ -158,6 +158,10 @@ class RegionSummary(base_model.APIBaseModel):
     lat: Optional[float] = pydantic.Field(
         ..., description="Latitude of point within the state or county"
     )
+    locationId: str = pydantic.Field(
+        ...,
+        description="Location ID as defined here: https://github.com/covidatlas/li/blob/master/docs/reports-v1.md#general-notes",
```
jabalsad, line 163: Took this description from the docs here: https://github.com/covid-projections/covid-data-public/blob/master/covidactnow/datapublic/common_fields.py#L39-L41

jabalsad (commented):
It's worth pointing out that adding `location_id` to the CSV will break consumers that rely on positional columns. Not sure if this will be an actual issue.

api/docs/open_api_schema.json
ghop02: I think you may need to update your requirements.  We bumped the version of pydantic which changed the schema generation a bit.

libs/functions/build_api_v2.py
```diff
@@ -73,6 +78,7 @@ def build_region_timeseries(
     for row in timeseries.yield_records():
         # Timeseries records don't have population
         row[CommonFields.POPULATION] = region_summary.population
+        row[CommonFields.LOCATION_ID] = region_summary.locationId
```
ghop02, line 81: I don't think we need to add this to the timeseries row.  Actually looking at it, I don't think we need population either (and don't believe it actually gets included in the actuals object).  I think you could remove population as well, but should at least remove location id here

ghop02 (approved):
One comment, but otherwise lgtm!  It is a good point about breaking the csv, but I don't think it's worth worrying about right now.  @TomGoBravo or @mikelehen do you agree or should we think more carefully about it?

mikelehen:
👍 Definitely don't think we should worry about changing the numbering of the columns in the CSV.

PR #755, opened 2020-10-24
Fix tests on macOS by disabling use of multiprocessing.

mikelehen:
Created a parallel_utils helper module that has parallel_map() and
pandas_parallel_apply() functions which use multiprocessing / pandarallel on
non-macOS but plain non-parallel versions on macOS.

This works around bugs in Python and pandarallel:
https://bugs.python.org/issue33725#msg343838
https://github.com/nalepae/pandarallel/issues/72


libs/parallel_utils.py
```diff
@@ -0,0 +1,41 @@
...
+pandarallel.initialize(progress_bar=VISIBIBLE_PROGRESS_BAR)
+
+# multiprocessing is unreliable on macOS. See https://bugs.python.org/issue33725#msg343838
+# In theory, using "spawn" start_method would work, but that triggers a bug in pandarallel
+# (https://github.com/nalepae/pandarallel/issues/72).
+USE_MULTIPROCESSING = platform.system() != "Darwin"
+
+T = TypeVar("T")
+R = TypeVar("R")
+SERIES_OR_DF = TypeVar("SERIES_OR_DF", pd.Series, pd.DataFrame)
```
mikelehen, line 20: Let me know if I'm doing this type stuff wrong or going overboard / whatever. 🤷

libs/parallel_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def parallel_map(func: Callable[[T], R], iterable: Iterable[T]) -> List[R]:
+    """Runs func on each item in iterable, in parallel if possible."""
+    if USE_MULTIPROCESSING:
```
ghop02, line 25: It might be nice to allow this to be overridden by an environment variable.  I know that I'm generally able to run something like pyseir in parallel on my mac (though not 100% sure how given the issues) and I would be disappointed if i had to change this variable manually to do so.  

May be we should log a warning that it's not being ran in parallel?

pyseir/inference/whitelist.py
```diff
@@ -45,14 +41,11 @@ def generate_whitelist(self, timeseries: MultiRegionTimeseriesDataset) -> pd.Dat
...
-            timeseries.get_counties()
-            .groupby_region()
-            # Use pandarallel. It doesn't support the `name` attribute so leave FIPS as a regular
-            # column so it can be read in the applied function.
-            .parallel_apply(_whitelist_candidates_per_fips)
-            .reset_index(drop=True)
-        )
+        # Use pandarallel. It doesn't support the `name` attribute so leave FIPS as a regular
+        # column so it can be read in the applied function.
+        df_candidates = pandas_parallel_apply(
```
ghop02, line 46: Argh we're going to be dropping the whitelist soon anyways (which will then drop our usage of pandarallel).  Can try to prioritize that so we don't have to work around pandarallel

ghop02 (approved):
Like the general structure (and typing!).  Few suggestions that I think would be good to deal with but nothing blocking

mikelehen (commented):
Thanks! Addressed everything, I think.

PR #757, opened 2020-10-26
provenance read and write

TomGoBravo:
This PR is the timeseries part of https://github.com/covid-projections/covid-data-model/pull/737

* Adds provenance to OneRegionTimeseriesDataset
* Changes `MultiRegionTimeseriesDataset.from_csv` to read provenance data if `to_csv` wrote it.

## Tested

Tests pass.

Only change introduced by `python ./run.py data update --summary-filename=''` is renaming `value` column in `data/multiregion-provenance.csv` to `provenance`.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -70,6 +70,8 @@ class OneRegionTimeseriesDataset:
 
     latest: Dict[str, Any]
 
+    provenance: Dict[str, str]
+
     def __post_init__(self):
         if CommonFields.LOCATION_ID in self.data.columns:
             region_count = self.data[CommonFields.LOCATION_ID].nunique()
@@ -105,12 +107,14 @@ class OneRegionTimeseriesDataset:
         rows_key = dataset_utils.make_rows_key(self.data, after=after,)
         columns_key = list(columns) if columns else slice(None, None, None)
         return OneRegionTimeseriesDataset(
-            self.data.loc[rows_key, columns_key].reset_index(drop=True), latest=self.latest,
+            self.data.loc[rows_key, columns_key].reset_index(drop=True),
+            latest=self.latest,
+            provenance=self.provenance,
         )
 
     def remove_padded_nans(self, columns: List[str]):
         return OneRegionTimeseriesDataset(
-            _remove_padded_nans(self.data, columns), latest=self.latest
+            _remove_padded_nans(self.data, columns), latest=self.latest, provenance=self.provenance,
         )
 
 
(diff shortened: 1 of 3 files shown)


libs/datasets/timeseries.py
```diff
@@ -479,9 +510,14 @@ def from_combined_dataframe(
 
     @staticmethod
     def from_csv(path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionTimeseriesDataset":
-        return MultiRegionTimeseriesDataset.from_combined_dataframe(
+        mrts = MultiRegionTimeseriesDataset.from_combined_dataframe(
             common_df.read_csv(path_or_buf, set_index=False)
         )
+        if isinstance(path_or_buf, pathlib.Path):
+            provenance_path = pathlib.Path(str(path_or_buf).replace(".csv", "-provenance.csv"))
+            if provenance_path.is_file():
```
ghop02, line 518: should this be `exists`?

libs/datasets/timeseries.py
```diff
@@ -479,9 +510,14 @@ def from_combined_dataframe(
 
     @staticmethod
     def from_csv(path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionTimeseriesDataset":
-        return MultiRegionTimeseriesDataset.from_combined_dataframe(
+        mrts = MultiRegionTimeseriesDataset.from_combined_dataframe(
```
ghop02, line 513: fwiw I'd prefer `dataset` over `mrts`, since mrts is pretty opaque and I think if you didn't see it created right by `MultiRegionTimeseriesDataset` it's not obvious that its the abbreviation for that

ghop02 (approved):

libs/datasets/timeseries.py
```diff
@@ -448,6 +457,25 @@ def append_latest_df(self, latest_df: pd.DataFrame) -> "MultiRegionTimeseriesDat
 
         return MultiRegionTimeseriesDataset(self.data, latest_df, provenance=self.provenance)
 
+    def append_provenance_csv(
+        self, path_or_buf: Union[pathlib.Path, TextIO]
+    ) -> "MultiRegionTimeseriesDataset":
```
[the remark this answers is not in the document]
TomGoBravo: Done

libs/datasets/timeseries.py
```diff
@@ -478,9 +506,17 @@ def from_combined_dataframe(
         return multiregion_timeseries
 
     @staticmethod
-    def from_csv(path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionTimeseriesDataset":
+    def from_csv(
+        path_or_buf: Union[pathlib.Path, TextIO], provenance: Optional[pd.Series] = None,
```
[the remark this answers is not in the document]
TomGoBravo: Oops, junk left from previous version of this PR. Removed provenance parameter

libs/datasets/timeseries.py
TomGoBravo: Hmmm, I guess that if a non-file exists with that name it is probably better to let pd.from_csv fail on it than silently ignore it. Changed.

libs/datasets/timeseries.py

PR #758, opened 2020-10-26
Record usage numbers in hubspot

ghop02:
This adds recording usage values to hubspot contacts, letting us better know who to reach out to and reduce reliance on the old spreadsheet


libs/update_api_user_metrics.py
```diff
@@ -18,17 +20,47 @@
...
 
 
+def update_hubspot_activity(email, latest_active_at, days_active):
+    """Updates Hubspot contact with latest activity dates."""
+    contacts_url = "https://api.hubapi.com/crm/v3/objects/contacts"
+
+    query_string = {"hapikey": HUBSPOT_API_KEY}
+
+    # Hubspot date field should be at UTC midnight.  Adding the timestamp parsing
+    # makes the date timestamp respect the UTC timezone.
```
TomGoBravo, line 42: "Adding the timestamp parsing makes the date timestamp respect the UTC timezone" -> "When the %z directive is provided to the strptime() method, a TZ aware datetime object will be produced.".... Though I think I'd use 
date.isoformat() and datetime.combine(... tzinfo=datetime.timezone.utc).

TomGoBravo (approved):
Some nits on date handling, otherwise LGTM

PR #759, opened 2020-10-26
Remove deaths from rt

ghop02:
Removes the code for handling deaths in the Rt calculation.

### DIFF
--- a/pyseir/rt/infer_rt.py
+++ b/pyseir/rt/infer_rt.py
@@ -1,8 +1,9 @@
+from typing import Optional
 import math
 from dataclasses import dataclass
 from datetime import timedelta
 import structlog
-from typing import Optional
+
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


pyseir/rt/infer_rt.py
```diff
@@ -115,95 +106,79 @@ def _generate_input_data(
             "the Infection Rate Metric",
             region=regional_input.display_name,
         )
-        return pd.DataFrame()
+        return None
```
BrettBoval, line 109: What's the thought of making the return type optional instead of returning an empty timeseries?

BrettBoval (approved):

pyseir/rt/infer_rt.py
ghop02: I like that since it's a series, it's a more atomic unit.  Feels like the resulting code paths make a bit more sense

BrettBoval: K. Seems like our personal opinions are different. I'm fine with it.

PR #760, opened 2020-10-27
Remove all webui uses

ghop02:
Removes all webui related output


BrettBoval (approved):
Gotta love a net -739 LOC commit.

PR #767, opened 2020-10-28
Add new cases to latest value data structure

ghop02:
With the way things work, we still need to manually maintain the latest value in the multi region dataset.

This adds logic to add new cases into that. 

I don't want to get in a habit of this custom logic.  @TomGoBravo and I have talked about ways to calculate that dataset on the fly better, but that logic requires more refactoring, and I'd like for it to be separate from getting new cases out.

### DIFF
--- a/libs/datasets/dataset_utils.py
+++ b/libs/datasets/dataset_utils.py
@@ -325,6 +325,5 @@ def build_latest_for_column(timeseries_df: pd.DataFrame, column: CommonFields) -
     assert CommonFields.LOCATION_ID in timeseries_df.columns
     assert CommonFields.DATE in timeseries_df.columns
 
-    data = timeseries_df.set_index([CommonFields.LOCATION_ID, CommonFields.DATE])
-    column_data = data[column].groupby([CommonFields.LOCATION_ID]).ffill()
-    return column_data.groupby([CommonFields.LOCATION_ID]).last()
+    data = timeseries_df.set_index([CommonFields.LOCATION_ID, CommonFields.DATE]).sort_index()
+    return data[column].groupby([CommonFields.LOCATION_ID], sort=False).last()
--- a/libs/datasets/dataset_utils.py
+++ b/libs/datasets/dataset_utils.py
@@ -311,3 +311,12 @@ def fips_index_geo_data(df: pd.DataFrame) -> pd.DataFrame:
     # are rows in the input data sources that have different values for county name, state etc.
     fips_indexed = all_identifiers.set_index(CommonFields.FIPS, verify_integrity=True)
     return fips_indexed
+
+
+def build_latest_for_column(timeseries_df: pd.DataFrame, column: CommonFields) -> pd.Series:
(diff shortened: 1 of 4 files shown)


libs/datasets/dataset_utils.py
```diff
@@ -311,3 +311,20 @@ def fips_index_geo_data(df: pd.DataFrame) -> pd.DataFrame:
...
+    Args:
+        timeseries_df: Timeseries DF with location_id and date columns.
+        column: Column to build latest value for.
+
+    Returns: Series indexed by location_id with the latest value for `column`.
+    """
+    assert CommonFields.LOCATION_ID in timeseries_df.columns
+    assert CommonFields.DATE in timeseries_df.columns
+
+    data = timeseries_df.set_index([CommonFields.LOCATION_ID, CommonFields.DATE])
```
TomGoBravo, line 328: add `sort_index()` and maybe a test that exercises handling data not in the right order.

libs/datasets/dataset_utils.py
```diff
@@ -311,3 +311,20 @@ def fips_index_geo_data(df: pd.DataFrame) -> pd.DataFrame:
...
+    column_data = data[column].groupby([CommonFields.LOCATION_ID]).ffill()
```
TomGoBravo, line 329: Do you need to ffill if you have the series sorted by date and use `groupby(sort=False)`? I think last() may skip NAs.

test/libs/datasets/timeseries_test.py
```diff
@@ -325,12 +328,18 @@ def test_calculate_new_cases():
...
+            "iso1:us#fips:2,,,2.0\n"
+            "iso1:us#fips:3,,,9.0\n"
+            "iso1:us#fips:4,,,\n"
         )
     )
 
-    mrts_after = timeseries.add_new_cases(mrts_before)
-    pd.testing.assert_frame_equal(mrts_after.data, mrts_expected.data, check_like=True)
+    timeseries_after = timeseries.add_new_cases(mrts_before)
+    pd.testing.assert_frame_equal(timeseries_after.data, mrts_expected.data, check_like=True)
```
TomGoBravo, line 339: how about `assert_combined_like(timeseries_after, mrts_expected)`?

TomGoBravo (changes requested):

libs/datasets/dataset_utils.py
ghop02: ooh cool, yeah looks like it does

libs/datasets/timeseries.py
```diff
@@ -753,6 +753,14 @@ def _diff_preserving_first_value(series):
     return series_diff
 
 
+def _add_new_cases_to_latest(timeseries_df: pd.DataFrame, latest_df: pd.DataFrame) -> pd.DataFrame:
```
TomGoBravo, line 756: how about adding `assert latest_df.index.names == [CommonFields.LOCATION_ID]` to catch when we change it and communicate the expectation to people reading the code?

TomGoBravo (approved):

PR #769, opened 2020-10-29
Add outlier detection

ghop02:


libs/datasets/timeseries.py
```diff
@@ -782,6 +782,39 @@ def add_new_cases(timeseries: MultiRegionTimeseriesDataset) -> MultiRegionTimese
     return new_timeseries
 
 
+def _calculate_modified_zscore(series: pd.Series, window: int = 10) -> pd.Series:
+    """Calculates zscore removing zeros"""
+    series = series.copy()
+    series[series == 0] = None
+    rolling_series = series.rolling(window=window, min_periods=3)
```
BrettBoval, line 789: Make the default explicit either here or in the docstring that this is a trailing, right inclusive equally weighted window.

libs/datasets/timeseries.py
```diff
@@ -782,6 +782,39 @@ def add_new_cases(timeseries: MultiRegionTimeseriesDataset) -> MultiRegionTimese
     return new_timeseries
 
 
+def _calculate_modified_zscore(series: pd.Series, window: int = 10) -> pd.Series:
+    """Calculates zscore removing zeros"""
```
BrettBoval, line 786: I think it's worth documenting the implementation for those that don't deal with pandas abstractions often. Some of the edge cases feel like unwanted behavior.

Some things to include:
- Trailing equally weighted window based only on valid measurements
- 0 is not treated as a valid measurement (add why?)
- Needs at least 3 valid (non-NaN, non-Zero) measurements in the window to return any value
- Time shifts the values forward one day (I don't understand why this is yet, feels like an off by one error but maybe I haven't found the right index alignments)
- No minimum positive value of cases needed for thresholding (pd.Series([1,1,1,2,1]) triggers right now, which feels wrong, I think that's just a 1/0 -> inf shifted to the right one but I don't know)

Given the number of different people that will want to know, and our push towards transparency in how we handle missing data, I think this needs to be much more explicit.

libs/datasets/timeseries.py
```diff
@@ -782,6 +782,39 @@ def add_new_cases(timeseries: MultiRegionTimeseriesDataset) -> MultiRegionTimese
...
+    series = series.copy()
+    series[series == 0] = None
+    rolling_series = series.rolling(window=window, min_periods=3)
+    # Shifting one to exclude current datapoint
+    mean = rolling_series.mean().shift(1)
+    std = rolling_series.std(ddof=0).shift(1)
+    z = (series - mean) / std
```
BrettBoval, line 793: What happens when std = 0, how do we handle the inf downstream?

libs/datasets/timeseries.py
```diff
@@ -782,6 +782,39 @@ def add_new_cases(timeseries: MultiRegionTimeseriesDataset) -> MultiRegionTimese
     return new_timeseries
 
 
+def _calculate_modified_zscore(series: pd.Series, window: int = 10) -> pd.Series:
+    """Calculates zscore removing zeros"""
+    series = series.copy()
+    series[series == 0] = None
+    rolling_series = series.rolling(window=window, min_periods=3)
+    # Shifting one to exclude current datapoint
+    mean = rolling_series.mean().shift(1)
```
BrettBoval, line 791: Not clear why we shift these forward one day. If we don't want to calculate for today, why not drop the last value in the column?

libs/datasets/timeseries.py
BrettBoval: If the time shift by 1 is to get the trailing sample mean exclusive of the new candidate point, then I'd prefer we just made that window shape more explicit.

BrettBoval (changes requested):
Let's add more context around the edge cases, but I'm liking the direction.

libs/datasets/timeseries.py
ghop02: ahh had to think about it a bit but it's because we're running the calculation for every day.  so all of the means/std have to be shifted for the z-score calculation

libs/datasets/timeseries.py
ghop02: hahah i bet there is never a timeseries that consistent

ghop02: but it's a good point i'll look into it

test/libs/datasets/timeseries_test.py
```diff
@@ -580,3 +579,51 @@ def test_merge_provenance():
...
+
+    for i, value in enumerate(values):
+        rows.append({CommonFields.LOCATION_ID: location_id, "date": dates[i], column: value})
+
+    # Find last non nan value, which simulates building the latest value entry
+    for value in reversed(values):
+        if not pd.isna(value) and value is not None:
+            break
+
+    rows.append({CommonFields.LOCATION_ID: location_id, "date": None, column: value})
```
TomGoBravo, line 599: None -> pd.NaT

test/libs/datasets/timeseries_test.py
```diff
@@ -580,3 +579,51 @@ def test_merge_provenance():
...
+    dataset = _build_one_column_multiregion_dataset(CommonFields.NEW_CASES, values)
+    dataset = timeseries.drop_new_case_outliers(dataset)
+
+    # Expected result is the same series with the last value removed
+    values = [10.0] * 7 + [None]
+    expected = _build_one_column_multiregion_dataset(CommonFields.NEW_CASES, values)
+    assert_combined_like(dataset, expected)
+
+
+def test_remove_outliers_threshold():
```
TomGoBravo, line 616: how about making sure a NEW_CASES series with all NAs, with one real value and two identical values work?

TomGoBravo (approved):

BrettBoval (commented):
lgtm

PR #770, opened 2020-10-29
Remove negative new cases

ghop02:
Sometimes the new case series can be negative, to preserve the existing behavior this makes those nans.  It looks like this generally happens if a region switches from reporting probable -> confirmed cases or small adjustments.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -770,9 +770,12 @@ def add_new_cases(timeseries: MultiRegionTimeseriesDataset) -> MultiRegionTimese
     # We want to capture the first day a region reports a case. Since our data sources have
     # been capturing cases in all states from the beginning of the pandemic, we are treating
     # The first days as appropriate new case data.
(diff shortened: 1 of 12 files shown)


libs/datasets/timeseries.py
```diff
@@ -770,9 +770,12 @@ def add_new_cases(timeseries: MultiRegionTimeseriesDataset) -> MultiRegionTimese
     # We want to capture the first day a region reports a case. Since our data sources have
     # been capturing cases in all states from the beginning of the pandemic, we are treating
     # The first days as appropriate new case data.
-    df_copy[CommonFields.NEW_CASES] = grouped_df[CommonFields.CASES].apply(
-        _diff_preserving_first_value
-    )
+    new_cases = grouped_df[CommonFields.CASES].apply(_diff_preserving_first_value)
+
+    # Remove the occasional negative case adjustments.
+    new_cases[new_cases < 0] = None
```
TomGoBravo, line 776: Maybe change None to pd.NA

TomGoBravo (approved):

PR #771, opened 2020-10-29
Valorum usafacts for cases and deaths in datasets_combined

TomGoBravo:
* add reading data/cases-covid-county-data/timeseries-usafacts.csv for cases and deaths
* add usafacts as highest priority source of cases and deaths info


TomGoBravo:
No longer needed, thanks to Debbie. :tada:

PR #774, opened 2020-10-29
Rm summary as its own file. Reduce FIPS use

TomGoBravo:
* removes qa/dataset_summary.py. The summary is still written as part of the wide dates file.
* change calculate_new_case_data_by_region to get only DATE column, not FIPS (via TimeseriesDataset.INDEX_FIELDS)


libs/qa/dataset_summary.py
```diff
@@ -18,35 +15,22 @@
...
-    CommonFields.LOCATION_ID,
+    CommonFields.FIPS,
 ]
 
 VARIABLE_FIELD = "variable"
 
 SUMMARY_PATH = dataset_utils.DATA_DIRECTORY / "timeseries_summary.csv"
 
 
-class TimeseriesSummary(pydantic.BaseModel):
```
ghop02, line 29: Should we just remove the dataset summary? Don't really think it's used anymore and we could always bring it back if needed

ghop02 (approved):

libs/qa/dataset_summary.py
TomGoBravo: Done! Please take another look.

ghop02 (approved):

PR #775, opened 2020-10-31
RFC: Install numexpr module per pandas recommendations.

mikelehen:
Adds numexpr 2.7.1 to our dependencies.  This is [recommended](https://pandas.pydata.org/pandas-docs/stable/getting_started/install.html#install-recommended-dependencies) by pandas and [according to the docs](https://pandas.pydata.org/pandas-docs/stable/user_guide/enhancingperf.html#enhancingperf-eval), "To benefit from using eval() you need to install numexpr."

That said, this PR has no meaningful perf impact (on `run.py data updata` or on full snapshot generation).  And in fact, to make our code work with numexpr I had to change one of our 2 uses of `eval()` and in doing so and measuring the perf impact I discovered that:
1. numexpr didn't affect the performance.
2. Avoiding eval() actually had better performance (saved ~0.07s, taking the CovidCountyDataDataSource.synthesize_test_metrics() method from ~0.94s to ~0.88s).

**So thoughts on whether we should bother checking this in given it has no tangible benefit other than following pandas recommendations?**

### DIFF
--- a/libs/datasets/sources/covid_county_data.py
+++ b/libs/datasets/sources/covid_county_data.py
@@ -72,18 +72,23 @@ class CovidCountyDataDataSource(data_source.DataSource):
 
         # Make a copy to avoid modifying the argument when using mask with inplace=True.
         df = data.copy()
-        # HACK: isnan() is only available in numexpr3 which isn't released yet. So we use nan != nan trick.
-        # https://github.com/pydata/numexpr/issues/23#issuecomment-58774009
-        isna = lambda column: f"({column} != {column})"
-        notna = lambda column: f"({column} == {column})"
-        tests_and_cases = df.eval(
-            f"{isna('negative_tests')} & {isna('positive_tests')} & {notna('total_tests')} & {notna('cases')}"
+        tests_and_cases = (
+            df["negative_tests"].isna()
+            & df["positive_tests"].isna()
+            & df["total_tests"].notna()
+            & df["cases"].notna()
         )
-        missing_neg = df.eval(
-            f"{isna('negative_tests')} & {notna('positive_tests')} & {notna('total_tests')} & {notna('cases')}"
+        missing_neg = (
+            df["negative_tests"].isna()
+            & df["positive_tests"].notna()
+            & df["total_tests"].notna()
+            & df["cases"].notna()
         )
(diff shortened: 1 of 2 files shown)


libs/datasets/sources/covid_county_data.py
```diff
@@ -72,14 +72,23 @@ def synthesize_test_metrics(cls, data: pd.DataFrame) -> Tuple[pd.DataFrame, pd.S
 
         # Make a copy to avoid modifying the argument when using mask with inplace=True.
         df = data.copy()
-        tests_and_cases = df.eval(
-            "negative_tests.isna() & positive_tests.isna() & total_tests.notna() & cases.notna()"
```
mikelehen, line 76: negative_tests.isna() returns an `ndarray` boolean mask indicating which rows are missing negative_tests, but numexpr doesn't support `ndarray` and indeed the point of using eval() and numexpr is so it can execute the operation on a row-by-row basis and get better cache utilization, etc (_I think!_).  So to make this work with numexpr I had to rewrite it, which unfortunately ended up really ugly because numexpr doesn't have a `isnan()` function yet (coming in numexpr3 which has been stalled for over a year).  But if you want to see what that looked like, see here: https://github.com/covid-projections/covid-data-model/pull/775/commits/1bd07996fb3846a3d8498f5b29c11d7cd29a94f1#diff-bcae9bb794eb90c4637980f820d3df847a66c5392b15fdbd6577e8b278d916c6R75

But as I mentioned in the PR description that ended up being equivalent speed-wise (and this code isn't a bottleneck in the first place).

So I ended up dropping eval() entirely which was both faster and more readable.

mikelehen:
@TomGoBravo @ghop02 Either of you have thoughts on if this is worth merging?  It doesn't have any meaningful perf impact but it follows pandas recommendations to include the `numexpr` module. 🤷 

Happy to drop it if we don't want to bother.

ghop02 (approved):
This looks good to me! I don't really have strong opinions on this either.  I think by adding numba, numexpr actually gets included as well, so probably better to make it explicit

TomGoBravo (approved):
Thank you for checking this out. I'm neutral on this change so happy to support your improvement.

PR #778, opened 2020-11-02
Use numba to speed up process matrix

ghop02:
This PR uses numba to speed up a scipy pdf function.  It speeds up the function call from 8ms per call to ~2ms per call. I checked and the input and output of the function was the same.

Loading the regions in parallel shaves off around 2 min of the run.  Now all of pyseir (rt calculation + icu calc) runs in 2 min 33 seconds.


mikelehen (approved):
LGTM.  I'm not fluent in the numba code, but if the output is the same and it's faster, I am happy. 😁

PR #780, opened 2020-11-02
Use CMS as backfill for test positivity, and disable questionable counties from Valorum

mikelehen:
* Adds CMSTesting dataset.
* Modifies Valorum dataset to drop test data from states we don't trust.
* Adds a `PassThruMethod` to test_positivity.py that lets us use the `test_positivity` column as-is.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -682,7 +682,9 @@ class MultiRegionTimeseriesDataset(SaveableDatasetInterface):
             provenance_path = str(path).replace(".csv", "-provenance.csv")
             self.provenance.sort_index().rename(PdFields.PROVENANCE).to_csv(provenance_path)
 
-    def join_columns(self, other: "MultiRegionTimeseriesDataset") -> "MultiRegionTimeseriesDataset":
+    def join_columns(
+        self, other: "MultiRegionTimeseriesDataset", replace: bool = False
+    ) -> "MultiRegionTimeseriesDataset":
         """Joins the timeseries columns in `other` with those in `self`."""
         if not other.latest_data.empty:
             raise NotImplementedError("No support for joining other with latest_data")
@@ -694,8 +696,11 @@ class MultiRegionTimeseriesDataset(SaveableDatasetInterface):
         )
         common_ts_columns = other_ts_columns & set(self.data_with_fips.columns)
         if common_ts_columns:
-            # columns to be joined need to be disjoint
-            raise ValueError(f"Columns are in both dataset: {common_ts_columns}")
+            if replace:
+                self_df = self_df.drop(common_ts_columns, axis="columns")
+            else:
+                # columns to be joined need to be disjoint
+                raise ValueError(f"Columns are in both dataset: {common_ts_columns}")
         common_geo_columns = list(set(self.data_with_fips.columns) & other_geo_columns)
         # TODO(tom): fix geo columns check, no later than when self.data is changed to contain only
         # timeseries
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -682,17 +682,25 @@ class MultiRegionTimeseriesDataset(SaveableDatasetInterface):
             provenance_path = str(path).replace(".csv", "-provenance.csv")
             self.provenance.sort_index().rename(PdFields.PROVENANCE).to_csv(provenance_path)
 
-    def join_columns(
-        self, other: "MultiRegionTimeseriesDataset", replace: bool = False
-    ) -> "MultiRegionTimeseriesDataset":
+    def drop_column_if_present(self, column: str) -> "MultiRegionTimeseriesDataset":
+        """Drops the specified column from the timeseries if it exists"""
+        df = self.data_with_fips.drop(column, axis="columns", errors="ignore")
+        latest_data = self.latest_data_with_fips.drop(
+            column, axis="columns", errors="ignore"
+        ).reset_index()
+        provenance = self.provenance[
+            self.provenance.index.get_level_values(PdFields.VARIABLE) != column
+        ]
+        return MultiRegionTimeseriesDataset.from_timeseries_df(
+            df, provenance=provenance
+        ).append_latest_df(latest_data)
+
+    def join_columns(self, other: "MultiRegionTimeseriesDataset") -> "MultiRegionTimeseriesDataset":
         """Joins the timeseries columns in `other` with those in `self`.
 
         Args:
             other: The timeseries dataset to join with `self`. all columns except the "geo" columns
                    will be joined into `self`.
-            replace: Whether to allow `other` to contain columns that already exist in `self`.
-                     If set to True, the existing columns will be replaced. Else, an error will
-                     be thrown.
         """
         if not other.latest_data.empty:
             raise NotImplementedError("No support for joining other with latest_data")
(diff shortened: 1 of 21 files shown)


libs/datasets/sources/hhs_testing_dataset.py
```diff
@@ -2,20 +2,15 @@
...
     DATA_PATH = "data/testing-hhs/timeseries-common.csv"
 
-    INDEX_FIELD_MAP = {
-        CommonFields.DATE: CommonFields.DATE,
-        CommonFields.AGGREGATE_LEVEL: CommonFields.AGGREGATE_LEVEL,
-        CommonFields.COUNTRY: CommonFields.COUNTRY,
-        CommonFields.STATE: CommonFields.STATE,
-        CommonFields.FIPS: CommonFields.FIPS,
-    }
+    INDEX_FIELD_MAP = {f: f for f in TimeseriesDataset.INDEX_FIELDS}
```
mikelehen, line 13: minor cleanup; no behavioral change

libs/datasets/timeseries.py
```diff
@@ -682,7 +682,9 @@ def to_csv(self, path: pathlib.Path):
             provenance_path = str(path).replace(".csv", "-provenance.csv")
             self.provenance.sort_index().rename(PdFields.PROVENANCE).to_csv(provenance_path)
 
-    def join_columns(self, other: "MultiRegionTimeseriesDataset") -> "MultiRegionTimeseriesDataset":
+    def join_columns(
+        self, other: "MultiRegionTimeseriesDataset", replace: bool = False
+    ) -> "MultiRegionTimeseriesDataset":
```
mikelehen, line 687: We now have test_positivity in the input combined dataset (as read from CMS) but then join in the test_positivity output from test_positivity.py, so I changed this to allow replacing the existing column (instead of erroring).

Alternatively we could have two different test_positivity columns, but I'm not sure if that's better or what to name them (test_positivity_in and test_positivity_out, or something?)

libs/test_positivity.py
```diff
@@ -21,35 +22,83 @@
 _log = structlog.get_logger()
 
 
-@dataclasses.dataclass
-class Method:
+class Method(ABC):
```
mikelehen, line 25: I refactored this to have a `Method` base class and a `DivisionMethod` and `PassThruMethod` derived classes.  python OO is new to me so holler if this doesn't seem right.

libs/test_positivity.py
```diff
@@ -21,35 +22,83 @@
...
 
     @property
+    @abstractmethod
     def columns(self) -> Set[FieldName]:
-        return {self.numerator, self.denominator}
+        pass
 
-    def calculate(self, delta_df: pd.DataFrame) -> pd.DataFrame:
+    @abstractmethod
+    def calculate(self, df: pd.DataFrame, delta_df: pd.DataFrame) -> pd.DataFrame:
```
mikelehen, line 39: We now pass in the "normal" dataframe as well as the "delta" dataframe, for the pass-thru method.

libs/test_positivity.py
```diff
@@ -115,19 +164,23 @@ def run(
             .reindex(columns=input_date_range)
             .rename_axis(columns=CommonFields.DATE)
         )
-        # This calculates the difference only when the cumulative value is a real value `diff_days` apart.
-        # It looks like our input data has few or no holes so this works well enough.
-        diff_df = input_wide.diff(periods=diff_days, axis=1)
 
         methods_with_data = AllMethods._methods_with_columns_available(
-            methods, diff_df.index.get_level_values(PdFields.VARIABLE).unique()
+            methods, input_wide.index.get_level_values(PdFields.VARIABLE).unique()
```
mikelehen, line 169: I tweaked this so we check for column availability on `df` instead of `diff_df`...  since we don't care about `diff_df` in the pass-thru method.  It would have worked as-is, but this seems more straightforward.  Now `diff_df` is passed into `calculate()` in case it needs it but it's not used for any other logic.

libs/top_level_metrics.py
```diff
@@ -22,7 +22,8 @@
 RT_TRUNCATION_DAYS = 7
 
 
-MAX_METRIC_LOOKBACK_DAYS = 7
+# CMS and HHS testing data can both lag by more than 7 days. Let's use it unless it's 2 weeks old.
+MAX_METRIC_LOOKBACK_DAYS = 13
```
mikelehen, line 26: cc/ @ghop02 FYI- I'm changing to allow for more lag, since HHS seems to lag by more than 7 days sometimes and we know CMS will lag by 7-10 days in many cases.

test/combined_dataset_test.py
```diff
@@ -59,8 +59,8 @@ def test_combined_county_has_some_data(fips):
     assert latest.get_record_for_fips(fips=fips)[CommonFields.DEATHS] > 1
 
 
-# Check some counties picked arbitrarily: (Marion County, IN)/18097 and (Harris County, TX)/48201
-@pytest.mark.parametrize("fips", ["18097", "48201"])
+# Check some counties picked arbitrarily: (Orange County, CA)/06059 and (Harris County, TX)/48201
+@pytest.mark.parametrize("fips", ["06059", "48201"])
```
mikelehen, line 63: Had to change this since we're disabling the valorum positive_tests / negative_tests for Indiana.

test/libs/datasets/sources/covid_county_data_test.py
```diff
@@ -14,9 +14,9 @@
 
 def test_fix_tests_and_cases():
     df = read_csv_and_index_fips_date(
-        "fips,date,negative_tests,positive_tests,total_tests,cases\n"
-        "97123,2020-04-01,9,1,10,1\n"
-        "97123,2020-04-02,,,20,2\n"
+        "fips,state,date,negative_tests,positive_tests,total_tests,cases\n"
+        "97123,,2020-04-01,9,1,10,1\n"
+        "97123,,2020-04-02,,,20,2\n"
```
mikelehen, line 19: `CovidCountyDataDataSource.synthesize_test_metrics()` now relies on the `state` column in order to do the disabling...

test/libs/top_level_metrics_test.py
```diff
@@ -342,7 +342,7 @@ def test_lookback_days():
     data = _build_metrics_df(
         f"2020-08-12,36,10,0.1,0.06,{prev_rt},{prev_rt_ci90}\n"
         f"2020-08-13,36,10,0.1,0.06,,\n"
-        "2020-08-20,,,,,,\n"
+        "2020-08-26,,,,,,\n"
```
mikelehen, line 345: had to adjust this since I changed lookback from 7 days to 13.

ghop02 (commented):
Took a quick pass through.  Structure looks good to me - didn't look too much into some of the tougher questions.  I do wonder about the join with duplicate column name, but probably easier to talk with @TomGoBravo

mikelehen (commented):
Thanks @ghop02! Good suggestions.

libs/datasets/timeseries.py
TomGoBravo: Having two test_positivity names seems pretty confusing. Another alternative is having test_positivity.py read the dataset instead of putting TEST_POSITIVITY in combined datasets.
How about `ts.drop_column_if_present(CommonFields.TEST_POSITIVITY).join_columns(test_positivity_results.test_positivity)`? It seems odd that you want to get test positivity for all regions from combined datasets or test_positivity_results instead of blending per region.

libs/test_positivity.py
TomGoBravo: Ha, I made a very similar change in a branch while fiddling with this.

TomGoBravo (approved):

libs/datasets/timeseries.py
mikelehen: Thanks!  I'm down with `drop_column_if_present()`.  Added.

libs/datasets/timeseries.py
```diff
@@ -682,17 +682,25 @@ def to_csv(self, path: pathlib.Path):
             provenance_path = str(path).replace(".csv", "-provenance.csv")
             self.provenance.sort_index().rename(PdFields.PROVENANCE).to_csv(provenance_path)
 
-    def join_columns(
-        self, other: "MultiRegionTimeseriesDataset", replace: bool = False
-    ) -> "MultiRegionTimeseriesDataset":
+    def drop_column_if_present(self, column: str) -> "MultiRegionTimeseriesDataset":
```
TomGoBravo, line 685: A unit test that exercises this would be nice.

TomGoBravo (approved):

libs/datasets/timeseries.py
mikelehen: Yeah. I will try to do that tomorrow.  Sorry!

PR #781, opened 2020-11-02
Trigger label api on morning build

ghop02:
Adds a trigger to run label-api automatically on the morning build


mikelehen (approved):
This looks okay based on our precedents, but I actually think we can/should do some major cleanup...

1. deploy_api.yml should take arguments for whether to kick off web snapshot PR and API labelling.
2. We can pass in the right args from the daily_build.yml scheduled job (I guess we might have to split it into two scheduled jobs if we want to have only the morning one kick off the extra stuff).
3. We should make deploy_api.yml into a workflow_dispatch so it's easy for people to kick off jobs and they can optionally decide if it should trigger web snapshot and API labelling.

Oh, and can you double-check if the deploy playbook needs any updates based on this change?

PR #783, opened 2020-11-04
MultiRegion append_regions preserves provenance

TomGoBravo:
* change `MultiRegionTimeseriesDataset.provenance` to never be None and always sorted, simplifying code that uses it
* in `MultiRegionTimeseriesDataset.append_regions` also append the provenance data.

This PR addresses https://trello.com/c/7tgdgPUV/547-provenance-file-stopped-updating-oct-28


TomGoBravo:
[Provenance file stopped updating Oct 28](https://trello.com/c/7tgdgPUV/547-provenance-file-stopped-updating-oct-28)

libs/datasets/timeseries.py
```diff
@@ -567,18 +583,27 @@ def __post_init__(self):
...
-            pd.concat([self.data, other.data], ignore_index=True)
-        ).append_latest_df(
-            pd.concat(
-                [self.latest_data.reset_index(), other.latest_data.reset_index()], ignore_index=True
+        return (
+            MultiRegionTimeseriesDataset.from_timeseries_df(
+                pd.concat([self.data, other.data], ignore_index=True)
+            )
+            .append_latest_df(
+                pd.concat(
```
ghop02, line 597: definitely not necessary, but making these dataframes their own variables would make this block more concise

ghop02 (approved):
Seems good to me! would prefer updating the files in data/ via the github action, but not too worried

libs/datasets/timeseries.py
TomGoBravo: Done in a0bdee18

PR #784, opened 2020-11-04
rm MultiRegionTimeseriesDataset.from_combined_dataframe

TomGoBravo:
This PR removes `MultiRegionTimeseriesDataset.from_combined_dataframe` which is useful for parsing a CSV with metrics in columns but wasn't otherwise useful. Removing it reduces the number of ways that a `MultiRegionTimeseriesDataset` can be created.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -490,41 +490,23 @@ class MultiRegionTimeseriesDataset(SaveableDatasetInterface):
         )
 
     @staticmethod
-    def from_combined_dataframe(
-        combined_df: pd.DataFrame, provenance: Optional[pd.Series] = None
-    ) -> "MultiRegionTimeseriesDataset":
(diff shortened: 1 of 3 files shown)


libs/datasets/timeseries.py
```diff
@@ -639,9 +621,9 @@ def get_counties(
         # metrics which no longer have real values are excluded.
         latest_df, provenance = self._get_latest_and_provenance_for_locations(location_ids)
 
-        return MultiRegionTimeseriesDataset.from_combined_dataframe(
-            pd.concat([ts_df, latest_df], ignore_index=True), provenance=provenance
-        )
+        return MultiRegionTimeseriesDataset.from_timeseries_df(
+            ts_df, provenance=provenance
+        ).append_latest_df(latest_df)
```
ghop02, line 626: Small but would love to find a way to remove this eventually.  Don't think we're quite there yet - just a thought!

ghop02 (approved):

PR #788, opened 2020-11-05
Stop updating api v1

ghop02:
Stops running updates for api v1.  The existing files in /latest should still remain, meaning you will still be able to download the existing api, but it will no longer update.


mikelehen (approved):

TomGoBravo (approved):

mikelehen:
✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅

PR #792, opened 2020-11-09
MultiRegion more private attributes

TomGoBravo:
This PR simplifies the MultiRegionTimeseriesDataset interface, easing future changes to it.

-  mark latest_data and provenance attributes as private with single underscore
-  replace use of `from_timeseries_df(provenance=...)` with call to new add_provenance_series method
-  in `test_recent_days replace` direct access `to _provenance` with getting it from get_one_region return value.
-  rename append_latest_df to add_latest_df
-  rename append_provenance_csv to add_provenance_csv
-  remove some access to latest values in effort to remove latest values
-  drop_column_if_present using _latest_data instead of latest_data_with_fips


ghop02 (approved):
Seems alright to me.  would like to move away from the `add_latest_df` functions and do them in one step but not for this PR

TomGoBravo:
You mean one method to add/join/append datasets instead of handling each attribute on its own?

PR #794, opened 2020-11-10
add MultiRegionDataset.get_subset

TomGoBravo:
## This PR

This PR removes another place where we depend on a separate latest values object. It prepares us for a PR that removes the latest values as its own first class thingy.

- replace use of combined_datasets.get_subset_regions with combined_datasets.load_us_timeseries_dataset().get_subset which is a faster way to iterate through the OneRegionTimeseriesDataset.

## Tested

For the main pipeline, the end_to_end tests pass.

For combined data: checked out public data repo sha 566c18c4 (as listed in data/multiregion.json), ran `python ./run.py data update` and checked that `git status` showed no changes in `data/*csv`


cli/api.py
```diff
@@ -126,7 +124,8 @@ def generate_api(input_dir, output, summary_output, aggregation_level, state, fi
 
     active_states = [state.abbr for state in us.STATES]
     active_states = active_states + ["PR", "MP"]
-    regions = combined_datasets.get_subset_regions(
+
+    selected_dataset = combined_datasets.load_us_timeseries_dataset().get_subset(
```
ghop02, line 128: I wonder when the right time to remove this code is.  Now that the CRS is using the new api, maybe soon?  Also may be good to refresh api v1 once before we remove the code

cli/api.py
```diff
@@ -197,10 +197,10 @@ def generate_api_v2(model_output_dir, output, level, state, fips):
     us_timeseries = combined_datasets.load_us_timeseries_dataset()
 
     # Load all API Regions
-    regions = list(
-        us_timeseries.iter_regions(level=level, exclude_county_999=True, state=state, fips=fips)
+    selected_dataset = combined_datasets.load_us_timeseries_dataset().get_subset(
```
ghop02, line 200: us_timeseries is already loaded above, can you use that?

ghop02 (approved):

cli/api.py
TomGoBravo: removed extra local in d07d90cd

cli/api.py
TomGoBravo: Sounds good to me.

PR #795, opened 2020-11-11
Add cbsa to docs + add API Updates Section

ghop02:
This adds sections about CBSAs in the API docs. 

Additionally I added an API Updates section to the API docs, giving us a place we can keep a changelog of additions to the API. 

Also, I removed a markdown file we're no longer using


ghop02:
Can check out the new API Updates page [here](https://5fac1b98c9e1f30008beb62e--can-api-docs.netlify.app/updates)

TomGoBravo (approved):
Looking at https://5fac1b98c9e1f30008beb62e--can-api-docs.netlify.app/api#tag/CBSA-Data/paths/~1cbsa~1{cbsa_code}.json?apiKey={apiKey}/get i think fips and state could do with explanation of what they mean in CBSA context.

PR #797, opened 2020-11-11
MultiRegion regional attributes

TomGoBravo:
## This PR

* Instead of calculating the latest value for each timeseries before combining data do it when the latest value is read from MultiRegionDataset
* Replace MultiRegionDataset data (geo data and timeseries together) and latest_values (timeseries latest values and constants such as population together) with `timeseries` (only timeseries of numeric values) and `static` (geo data and constants)
* Remove NEW_CASES logic for updating latest values, which is no longer needed.
* add parameter to_csv(write_timeseries_latest_values) to help with testing
* Add backports.cached_property and use it for non-trivial properties

## Testing

Manually set to_csv(write_timeseries_latest_values=True), ran `python ./run.py data update` and then `git difftool -t vimdiff data/multiregion.csv` to see what changed. This showed that county and state data was identical using the new latest value logic as the old logic. CBSA recent values changed, probably because aggregation is broken, but not an issue because these regions are in beta; fixing them is my next task and will be easier after this is in.

I ran built snapshot in this branch using the same data as a recent master snapshot. https://covid-projections-git-bump-snapshot-1316-326.covidactnow.now.sh/internal/compare/?left=1336&locations=2&metric=3&right=1335&sort=3 shows no differences in any timeseries produced for the website.


pyseir/cli.py
```diff
@@ -164,13 +164,13 @@ def _write_pipeline_output(
         output_dir_path.mkdir()
 
     infection_rate_metric_df = pd.concat((p.infer_df for p in pipelines), ignore_index=True)
-    multiregion_rt = MultiRegionDataset.from_timeseries_df(infection_rate_metric_df)
+    multiregion_rt = MultiRegionDataset.from_geodata_timeseries_df(infection_rate_metric_df)
```
ghop02, line 167: I'm not sure I would expect this to be a geodata timeseries?

libs/datasets/timeseries.py
```diff
@@ -877,10 +924,10 @@ def drop_regions_without_population(
     known_location_id_to_drop: Sequence[str],
     log: Union[structlog.BoundLoggerBase, structlog._config.BoundLoggerLazyProxy],
 ) -> MultiRegionDataset:
-    # latest_population is a Series with location_id index
-    latest_population = mrts._latest_data[CommonFields.POPULATION]
-    locations_with_population = mrts._latest_data.loc[latest_population.notna()].index
-    locations_without_population = mrts._latest_data.loc[latest_population.isna()].index
+    assert mrts._regional_attributes.index.names == [CommonFields.LOCATION_ID]
```
ghop02, line 927: Should this be part of the public api? don't love accessing the "private" property outside of the class

libs/test_positivity.py
```diff
@@ -207,8 +207,8 @@ def run(
         )
         positivity = first.drop(columns=[PdFields.VARIABLE])
 
-        test_positivity = MultiRegionDataset.from_timeseries_df(
-            positivity.stack().rename(CommonFields.TEST_POSITIVITY).reset_index(),
+        test_positivity = MultiRegionDataset.from_geodata_timeseries_df(
```
ghop02, line 210: why is this from a geodata timeseries?

ghop02 (approved):
This is a good next step! 

One thing I'd like to work on is figuring out the best ways to remove the necessity for the various geodata columns and adding accessors onto multiregion dataset and one region dataset.  for example, places where we group by state, can we have a method `groupby_parent_region` that walks up the location_id one level.

another thing I would love to think about how to do is push more of the data to the region itself.  For example: `region.county` and `region.state` seem more well placed than having to pull from a dictionary of that data. I think that would help reduce some of the logic around that shuffling

pyseir/cli.py
TomGoBravo: I think infection_rate_metric_df has some static data, such as aggregation level so it is geodata and timeseries in one DataFrame.

libs/datasets/timeseries.py
```diff
@@ -388,21 +424,30 @@ class MultiRegionDataset(SaveableDatasetInterface):
...
+    _regional_attributes: pd.DataFrame = _EMPTY_REGIONAL_ATTRIBUTES_DF
 
     # `provenance` is an array of str with a MultiIndex with names LOCATION_ID and VARIABLE.
-    _provenance: pd.Series
+    _provenance: pd.Series = _EMPTY_PROVENANCE_SERIES
+
+    # `data` has a simple integer index and columns from CommonFields. DATE and LOCATION_ID must
+    # be non-null in every row.
+    @property
+    def data(self) -> pd.DataFrame:
```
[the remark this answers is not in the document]
TomGoBravo: Added TODO to remove it.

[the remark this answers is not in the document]
TomGoBravo: Changing all callers is beyond scope of this PR.

libs/datasets/timeseries.py
TomGoBravo: > This is a good next step!
> 
> One thing I'd like to work on is figuring out the best ways to remove the necessity for the various geodata columns and adding accessors onto multiregion dataset and one region dataset. for example, places where we group by state, can we have a method `groupby_parent_region` that walks up the location_id one level.

> 
> another thing I would love to think about how to do is push more of the data to the region itself. For example: `region.county` and `region.state` seem more well placed than having to pull from a dictionary of that data. I think that would help reduce some of the logic around that shuffling


Yes, I'm trying to add accessors onto multiregion dataset and one region dataset to reduce direct use of the dataframes. One big benefit of having all the data in DataFrames (at least internal MultiRegionDataset) is it provides an easy fast way to copy to and from disk.

TomGoBravo: Accessing private / single underscore attributes outside the class is a bit ugly, but I'm trying to discourage adding new dependencies on them... while adding more myself.

PR #798, opened 2020-11-12
Removing api v1 code

ghop02:
👋 api v1 code.

Removes all API V1 code.  Most of it is pure removal, but moved a couple files around and moved some references we used in the API v2 definitions.


libs/datasets/sources/dh_beds.py
```diff
@@ -12,6 +11,10 @@
 
 _logger = logging.getLogger(__name__)
 
+# Fips code chosen for all unknown fips values.
+# TODO: This should maybe be unique per state.
```
TomGoBravo, line 15:
```
# Use libs.us_state_abbrev.ABBREV_US_UNKNOWN_COUNTY_FIPS when state is known.
```

TomGoBravo (approved):

PR #800, opened 2020-11-14
Increase lookback days to 15 to reduce risk of dropping CMS data.

mikelehen:


libs/top_level_metrics.py
```diff
@@ -23,8 +23,9 @@
 RT_TRUNCATION_DAYS = 7
 
 
-# CMS and HHS testing data can both lag by more than 7 days. Let's use it unless it's 2 weeks old.
-MAX_METRIC_LOOKBACK_DAYS = 13
+# CMS and HHS testing data can both lag by more than 7 days. Let's use it unless it's >2 weeks old.
+# TODO(michael): Consider having different lookback values per metric, but this is fine for now.
```
TomGoBravo, line 27: And somewhat related, calculate_or_copy_test_positivity does its own thing with has_data_in_past_10_days

TomGoBravo (approved):

PR #801, opened 2020-11-15
Disable RI county test positivity from covid-county-data, since the rates seem to be way higher than CMS.

mikelehen:
The CMS rates are way lower and inline with the state-level positivity, so I think we should use CMS.
https://trello.com/c/e9o3mhjW/585-try-switching-to-cms-test-positivity-data-for-rhode-island

Demo snapshot here (FYI- also includes reenabling OR, NE, and CA HHS data): https://covid-projections-git-mikelehen-demo-snapshot.covidactnow.vercel.app/internal/compare


mikelehen:
[Try switching to CMS test positivity data for Rhode Island.](https://trello.com/c/e9o3mhjW/585-try-switching-to-cms-test-positivity-data-for-rhode-island)

TomGoBravo (approved):

PR #803, opened 2020-11-17
Add workflow dispatch for deploy_api

ghop02:
Updates deploy_api to use workflow dispatch.


TomGoBravo (approved):
please add something about how you tested this.

ghop02:
> please add something about how you tested this.

Ha, I honestly didn't test it.  Since it's not on the master branch I don't think I can easily kick it off without a separate build file (which I would need to merge to master to test).  Planning on testing right now

PR #806, opened 2020-11-17
Better empty timeseries handling

TomGoBravo:

### Please Check
- [ ] Are tests passing?


ghop02 (approved):
Lgtm - there is a test that's failing but that I think is from the master build.

PR #807, opened 2020-11-17
Use Repo datasets in update_repo_datasets

ghop02:
Before merging this, I'd like to update the workflow in covid data public that calls this, but independently this is good for a review!


.github/workflows/update_repo_datasets.yml
```diff
@@ -1,11 +1,14 @@
...
-  # This hook is automatically triggered by
+  # This workflow is automatically triggered by
   # https://github.com/covid-projections/covid-data-public/blob/master/.github/workflows/update_data.yml
   # after covid-data-public datasets have been updated.
-  repository_dispatch:
-    types: update-dataset-snapshot
+  workflow_dispatch:
+    inputs:
+      trigger_api_build:
+        description: 'If "true" API Build will be triggered after dataset update.'
```
mikelehen, line 10: nit: It took me a minute to figure out that "API Build" meant "snapshot build".  Might be worth trying to unify our internal terminology.  As a compromise we could kinda' combine terminology and say 'If "true" a new API snapshot build will be triggered ..."  

I don't care super strongly though.

tools/trigger-update-datasets.sh
```diff
@@ -14,13 +14,16 @@ prepare () {
     echo "https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line"
     exit 1
   fi
+
+
```
mikelehen, line 18: nit: intentional newlines?

mikelehen (approved):
LGTM with optional nits.

.github/workflows/update_repo_datasets.yml
ghop02: I like API snapshot build I think, will change it!

PR #810, opened 2020-11-18
General aggregate for country and CBSA

TomGoBravo:
* Changes `data update` to always aggregate CBSAs
* Adds flag `--aggregate-to-country` to create a USA region
* Adds commands `aggregate_cbsa` and `aggregate_states_to_country` to test these aggregations
* Adds `timeseries.aggregate_regions` used by all the above to aggregate regions using a map from old to new region
* Adds `Region.from_iso1`, to create the Region object for USA
* Removes parameter `interpolate_stalled_and_missing_values(interpolate_stalled)` which is never used (unrelated to other changes).


libs/datasets/timeseries.py
```diff
@@ -952,3 +953,97 @@ def drop_regions_without_population(
...
         )
     return mrts.get_locations_subset(locations_with_population)
+
+
+LOCATION_ID_AGG = "location_id_agg"
+
+
+def _aggregate_dataframe_by_region(
+    df_in: pd.DataFrame, location_id_map: Mapping[str, str]
+) -> pd.DataFrame:
```
ghop02, line 963: The logic here overall seems correct, but I'm a bit unsure it'll provide for easy flexibility with some of the other sorts of calculations.

How would you feel about an approach that takes advantage of the apply function with a named series? I would imagine it would be something like:

```python
def sum_no_nan(series):
    return series.sum(min_count=len(series))

data.groupby(groupby_columns).apply(
    lambda x: pd.Series({
        column: sum_no_nan(x[column])
        for column in columns
    })
)
```

The idea being that it would be pretty easy to treat different columns differently and apply different aggregating functions. 

It then lessens the need to go back and forth between wide and long form columns and I think the code will be a bit more concise to digest.

TomGoBravo: A simple test found this will take ~30min to run with bare python, vs seconds without apply. It is quite possible numba can bring that down to a few seconds, but I haven't tried.  See also the description of https://github.com/covid-projections/covid-data-model/pull/813

TomGoBravo:
Replaced by https://github.com/covid-projections/covid-data-model/pull/811

PR #811, opened 2020-11-20
General aggregate for country and CBSA, and filter by count

TomGoBravo:
Similar to https://github.com/covid-projections/covid-data-model/pull/810 but I think this more correctly removes output values that are not aggregated from the correct count of input values.

* Changes `data update` to always aggregate CBSAs
* Adds flag `--aggregate-to-country` to create a USA region
* Adds commands `aggregate_cbsa` and `aggregate_states_to_country` to test these aggregations
* Adds `timeseries.aggregate_regions` used by all the above to aggregate regions using a map from old to new region
* Adds `Region.from_iso1`, to create the Region object for USA
* Removes parameter `interpolate_stalled_and_missing_values(interpolate_stalled)` which is never used (unrelated to other changes).


ghop02 (approved):
This looks good to me! I still may try and play around with a different aggregation function at some point, but fine with this as it is

PR #812, opened 2020-11-20
Remove time restrictions on label-api and pr-generation

ghop02:
This removes the time restrictions to kick of snapshot PRs on the website and label api.


mikelehen (approved):
Looks good, though I think we should move towards ditching "maybe-trigger-web-snapshot-update.sh" and instead add a "trigger_website_pr" input to the deploy_api workflow and explicitly set it to true for "official" builds.

PR #813, opened 2020-11-20
add _apply_scaling_factor and test

TomGoBravo:
This PR builds on the un-merged https://github.com/covid-projections/covid-data-model/pull/811 to add scaling of certain timeseries using certain static values.

Not hard-coding the fields may be making this code more flexible than we need, but does make writing small unit tests easier.

This development direction uses panda parallel operations pretty heavily to get the speed to aggregate all CBSA's in a few seconds on my laptop. This makes the code somewhat harder to read because each value and operation must handle multiple locations at the same time. For example, instead of multiplying by a scalar weight the code needs to multiple by a vector of weights indexed by location.

Alternative development directions that have code paths processing one location at a time are likely easier to read but harder to run. Panda's builtin aggregation functions are not flexible enough to do things like weighted averages. It'd be really nice to have a function that takes the unaggregated static + timeseries DataFrame for each aggregated region, and returns the static and timeseries of that one region. A simple test with `apply` suggests this will take ~30min to run with bare python. It is quite possible numba can bring that down to a few seconds, but I haven't tried. It'd be nice if we had an easy way to run this in the cloud (I'd like to try Dask/Prefect) but that is a much bigger task.


libs/datasets/timeseries.py
```diff
@@ -959,6 +960,67 @@ def drop_regions_without_population(
...
+)
+
+
+def _apply_scaling_factor(
+    timeseries: pd.DataFrame,
+    static_in: pd.DataFrame,
+    static_agg: pd.DataFrame,
+    location_id_map: Mapping[str, str],
+    aggregations: Sequence[AverageStaticWeightAggregation],
+) -> pd.DataFrame:
```
ghop02, line 985: I'm not sure this solves the problem for the hospitalization occupancy rates.  Those are only in static data and I don't see that being aggregated anywhere?

libs/datasets/timeseries.py
```diff
@@ -959,6 +960,67 @@ def drop_regions_without_population(
...
+    """Information about a timeseries field which is aggregated using a weighted average."""
+
+    # timeseries field that gets aggregated using a weighted average
+    field: FieldName
+    # static field that used to produce the weights
+    scale_factor: FieldName
+
+
+AGGREGATIONS = (
+    AverageStaticWeightAggregation(CommonFields.TEST_POSITIVITY, CommonFields.POPULATION),
```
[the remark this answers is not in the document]
ghop02: bump on this comment

libs/datasets/timeseries.py
```diff
@@ -959,6 +960,67 @@ def drop_regions_without_population(
...
+    AverageStaticWeightAggregation(CommonFields.CUMULATIVE_HOSPITALIZED, CommonFields.ICU_BEDS),
```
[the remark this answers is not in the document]
TomGoBravo: Addressed with b7b6603cab

libs/datasets/timeseries.py
```diff
@@ -1031,10 +1031,84 @@ def drop_regions_without_population(
 LOCATION_ID_AGG = "location_id_agg"
 
 
+@dataclass(frozen=True)
+class AverageStaticWeightAggregation:
```
ghop02, line 1035: Would `WeightedAverageAggregation` be a better name? I'm not sure what role "static" is playing in the name

libs/datasets/timeseries.py
```diff
@@ -1031,10 +1031,84 @@ def drop_regions_without_population(
...
+    """Information about a timeseries field which is aggregated using a weighted average."""
+
+    # timeseries field that gets aggregated using a weighted average
+    field: FieldName
+    # static field that used to produce the weights
+    scale_factor: FieldName
+
+
+AGGREGATIONS = (
```
ghop02, line 1044: Maybe this could be something like `WEIGHTED_AGGREGATIONS` or `CUSTOM_AGGREGATIONS`.

libs/datasets/timeseries.py
```diff
@@ -1031,10 +1031,84 @@ def drop_regions_without_population(
...
+
+    for agg in aggregations:
+        if agg.field in df_in.columns and agg.scale_factor in scale_factors.columns:
+            df_out[agg.field] = df_out[agg.field] * scale_factors[agg.scale_factor]
+
+    return df_out
+
+
+def _find_scale_factors(
+    aggregations, location_id_map, static_agg, static_in, location_ids
```
ghop02, line 1083: Could you add types here? would be helpful for context

libs/datasets/timeseries.py
```diff
@@ -1088,22 +1162,62 @@ def aggregate_regions(
     dataset_in: MultiRegionDataset,
     aggregate_map: Mapping[Region, Region],
     aggregate_level: AggregationLevel,
+    aggregations: Sequence[AverageStaticWeightAggregation] = AGGREGATIONS,
 ) -> MultiRegionDataset:
```
ghop02, line 1166: I think a docstring would be handy here

ghop02 (approved):
This seems like a good step in the right direction!

libs/datasets/timeseries.py
TomGoBravo: The weights are static, instead of time-varying.

libs/datasets/timeseries.py
TomGoBravo: Changed to DEFAULT_AGGREGATIONS. If you don't find a better way of structuring this we may end up extending the list with other types of aggregations.

libs/datasets/timeseries.py
```diff
@@ -1088,22 +1162,62 @@ def aggregate_regions(
...
+        static_in_scale_fields,
+        location_ids,
     )
-    static_out[CommonFields.AGGREGATE_LEVEL] = aggregate_level.value
 
-    return MultiRegionDataset(timeseries=timeseries_out, static=static_out)
+    static_other_fields_scaled = _apply_scaling_factor(
+        static_in_other_fields, scale_factors, aggregations
+    )
+    timeseries_scaled = _apply_scaling_factor(dataset_in.timeseries, scale_factors, aggregations)
```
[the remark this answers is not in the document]
TomGoBravo: Added TODO. We could perhaps get the same result by keeping timeseries positive and negative tests as separate metrics until after aggregation, than compute test positivity.

test/libs/datasets/timeseries_test.py
```diff
@@ -1030,3 +1031,68 @@ def test_combined_missing_field():
         {name: reversed(source_list) for name, source_list in field_source_map.items()},
     )
     assert_dataset_like(expected, combined_2)
+
+
+def test_aggregate_states_to_country_scale():
+    ts = timeseries.MultiRegionDataset.from_csv(
```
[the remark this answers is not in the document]
TomGoBravo: I'll add this as a possibility in https://trello.com/c/msqrTIB3/528-make-tabular-data-in-unittests-easier-to-read

libs/datasets/timeseries.py
TomGoBravo: Done

libs/datasets/timeseries.py

libs/datasets/timeseries.py
TomGoBravo: and then changed again to WEIGHTED_AGGREGATIONS because the default is `sum`.

ghop02 (approved):

PR #821, opened 2020-11-25
Two fixes for Unknown Aggregation Level

TomGoBravo:
Fix `NotImplementedError Unknown Aggregation Level`. See https://covidactnow.slack.com/archives/C011Z21ST8V/p1606261285011500

This partially addresses https://trello.com/c/EyISOtVY/623-country-aggregation-breaking-pyseir-output-and-doesnt-output-api-files-properly

* change data update default to --no-aggregate-to-country, disabling production of the unsupported iso1:us location_id.
* location_id_to_level support for country
* adds test that makes sure every `AggregationLevel` value is tested with location_id_to_level


TomGoBravo:
[Country aggregation breaking pyseir output and doesn't output API files properly](https://trello.com/c/EyISOtVY/623-country-aggregation-breaking-pyseir-output-and-doesnt-output-api-files-properly)

test/libs/pipeline_test.py
```diff
@@ -0,0 +1,17 @@
+from libs import pipeline
+from libs.datasets import AggregationLevel
+
+
+def test_location_id_to_level():
+    param_result_map = {
+        "iso1:us#fips:99111": AggregationLevel.COUNTY,
```
ghop02, line 7: I don't know if I realized this was a valid location ID - does the county location ID not require state?

cli/data.py
```diff
@@ -60,7 +60,7 @@ def update_forecasts(filename):
     "--aggregate-to-country/--no-aggregate-to-country",
     is_flag=True,
     help="Aggregate states to one USA country region",
-    default=True,
+    default=False,
```
ghop02, line 63: fwiw I added a line in the github actions yaml to disable county aggregations by default so not sure if this is necessary or not

ghop02 (approved):

test/libs/pipeline_test.py
TomGoBravo: I created https://trello.com/c/QEbSwjSQ/631-remove-support-for-county-locationid-without-state to fix this.

cli/data.py
TomGoBravo: Yes, I saw that after making this change :-/ I like having `python ./run.py data update` reproduce what is in master.

PR #828, opened 2020-12-02
adding can location page urls to api output

ghop02:
Adds in the location page url to API output

Here's the data public PR
https://github.com/covid-projections/covid-data-public/pull/171

This seems to do alright, but need to figure out a less hacky way for filtering out the numeric column


libs/datasets/timeseries.py
```diff
@@ -396,7 +396,10 @@ def _merge_attributes(df1: pd.DataFrame, df2: pd.DataFrame) -> pd.DataFrame:
         _log.warning(f"Re-adding empty columns: {missing_columns}")
         wide = wide.reindex(columns=[*wide.columns, *missing_columns])
     # Make all non-GEO_DATA_COLUMNS numeric so that aggregation functions work on them.
-    numeric_columns = list(all_columns - set(GEO_DATA_COLUMNS))
+    df1_numeric_columns = df1.columns[df1.dtypes.apply(is_numeric_dtype)]
```
TomGoBravo, line 399: I'm not yet a fan of this change. L 403 `apply(pd.to_numeric` is there so we notice when something non-numeric slips in to a column that is expected to be floats. This change will let such values go deeper in where they break other code in hard to debug ways. I'd like to stick with explicitly listing columns that can be non-numeric for now.

ghop02: 👍  is it too hacky to just include this column with a TODO to generalize better to non-numeric columns

TomGoBravo: hard-coding removing the new column from numeric_columns is totally fine with me. It seems less likely to cause a problem than depending on the dtypes.

ghop02: Fixes some tests too 😅

libs/datasets/timeseries.py
```diff
@@ -395,8 +396,8 @@ def _merge_attributes(df1: pd.DataFrame, df2: pd.DataFrame) -> pd.DataFrame:
     if missing_columns:
         _log.warning(f"Re-adding empty columns: {missing_columns}")
         wide = wide.reindex(columns=[*wide.columns, *missing_columns])
-    # Make all non-GEO_DATA_COLUMNS numeric so that aggregation functions work on them.
-    numeric_columns = list(all_columns - set(GEO_DATA_COLUMNS))
+    # Make all numeric columns numeric so that aggregation functions work on them.
```
TomGoBravo, line 399: "Make columns expected to be numeric have a numeric dtype so that aggregation functions work on them."

TomGoBravo (approved):

PR #831, opened 2020-12-07
drop staffed_beds, regions without population

TomGoBravo:
## This PR

* Adds call to `timeseries.drop_regions_without_population` which removes AS, GU, VI and Unknown Counties from combined data. Fixes `test_generate_infection_rate_metric_fake_fips` which was broken by the change.
* Removes `staffed_beds` from combined data. It is the only static metric that is build from a timeseries data source and removing it simplifies disentangling these
* Writes a copy of the static data next to the wide-dates csv. Unlike the latest.csv this includes regions without FIPS.
* Changes `add_new_cases` to use timeseries_wide_dates so that it produces identical results even if some dates with NA values for cases are removed
* Factors out the interesting parts of `MultiRegionDataset.from_timeseries_and_latest` so it is easier to remove.  Changes a bunch of tests in test/libs/datasets/timeseries_test.py to stop using `TimeseriesDataset` and `LatestValuesDataset` because they are about to be removed

## Tested

Ran `time ./run.py data update` and looked at `git difftool -t vimdiff data/multiregion.csv` to verify that the expected changes, and no more, took place.


libs/datasets/timeseries.py
```diff
@@ -929,34 +951,41 @@ def _remove_padded_nans(df, columns):
...
     new_cases[new_cases < 0] = pd.NA
 
-    df_copy[CommonFields.NEW_CASES] = new_cases
+    timeseries_copy = dataset_in.timeseries.copy()
 
-    new_timeseries = dataclasses.replace(dataset_in, timeseries=df_copy)
-
-    return new_timeseries
+    timeseries_copy[CommonFields.NEW_CASES] = new_cases
+    dataset_out = dataclasses.replace(dataset_in, timeseries=timeseries_copy)
```
ghop02, line 987: A function along the lines of `adding_timeseries_column(column_name, series)` that produces a new dataset object on the timeseries could be useful

ghop02: I always feel a bit dirty when doing all of this dataset copying logic

ghop02 (approved):
LGTM

libs/datasets/timeseries.py
TomGoBravo: Because of the cost of shuffling so much stuff around RAM or dataclasses.replace fiddling with attributes inside the class?

ghop02: oh, mostly the logic around fiddling with the attributes - requires a bit of insider knowledge to know that's the surgery necessary

PR #832, opened 2020-12-07
Add extreme

ghop02:

### Please Check
- [ ] Are tests passing?


libs/top_level_metric_risk_levels.py
```diff
@@ -18,8 +18,12 @@ def calc_risk_level(value: Optional[float], thresholds: List[float]) -> RiskLeve
...
         A RiskLevel.
     """
-    assert len(thresholds) == 3, "Must pass low, med and high thresholds."
-    level_low, level_med, level_high = thresholds
+    assert len(thresholds) in [3, 4], "Must pass low, med and high thresholds."
+    if len(thresholds) == 3:
+        level_low, level_med, level_high = thresholds
+        level_extreme = math.inf
+    else:
+        level_low, level_med, level_high, level_extreme = thresholds
```
TomGoBravo, line 26: Why is the last threshold not level_critical?

ghop02: that risk level is actually titled "critical" in the enum, so would be a misleading name

TomGoBravo (approved):

PR #835, opened 2020-12-07
Add back ICU calculation for CBSAs now that typical icu usage properly calculated

ghop02:
Now that the underlying typical icu utilization aggregation problem was fixed, we can start calculating the icu metric again


ghop02:
closing this as primary ICU metric has changed

PR #841, opened 2020-12-09
rm LatestValuesDataset and more

TomGoBravo:
## This PR

* removes classes LatestValuesDataset and TimeseriesDataset, moving the only used constants to libs/datasets/dataset_utils.py
* removes some hospitalization stuff that seems unused 
* removes some TimeseriesDataset properties that was only used by the hospitalization stuff and to_csv.

There is plenty more that can be cleaned up and deleted but this is enough for one PR


libs/datasets/dataset_utils.py
```diff
@@ -20,6 +20,21 @@
 
 NON_NUMERIC_COLUMNS = GEO_DATA_COLUMNS + [CommonFields.CAN_LOCATION_PAGE_URL]
 
+LATEST_VALUES_INDEX_FIELDS = [
```
ghop02, line 23: `STATIC_INDEX_FIELDS`?

ghop02 (approved):
This looks good to me! we should probably delete the files that aren't being written to disk anymore

PR #843, opened 2020-12-09
add pyseir build-all --generate-api-v2

TomGoBravo:
## This PR

* Adds `pyseir build-all --generate-api-v2` so making API output is one command after updating combined datasets. This makes testing the API output of newly aggregated and dis-aggregated regions easier.
* Adds `pyseir.run.PyseirOutputDatasets` which wraps reading and writing pyseir output
* Factors out most of click.command generate_api_v2 into generate_api_v2_from_loaded_data which gets data as args instead of from the file system
* Changes pyseir build-all --states-only to --level=state. This adds support for --level=country
* Renames `pyseir.cli.RegionPipeline` to `pyseir.run.OneRegionPipeline`
* Removes `RegionPipelineInput` which wasn't doing much after the fips/states/level filter params were replaced with a combined MultiRegionDataset.


cli/api.py
```diff
@@ -11,18 +11,20 @@
...
-from libs.pipelines import api_v2_pipeline
 from libs.datasets import combined_datasets
-from libs.datasets.timeseries import MultiRegionDataset
 from libs.datasets.dataset_utils import REPO_ROOT
 from libs.datasets.dataset_utils import AggregationLevel
-from pyseir.utils import SummaryArtifact
+import pyseir.utils
 import pandas as pd
 
+from libs.pipelines.api_v2_pipeline import generate_api_v2_from_loaded_data
```
ghop02, line 26: `from libs.pipelines import api_v2_pipeline`?

libs/pipelines/api_v2_pipeline.py
```diff
@@ -240,3 +243,34 @@ def deploy_csv_api_output(
...
     rows = dataset_deployer.remove_root_wrapper(api_output.dict())
     dataset_deployer.write_nested_csv(rows, output_path, keys_to_skip=keys_to_skip)
+
+
+def generate_api_v2_from_loaded_data(
+    model_output: pyseir.run.PyseirOutputDatasets,
+    output: pathlib.Path,
+    selected_dataset: MultiRegionDataset,
+    log,
+):
```
ghop02, line 253: docstring would be great

pyseir/run.py
```diff
@@ -0,0 +1,103 @@
...
+            region=input.region, infer_df=infer_df, icu_data=icu_data, _combined_data=input,
+        )
+
+    def population(self) -> float:
+        return self._combined_data.latest[CommonFields.POPULATION]
+
+
+@final
+@dataclasses.dataclass(frozen=True)
+class PyseirOutputDatasets:
```
ghop02, line 66: Is this different from the other `PyseirOutputDatasets`?

pyseir/cli.py
```diff
@@ -229,16 +127,84 @@ def run_infer_rt(state, states_only):
...
+    help="Directory to deploy " "webui output.",
+)
+@click.option(
+    "--generate-api-v2",
+    default=False,
+    is_flag=True,
+    type=bool,
+    help="Generate API v2 output after PySEIR finishes",
+)
+def build_all(states, output_dir, level, fips, generate_api_v2: bool):
```
ghop02, line 144: I would love to eventually move the pyseir code into `libs/`

ghop02 (approved):

pyseir/run.py
TomGoBravo: Yes! This one is intentional. The other is a goof I forgot to delete before asking you to review.

libs/pipelines/api_v2_pipeline.py
TomGoBravo: added in edce253c

cli/api.py
TomGoBravo: done 8cde3662

pyseir/run.py
TomGoBravo: fixed e726b6e

pyseir/cli.py
TomGoBravo: We have a lot of desired improvements for such a young small code base :-/

PR #846, opened 2020-12-11
Fix additional test positivity columns and NYC CBSA aggregation

ghop02:
This fixes a couple issues with the aggregator:
1. I updated the column used for test positivity. It was simply summing the new test positivity columns, not good.
2. NYC CBSA was missing.  This is because the boroughs are aggregated, so the missing counties made all columns be null. this is very much a short lived fix, but in the email mentioning the NYC switch, we refer to the new CBSA, so I'd like this to be included before sending out and not have everyone query a missing CBSA.

### DIFF
--- a/pyseir/run.py
+++ b/pyseir/run.py
@@ -10,7 +10,6 @@ from covidactnow.datapublic.common_fields import CommonFields
 from typing_extensions import final
 
 from libs import pipeline
-from libs.datasets import AggregationLevel
 from libs.datasets.timeseries import MultiRegionDataset
 from libs.datasets.timeseries import OneRegionTimeseriesDataset
 from pyseir.icu import infer_icu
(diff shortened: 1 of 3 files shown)


pyseir/run.py
```diff
@@ -43,15 +43,13 @@ def run(input: OneRegionTimeseriesDataset) -> "OneRegionPipeline":
...
-            try:
-                icu_data = infer_icu.get_icu_timeseries_from_regional_input(
-                    icu_input, weight_by=infer_icu.ICUWeightsPath.ONE_MONTH_TRAILING_CASES
-                )
-            except KeyError:
-                _log.exception(f"Failed to run icu data for {input.region}")
+        icu_input = infer_icu.RegionalInput.from_regional_data(input)
+        try:
+            icu_data = infer_icu.get_icu_timeseries_from_regional_input(
+                icu_input, weight_by=infer_icu.ICUWeightsPath.ONE_MONTH_TRAILING_CASES
```
TomGoBravo, line 49: Are you sure this works? I don't see CBSAs in data/one_month_trailing_weights_via_fips.json and don't know how they'd get aggregated before this call.

ghop02: yeah that's a good point.  Will leave that out

TomGoBravo (approved):

TomGoBravo:
[CBSA Test Positivity aggregation not scaled by county](https://trello.com/c/TeZccHQf/670-cbsa-test-positivity-aggregation-not-scaled-by-county)

PR #847, opened 2020-12-11
Use HHS hospital data imported from CCD.

mikelehen:


test/combined_dataset_test.py
```diff
@@ -63,7 +63,7 @@ def test_combined_county_has_some_timeseries_data(fips):
     region = Region.from_fips(fips)
     latest = combined_datasets.load_us_timeseries_dataset().get_one_region(region)
     df = latest.data.set_index(CommonFields.DATE)
-    date = "2020-09-01"  # Arbitrary date that both FIPS have data for.
+    date = "2020-09-04"  # Arbitrary date that both FIPS have data for.
```
mikelehen, line 66: This was to fix this test for the HHS county data (had to pick the day within the week that they collected data). Since we're punting on county data it's not needed, but it's non-harmful and will save re-making the change later.

data/multiregion.json
```diff
@@ -2,14 +2,14 @@
   "dataset_type": "multiregion",
   "path": "data/multiregion.csv",
   "data_git_info": {
-    "sha": "623fa8940f475d3eb6dc0b3c4c9cdb6c7f4979a9",
-    "branch": "master",
+    "sha": "d6c4bfd5f098ddbbf63cf0a15fd22fbbe9114eb7",
+    "branch": "mikelehen/drop-county-hhs-data",
```
mikelehen, line 6: I won't merge all of these data files.  I'll just merge the code and then run the github action to update combined datasets.

PR #848, opened 2020-12-11
Add HHSHospital dataset and include in combined datasets.

mikelehen:


test/combined_dataset_test.py
```diff
@@ -63,7 +63,7 @@ def test_combined_county_has_some_timeseries_data(fips):
     region = Region.from_fips(fips)
     latest = combined_datasets.load_us_timeseries_dataset().get_one_region(region)
     df = latest.data.set_index(CommonFields.DATE)
-    date = "2020-09-01"  # Arbitrary date that both FIPS have data for.
+    date = "2020-09-04"  # Arbitrary date that both FIPS have data for.
```
mikelehen, line 66: This was to hit a date that had data when we were pulling in the weekly county data.  It's no longer necessary, but it's non-harmful and means it won't break again when we try to add county data again.

BrettBoval (approved):
👍

PR #851, opened 2020-12-15
HHS latest ICU_BEDS in static

TomGoBravo:
## This PR

* Adds HHSHospitalDataset as highest priority source for static ICU_BEDS
* Factors out MultiRegionDataset.latest_in_static from CovidCountyDataDataSource and uses it in HHSHospitalDataset

## Tested

Ran `./run.py data update`. Saw that `multiregion-wide-dates.csv` didn't change (no change in timeseries) and only change in `multiregion-static.csv` was static data changed for states, notably 674 changed to 1129 in us-ct and other columns didn't change.


ghop02 (approved):

libs/datasets/timeseries.py
```diff
@@ -316,6 +316,16 @@ def _timeseries_latest_values(self) -> pd.DataFrame:
...
         return long.loc[unduplicated_and_last_mask, :].unstack()
 
+    def latest_in_static(self, field: FieldName) -> "MultiRegionDataset":
+        """Returns a new object with the latest values from timeseries 'field' copied to static."""
+        latest_series = self._timeseries_latest_values()[field]
+        if field in self.static.columns and self.static[field].notna().any():
+            # This is not expected, so detect it and don't continue
+            raise ValueError("Can only copy field when static is unset")
+        static_copy = self.static.drop(columns=[field], errors="ignore")
+        static_copy.insert(len(static_copy.columns), field, latest_series)
```
mikelehen, line 326: Is this equivalent to:
```
static_copy[field] = latest_series
```
?

Is there a reason to prefer insert()?  Seems undesirable to have to specify the column position since I assume we don't care?

test/libs/datasets/timeseries_test.py
```diff
@@ -651,6 +651,31 @@ def test_timeseries_latest_values():
...
+
+    # Check access to latest values as copied to static
+    t1 = FieldName("t1")
+    s1 = FieldName("s1")
+    dataset_t1_latest_in_static = dataset.latest_in_static(t1)
+    assert dataset_t1_latest_in_static.static.loc["iso1:us#cbsa:10100", t1] == 10
+    assert dataset_t1_latest_in_static.static.loc["iso1:us#fips:97111", t1] == 4
+
+    with pytest.raises(ValueError):
+        dataset.latest_in_static(s1)
```
mikelehen, line 676: It's not immediately obvious to me why this is supposed to raise an exception.  Is it because some regions don't have a latest value (while others do)?  Can you add a comment?

mikelehen (approved):
Couple nit / questions, but LGTM.

libs/datasets/timeseries.py
TomGoBravo: Nice catch. I started by only calling `drop` when the column was present but removed a branch and didn't simplify the field creation.

test/libs/datasets/timeseries_test.py
TomGoBravo: Done in 75957449

PR #855, opened 2020-12-16
Drop stalled timeseries tails

TomGoBravo:
## This PR

This PR addresses https://trello.com/c/UKy1yEtd/664-hhs-test-positivity-handle-bad-data-in-recent-days-better and https://trello.com/c/W96ocnI5/663-daily-new-cases-handle-anomalous-trailing-zeros-better and is a first basic attempt at https://trello.com/c/Cq2zNk2M/674-annotate-data-outliers-that-weve-flagged-in-the-ui

* Adds `class TailFilter` which looks for and removes suspiciously low recent values of a cumulative timeseries.
* Adds `AnnotationField` and `AnnotationType` as a way to tag a particular location, variable, date as a place where the data was modified.
* Uses the above to filter the cumulative timeseries put in combined datasets
* Writes the annotations to a new CSV during `data update`
* Adds a new command `run-bad-tails-filter` for testing the filter.

### For [HHS Test Positivity: Handle bad data in recent days better](https://trello.com/c/UKy1yEtd/664-hhs-test-positivity-handle-bad-data-in-recent-days-better)

Checked out the data from ~Dec 8 where TN and VA were broken, than used `run-bad-tails-filter` to check what the filter would have done. It looks like the stalled values would have been removed. The other changes I saw when skimming through the diff looked reasonable.
```
git checkout 04885f796deebab7cc6ed1e607a080d4483bff6a -- data/multiregion.csv data/multiregion-wide-dates.csv data/multiregion-static.csv data/multiregion-provenance.csv
./run.py data run-bad-tails-filter data/multiregion-wide-dates-bad-tails-filter.csv 
vimdiff data/multiregion-wide-dates.csv data/multiregion-wide-dates-bad-tails-filter.csv
```

### For [Daily New Cases: Handle anomalous trailing zeros better](https://trello.com/c/W96ocnI5/663-daily-new-cases-handle-anomalous-trailing-zeros-better)

Tested by checking out the data from snapshot 1413 with `git checkout 8beea39fc37ffb246f6e9f3299ba54071118685f -- data/` then running `./run.py data run-bad-tails-filter data/multiregion-wide-dates-bad-tails-filter.csv` and comparing with `vimdiff data/multiregion-wide-dates.csv data/multiregion-wide-dates-bad-tails-filter.csv` which shows that lots of bad tails were removed, including one value from NM cases. This is confirmed by looking at `grep us-nm,cases data/multiregion-wide-dates-bad-tails-filter-annotations.csv` which outputs `iso1:us#iso2:us-nm,cases,cumulative_tail_truncated,Removed 1 values less than 20.,2020-12-07`


TomGoBravo:
[HHS Test Positivity: Handle bad data in recent days better.](https://trello.com/c/UKy1yEtd/664-hhs-test-positivity-handle-bad-data-in-recent-days-better)

libs/datasets/timeseries.py
```diff
@@ -1224,3 +1228,99 @@ def aggregate_puerto_rico_from_counties(dataset: MultiRegionDataset) -> MultiReg
...
+        return tail_filter, dataclasses.replace(dataset, timeseries=timeseries_wide_variables)
+
+    def annotations_as_dataframe(self):
+        return pd.DataFrame(self.annotations)
+
+    def _filter_one_series(self, series_in: pd.Series) -> pd.Series:
+        """Filters one timeseries of cumulative values. This is a method so self can be used to
+        store side outputs."""
+
+        if len(series_in) < 28:
```
ghop02, line 1285: Can you either pass these in as kwargs or set them as constants?

test/libs/datasets/timeseries_test.py
```diff
@@ -1186,7 +1348,7 @@ def test_multi_region_dataset_get_subset():
             "iso1:us#iso2:us-tx,state,TX,,2020-04-01,4,2,\n"
             "iso1:us#iso2:us-tx,state,TX,,,,,5000\n"
             "iso1:us#fips:97222,county,,97222,2020-04-01,1,2,\n"
-            "iso1:us#fips:97222,county,,97222,,,,1000\n"
+            "iso-2:us#fips:97222,county,,97222,,,,1000\n"
```
ghop02, line 1351: What is this change for?

cli/data.py
```diff
@@ -30,6 +30,18 @@
...
+    CommonFields.CASES,
+    CommonFields.DEATHS,
+    CommonFields.POSITIVE_TESTS,
+    CommonFields.NEGATIVE_TESTS,
+    CommonFields.TOTAL_TESTS,
+    CommonFields.POSITIVE_TESTS_VIRAL,
+    CommonFields.POSITIVE_CASES_VIRAL,
+    CommonFields.TOTAL_TESTS_VIRAL,
+    CommonFields.TOTAL_TESTS_PEOPLE_VIRAL,
+    CommonFields.TOTAL_TEST_ENCOUNTERS_VIRAL,
```
ghop02, line 43: I realize this isn't a cumulative field, but do we want to filter on TEST_POSITIVITY_7D and TEST_POSITIVITY_14D?

libs/datasets/timeseries.py
```diff
@@ -1224,3 +1228,99 @@ def aggregate_puerto_rico_from_counties(dataset: MultiRegionDataset) -> MultiReg
...
+            return series_in
+        # Pick a bound used to determine if the diff of recent values are reasonable. For
+        # now values less than 100th of the mean are considered stalled. The variance could be
+        # used to create a tighter bound but the simple threshold seems to work for now.
+        # TODO(tom): Experiment with other ways to calculate the bound.
+        threshold = math.floor(mean / 100)
+        # Go backwards in time, starting with the most recent observation. Stop as soon as there
+        # is an observation with diff that looks reasonable / not stalled / over the threshold.
+        count_observation_diff_under_threshold = 0
+        for i in range(1, 15):
```
ghop02, line 1302: is there a large benefit to going back this far? this seems like a long way back to be checking.  But also I guess by this threshold, any region with a case count avg of < 100 won't ever apply any checks which seems reasonable.

libs/datasets/timeseries.py
```diff
@@ -1224,3 +1228,99 @@ def aggregate_puerto_rico_from_counties(dataset: MultiRegionDataset) -> MultiReg
...
+    ) -> Tuple["TailFilter", MultiRegionDataset]:
+        """Returns a dataset with recent data that looks bad removed from fields."""
+        timeseries_wide_dates = dataset.timeseries_wide_dates()
+
+        fields_mask = timeseries_wide_dates.index.get_level_values(PdFields.VARIABLE).isin(fields)
+        to_filter = timeseries_wide_dates.loc[pd.IndexSlice[:, fields_mask], :]
+        not_filtered = timeseries_wide_dates.loc[pd.IndexSlice[:, ~fields_mask], :]
+
+        tail_filter = TailFilter()
+        filtered = to_filter.apply(tail_filter._filter_one_series, axis=1)
```
ghop02, line 1271: nit but since you're calling this from the object, I'd prefer for it to be public

ghop02 (commented):
This looks good to me! I think before giving it a final ✅ I'd like to see a compare, just to see the effects of it.  Also 🎉 for annotations!

libs/datasets/timeseries.py
TomGoBravo: made them class constants

test/libs/datasets/timeseries_test.py
TomGoBravo: To make sure you are reviewing all the code. :-P undone by f091d94c

cli/data.py
TomGoBravo: Sure, but as you say, this filter is for cumulative metrics, not ratios. Are you thinking of something like 'guardrails' that trigger when it is outside a sensible range (0.1% - 60%) ?

libs/datasets/timeseries.py
TomGoBravo: I think one of the stalls that motivated this change was ~9 days long.

libs/datasets/timeseries.py
TomGoBravo: I'm thinking the method is internal to code within the class, not part of the API, so a sunder is appropriate. `run` is in the class  so accessing things not in the API seems reasonable.

cli/data.py
ghop02: Yeah and actually we already have the filter that removes trailing zeros in covid-data-public so I don't think it's necessary here

ghop02:
One small point that we can 100% change later is it might be helpful to separate the annotations out to 1 per day - thinking about how we would display this data on a chart and daily seems more reasonable.  But I don't think we should change the current implementation, something to think about for the future

TomGoBravo:
[Daily New Cases: Handle anomalous trailing zeros better.](https://trello.com/c/W96ocnI5/663-daily-new-cases-handle-anomalous-trailing-zeros-better)

ghop02 (approved):

PR #856, opened 2020-12-17
More lenient daily new cases

ghop02:
this makes a couple changes to new cases calculation:

 - Before we were setting all values < 0 to nan.  Looking at the data, a lot of counties have days of -1, but setting the -1s to zero as opposed to nan, we have a whole lot of nans.  I think this is an alright change because it's a very small adjustment and it makes the timeseries much more useable for Rt.
 - Sets all days before the first day to zero.  probably doesn't make that big of a difference but when the first reported case was on different days, wouldn't calculate anything until all counties have at least one case. 

You can see that most msa's now have rt with these two changes: https://covid-projections-git-ghop02-snapshot-with-more-lenient-zeros.covidactnow.now.sh/


libs/datasets/timeseries.py
```diff
@@ -740,6 +740,8 @@ def _diff_preserving_first_value(series):
     first_date = cases.notna().idxmax()
     if pd.notna(first_date):
         new_cases[first_date] = cases[first_date]
+        new_cases[:first_date] = 0
```
TomGoBravo, line 743: I'm surprised setting all days before the first day to zero doesn't cause more of a problem; I thought many cases timeseries start midway through the pandemic and the jump from 0 to that midpoint would cause a wacky Rt around that date. Could you check on the location with largest cases[first_date] that is modified by this change?

TomGoBravo (approved):

libs/datasets/timeseries.py
ghop02: It looks like this affects the counties in PR the most.  most others (with one or two exceptions) were < 10 on their first day.  I might change this back - it shouldn't affect rt calcs

mikelehen: IIRC pandas slices are inclusive and so this will overwrite first_date?

mikelehen: n/m. Just saw you removed this. 😁

TomGoBravo: Oh good call about loc label indexing being inclusive at the end. I was curious about a fix... something like
```
first_date_ix = cases.notna().argmax()
if pd.notna(cases.iat[first_date_ix]:
    new_cases.iat[first_date_ix] = cases.iat[first_date_ix]
    new_cases.iloc[:first_date_ix] = 0
```

Moot point now.

PR #859, opened 2020-12-17
Moving files related to metrics to metrics/ and rename references

ghop02:
Now that we have a few more metrics that we're calculating, i wanted to move all of the related "metrics" files in libs into it's own directory.


TomGoBravo:
Can you create a similar directory in `test/`? How about renaming `icu_headroom_metric.py` to `icu_headroom.py`?

TomGoBravo (approved):
test/libs/metrics/icu_headroom_test.py (rm metric in filename)

PR #860, opened 2020-12-18
Add icu capacity

ghop02:
Adds ICU capacity to the metrics + risk level (though it's not included in the overall risk level) - it follows the existing pattern of computing metrics.  There is definitely some work that can be done around refactoring in the future, but don't think we need it yet.  

I did refactor the tests a bit - hopefully will make further tests/debugging a bit easier


libs/metrics/icu_capacity.py
```diff
@@ -0,0 +1,24 @@
...
+
+def calculate_icu_capacity(region_df: pd.DataFrame):
+    """Calculate ICU Capacity ratio.
+
+    Args:
+        region_df: Data for a specific region.
+
+    Returns: np.nan if data missing or pd.Series of ICU capacity.
+    """
+    # TODO(chris): Why does the common fields break if it's not in the array?
```
mikelehen, line 15: Can you resolve this TODO or else expand on it?  I can't quite tell what it's talking about or what action we'd want to take.

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -101,24 +151,34 @@ def test_top_level_metrics_basic():
...
     )
+    metrics = [
+        "caseDensity",
+        "testPositivityRatio",
+        "contactTracerCapacityRatio",
+        "infectionRate",
+        "infectionRateCI90",
+        "icuHeadroomRatio",
+        "icuCapacityRatio",
+    ]
```
mikelehen, line 172: unused?

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -1,40 +1,90 @@
...
 from libs.metrics import top_level_metrics
 from libs.datasets.timeseries import MultiRegionDataset
 from libs.datasets.timeseries import OneRegionTimeseriesDataset
 from libs.pipeline import Region
-from freezegun import freeze_time
 
+from test.libs.datasets import timeseries_test
 from test.dataset_utils_test import read_csv_and_index_fips_date
 
+INPUT_COLUMNS = [
```
mikelehen, line 18: This is used by one test and passed directly into _build_timeseries_dataframe().  Would it make sense to just move this inside of _build_timeseries_dataframe()?  Else, can you add some comments / TODOs to make it clearer how you expect this to be evolved in the future?

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -1,40 +1,90 @@
...
+
+def _build_timeseries_dataframe(fips, all_columns, start_date=None, dates=None, **column_data):
+
+    data = {column: column_data.get(column, np.nan) for column in all_columns}
 
-def _build_metrics_df(content: str) -> pd.DataFrame:
-    header = (
-        "date,fips,caseDensity,testPositivityRatio,contactTracerCapacityRatio,"
-        "infectionRate,infectionRateCI90,icuHeadroomRatio\n"
+    # TODO: Clean this up
```
mikelehen, line 35: Can you resolve this or make it more specific?

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -1,40 +1,90 @@
...
+    "positive_tests",
+    "negative_tests",
+    "contact_tracers_count",
+    "current_icu",
+    "icu_beds",
+    "current_icu_total",
+]
+
+
+def _build_timeseries_dataframe(fips, all_columns, start_date=None, dates=None, **column_data):
```
mikelehen, line 31: Can you add a minimal comment about this helper?  And maybe a TODO to convert the rest of the tests to use it for their input?

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -1,40 +1,90 @@
...
+        dates = [pd.Timestamp(date) for date in dates]
+    elif start_date:
+        dates = pd.date_range(start_date, periods=max_len, freq="D")
+    else:
+        raise ValueError("Must set either `start_date` or `dates`")
+
+    return pd.DataFrame({"date": dates, "fips": fips, **data})
+
+
+def _build_metrics_df(fips, start_date=None, dates=None, **metrics_data) -> pd.DataFrame:
```
mikelehen, line 48: Another minimal comment about this helper?

mikelehen (approved):
Nice!  I like the direction this is going with the testing. 😁

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -1,40 +1,112 @@
...
+        return dataclasses.replace(dataset, latest=latest)
+    else:
+        return dataset
+
+
+def _fips_df_to_one_region(df, region: Region, latest=None) -> OneRegionTimeseriesDataset:
+    # from_timeseries_and_latest adds the location_id column needed by get_one_region
+    dataset = MultiRegionDataset.from_fips_timeseries_df(df).get_one_region(region)
+    if latest:
+        return dataclasses.replace(dataset, latest=latest)
```
TomGoBravo, line 107: I'm fairly wary of using replace with OneRegionTimeseriesDataset.latest. Unlike MultiRegionDataset it doesn't (yet) have `__post_init__` checks that the data has the intended structure. Some subset of the latest values are derived from the timeseries data in OneRegionTimeseriesDataset.data so _fips_df_to_one_region could quite easily return an object that won't exist outside of a test.
Long term I think it is worth giving OneRegionTimeseriesDataset the same treatment MultiRegionDataset got where static and timeseries are separated and `latest` is derived from them dynamically.
I think tests will be easier to read and maintain if the parameter used to tweak `latest` only modified the dict that is already in the OneRegionTimeseriesDataset.

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -101,24 +173,25 @@ def test_top_level_metrics_basic():
...
     results, _ = top_level_metrics.calculate_metrics_for_timeseries(
         one_region, None, None, structlog.get_logger(), require_recent_icu_data=False
     )
 
     expected = _build_metrics_df(
-        "2020-08-17,36,10,,0.02,,,0.5\n"
-        "2020-08-18,36,10,0.1,0.04,,,0.5\n"
-        "2020-08-19,36,,0.1,,,,0.5\n"
-        "2020-08-20,36,,0.1,,,,0.5\n"
+        "36",
```
TomGoBravo, line 189: It'd be nice to replace FIPS with location_id so we can more easily test MSAs etc. But maybe that is too much PR feature creep.

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -101,24 +173,25 @@ def test_top_level_metrics_basic():
...
+        start_date="2020-08-17",
```
TomGoBravo, line 190: I'm not sure this treatment makes sense for all existing literal CSVs in test code but it sure makes these easier to read.

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -1,40 +1,112 @@
...
 
 
-def _fips_csv_to_one_region(csv_str: str, region: Region) -> OneRegionTimeseriesDataset:
+def _fips_csv_to_one_region(
+    csv_str: str, region: Region, latest=None
+) -> OneRegionTimeseriesDataset:
     df = read_csv_and_index_fips_date(csv_str).reset_index()
     # from_timeseries_and_latest adds the location_id column needed by get_one_region
-    return MultiRegionDataset.from_fips_timeseries_df(df).get_one_region(region)
+    dataset = MultiRegionDataset.from_fips_timeseries_df(df).get_one_region(region)
```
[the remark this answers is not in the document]
ghop02: mostly because I just moved around existing code :)

libs/metrics/top_level_metrics.py
```diff
@@ -89,6 +91,8 @@ def calculate_metrics_for_timeseries(
     )
     icu_metric, icu_metric_details = icu_headroom.calculate_icu_utilization_metric(icu_data)
 
+    icu_capacity_ratio = icu_capacity.calculate_icu_capacity(data)
```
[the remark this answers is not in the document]
ghop02: yeah - I agree but this copies existing code patterns and I'd rather for this just keep it consistent and later refactor this function to consistently use one region (or maybe figure out how to do multi-region)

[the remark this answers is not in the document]
TomGoBravo: Since we know what direction we'd like this to go why not start doing it for all new code, basically slipping in the refactor cleanup as we go.

PR #862, opened 2020-12-19
Add filtering to current timeseries

ghop02:
This repurposes this existing cumulative tail series filter to filter both cumulative and current values.  I'm not sure I love calling it this way, but was the quickest way to repurpose the existing code to get both cumulative and current tail filters.  

I mostly went down this route because TN hospitalizations were cut off, this filter does end up removing those tail values.  Happy to take another approach or just disable TN for now.

### DIFF
--- a/cli/data.py
+++ b/cli/data.py
@@ -8,6 +8,7 @@ import shutil
 import structlog
 
 import click
+import pandas as pd
 from covidactnow.datapublic import common_df
 from covidactnow.datapublic.common_fields import CommonFields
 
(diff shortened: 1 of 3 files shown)


TomGoBravo (approved):
This isn't horrible. Way better than using it to filter positivity ratio :-P That said metrics we fetch as non-cumulative (here net daily incidence of cumulative_icu_admit - cumulative_icu_discharge) likely have much lower daily incidence than test results so my crude floor(1/100) threshold could likely do with more tuning to better separate good and bad values.

Please factor out the 2 * (two timeseries.TailFilter.run calls) so `update` stays in sync with `run_bad_tails_filter`.

Nit: Text of AnnotationField.COMMENT is not accurate for cumulative=False; no longer mean of diff.

PR #863, opened 2020-12-21
calculate_test_positivity takes Dataset

TomGoBravo:
## This PR

This PR is part of https://trello.com/c/6ZPv7k9T/477-clean-up-redundant-test-positivity-calculation-code-api-vs-combined-dataset

* Changes `calculate_test_positivity` to run the smoothing itself and take a `OneRegionTimeseriesDataset` instead of raw DataFrame.
* Adds test helpers build_dataset and build_one_region_dataset to simplify making test data

## Tested

Ran `python ./run.py api generate-test-positivity` in master and this branch. Compared results/output/test-positivity-all.csv and results/output/test-positivity-output.csv from each. They were identical.

### DIFF
--- a/libs/metrics/top_level_metrics.py
+++ b/libs/metrics/top_level_metrics.py
@@ -196,17 +196,9 @@ def calculate_case_density(
 
 
 def calculate_test_positivity(
-    region_dataset: OneRegionTimeseriesDataset, smooth: int = 7, lag_lookback: int = 7
+    region_dataset: OneRegionTimeseriesDataset, lag_lookback: int = 7
 ) -> pd.Series:
-    """Calculates positive test rate.
-
-    Args:
-        positive_tests: Number of cumulative positive tests.
-        negative_tests: Number of cumulative negative tests.
-
-    Returns:
-        Positive test rate.
-    """
+    """Calculates positive test rate from combined data."""
     data = region_dataset.date_indexed
     positive_tests = series_utils.interpolate_stalled_and_missing_values(
         data[CommonFields.POSITIVE_TESTS]
--- a/libs/metrics/top_level_metrics.py
+++ b/libs/metrics/top_level_metrics.py
@@ -133,11 +133,11 @@ def _lookup_test_positivity_method(
 
 
 def calculate_or_copy_test_positivity(
-    ts: OneRegionTimeseriesDataset, log,
+    dataset_in: OneRegionTimeseriesDataset, log,
 ) -> Tuple[pd.Series, TestPositivityRatioDetails]:
-    # TODO(tom): Move all these calculations to test_positivity.AllMethods or something applied to each
-    # datasource with test metrics.
-    data = ts.date_indexed
+    # TODO(tom): Move all these calculations to test_positivity.AllMethods or something applied to
+    #  each datasource with test metrics.
+    data = dataset_in.date_indexed
     # Use POSITIVE_TESTS and NEGATIVE_TEST if they are recent or TEST_POSITIVITY is not available
     # for this region.
     positive_negative_recent = has_data_in_past_10_days(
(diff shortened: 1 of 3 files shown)


libs/metrics/top_level_metrics.py
```diff
@@ -145,20 +145,12 @@ def calculate_or_copy_test_positivity(
...
-        )
         method = _lookup_test_positivity_method(
             ts.provenance.get(CommonFields.POSITIVE_TESTS),
             ts.provenance.get(CommonFields.NEGATIVE_TESTS),
             log,
         )
-        test_positivity = calculate_test_positivity(
-            cumulative_positive_tests, cumulative_negative_tests
-        )
+        test_positivity = calculate_test_positivity(ts)
```
ghop02, line 153: small nit: don't love `ts` abbreviation, doesn't say much about the incoming data

libs/metrics/top_level_metrics.py
```diff
@@ -204,17 +196,17 @@ def calculate_case_density(
...
 def calculate_test_positivity(
-    positive_tests: pd.Series, negative_tests: pd.Series, smooth: int = 7, lag_lookback: int = 7
+    region_dataset: OneRegionTimeseriesDataset, lag_lookback: int = 7
 ) -> pd.Series:
-    """Calculates positive test rate.
-
-    Args:
-        positive_tests: Number of cumulative positive tests.
-        negative_tests: Number of cumulative negative tests.
+    """Calculates positive test rate from combined data."""
```
ghop02, line 201: Add arguments? (i realize we didn't have lag_lookback in the last docstring either)

test/libs/datasets/timeseries_test.py
```diff
@@ -29,6 +29,12 @@
 pytestmark = pytest.mark.filterwarnings("error", "ignore::libs.pipeline.BadFipsWarning")
 
 
+DEFAULT_FIPS = "97222"
```
ghop02, line 32: Super small, but I prefer naming this the name of the region this represents.  Not sure why i would expect this fips to be the default (or even where it is)

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -31,6 +39,42 @@
...
 
 
+def build_dataset(
+    metrics: Mapping[Region, Mapping[FieldName, Sequence[float]]], start_date="2020-04-01",
+) -> timeseries.MultiRegionDataset:
+    """Returns a dataset for multiple regions and metrics. Each sequence of values represents a
+    timeseries metric with identical length."""
+    # From https://stackoverflow.com/a/47416248. Make a dictionary listing all the timeseries
+    # sequences in metrics.
+    d = {
```
ghop02, line 49: can you give this a better name?

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -31,6 +39,42 @@
 ]
 
 
+def build_dataset(
```
ghop02, line 42: This looks like it largely duplicates what's in `_build_timeseries_dataframe`.  Maybe worth adding a comment why this is different and direction on how to consolidate the two in the future?

test/libs/metrics/top_level_metrics_test.py
```diff
@@ -31,6 +39,42 @@
...
+
+
+def build_one_region_dataset(
+    metrics: Mapping[FieldName, Sequence[float]],
+    region: Region = DEFAULT_REGION,
+    start_date="2020-08-25",
+) -> timeseries.OneRegionTimeseriesDataset:
+    """Returns a dataset for one region with given timeseries metrics, each having the same
+    length."""
+    return build_dataset({region: metrics}, start_date).get_one_region(region)
```
ghop02, line 75: Can you call `start_date=start_date` since it's a keyword argument

libs/metrics/top_level_metrics.py
```diff
@@ -145,20 +145,12 @@ def calculate_or_copy_test_positivity(
     ) and has_data_in_past_10_days(data[CommonFields.NEGATIVE_TESTS])
     test_positivity = common_df.get_timeseries(data, CommonFields.TEST_POSITIVITY, EMPTY_TS)
     if positive_negative_recent or not test_positivity.notna().any():
```
ghop02, line 147: This is probably re-litigating old code, but does this cause us to not use test positivity when calculated in `test_positivity.py` if we have positive tests and negative tests?  I guess we've disabled most inputs of that? 

I guess a large part of this is due to how we choose what goes into postive tests and negative tests in earlier parts of the code.

ghop02 (commented):
This all looks good to me - It would be helpful if you could add some hints as to the patterns we want in the future, not sure I can deduce the direction from just the code as it is

ghop02 (approved):

test/libs/metrics/top_level_metrics_test.py
TomGoBravo: I didn't see an obvious way to merge them because they each make a slightly bespoke DataFrame for MultiRegionDataset and the metrics output. I went the opposite direction and removed _build_timeseries_dataframe.

TomGoBravo:
> This all looks good to me - It would be helpful if you could add some hints as to the patterns we want in the future, not sure I can deduce the direction from just the code as it is

I'm working towards making functions and take and return OneRegionTimeseriesDataset / MultiRegionDataset objects with more use of provenance for tracking how data flowed so we can more easily retire unused parts.

PR #865, opened 2020-12-21
Test positivity calculate takes dataset

TomGoBravo:
## This PR

This PR is part of https://trello.com/c/6ZPv7k9T/477-clean-up-redundant-test-positivity-calculation-code-api-vs-combined-dataset

* Changes the test_positivity `calculate` methods to read from a dataset class instead of raw DataFrame
* Adds cached data access methods in `MultiRegionDataset` to replace calculations that were done in `AllMethods.run`
* Adds `_make_result_dataset` which builds a Series used by add_provenance_series, replacing the call to `add_provenance_all`. This is a step towards copying provenance per location which will replace the hack of having a CommonField per source. 
* Extends `build_dataset` to use a new timeseries literal class to optionally set provenance info.

## Tested

```
git checkout master
./run.py api generate-test-positivity --output-dir results/output-master/
git checkout tgb-test-pos-df-arg
./run.py api generate-test-positivity
```
Then compared output to find no diff.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -292,6 +292,7 @@ class MultiRegionDataset(SaveableDatasetInterface):
             .sort_index()
         )
 
+    @lru_cache(maxsize=None)
     def timeseries_wide_dates(self) -> pd.DataFrame:
         """Returns the timeseries in a DataFrame with LOCATION_ID, VARIABLE index and DATE columns."""
         if self.timeseries.empty:
@@ -309,6 +310,18 @@ class MultiRegionDataset(SaveableDatasetInterface):
             .rename_axis(columns=CommonFields.DATE)
         )
         return timeseries_wide
+
+    @lru_cache(maxsize=None)
+    def timeseries_wide_dates_variable_first(self) -> pd.DataFrame:
+        return self.timeseries_wide_dates().reorder_levels(
+            [PdFields.VARIABLE, CommonFields.LOCATION_ID]
+        )
+
+    @lru_cache(maxsize=None)
+    def timeseries_wide_dates_variable_first_diff(self, *, periods: int) -> pd.DataFrame:
+        # This calculates the difference only when the cumulative value is a real value `diff_days` apart.
+        # This only works on input data that has few or no holes.
+        return self.timeseries_wide_dates_variable_first().diff(periods=periods, axis=1)
 
     def _timeseries_latest_values(self) -> pd.DataFrame:
         """Returns the latest value for every region and metric, derived from timeseries."""
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -311,16 +311,6 @@ class MultiRegionDataset(SaveableDatasetInterface):
         )
         return timeseries_wide
 
-    def timeseries_wide_dates_variable_first(self) -> pd.DataFrame:
-        return self.timeseries_wide_dates().reorder_levels(
-            [PdFields.VARIABLE, CommonFields.LOCATION_ID]
-        )
-
-    def timeseries_wide_dates_variable_first_diff(self, *, periods: int) -> pd.DataFrame:
-        # This calculates the difference only when the cumulative value is a real value `diff_days` apart.
-        # This only works on input data that has few or no holes.
-        return self.timeseries_wide_dates_variable_first().diff(periods=periods, axis=1)
-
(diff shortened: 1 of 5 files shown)


test/libs/metrics/top_level_metrics_test.py
```diff
@@ -41,11 +43,22 @@
 ]
 
 
+class TimeseriesLit(UserList):
```
ghop02, line 46: TimseriesLit was pretty confusing to me until I read the comment, maybe `TimeseriesLiteral`?

libs/datasets/timeseries.py
```diff
@@ -310,6 +311,18 @@ def timeseries_wide_dates(self) -> pd.DataFrame:
         )
         return timeseries_wide
 
+    @lru_cache(maxsize=None)
+    def timeseries_wide_dates_variable_first(self) -> pd.DataFrame:
```
ghop02, line 315: fwiw, I think this might be more clear as an argument on `timeseries_wide_dates`, maybe:
```
def timeseries_wide_dates(self, variable_first_in_index=False)

```

libs/datasets/timeseries.py
```diff
@@ -310,6 +311,18 @@ def timeseries_wide_dates(self) -> pd.DataFrame:
...
+        return self.timeseries_wide_dates().reorder_levels(
+            [PdFields.VARIABLE, CommonFields.LOCATION_ID]
+        )
+
+    @lru_cache(maxsize=None)
+    def timeseries_wide_dates_variable_first_diff(self, *, periods: int) -> pd.DataFrame:
```
ghop02, line 321: this feels pretty specific to the test positivity dataset calculation - thoughts on just adding this where it's called?

libs/metrics/test_positivity.py
```diff
@@ -56,6 +56,46 @@ def _append_variable_index_level(df: pd.DataFrame, field_name: FieldName) -> Non
...
+
+    # TODO(tom): Add support for copying per-location provenance information from dataset_in.
+    #  Something like:
+    # assert dataset.provenance.index.names == [CommonFields.LOCATION_ID, PdFields.VARIABLE]
+    # provenance = dataset.provenance.loc[locations, self._column]
+    assert dataset_in.provenance.index.names == [CommonFields.LOCATION_ID, PdFields.VARIABLE]
+
+    # Make sure every timeseries has something in `provenance`.
+    provenance_index = pd.MultiIndex.from_product(
+        [locations, [result_metric]], names=[CommonFields.LOCATION_ID, PdFields.VARIABLE],
```
ghop02, line 89: Do you need to use `from_product` here since there's only one `result_metric`?

libs/metrics/test_positivity.py
```diff
@@ -56,6 +56,46 @@ def _append_variable_index_level(df: pd.DataFrame, field_name: FieldName) -> Non
     )
 
 
+def _make_result_dataset(
+    dataset_in: MultiRegionDataset,
+    default_provenance: str,
+    wide_date_df: pd.DataFrame,
+    result_metric: FieldName,
```
ghop02, line 63: `output_field`?

libs/metrics/test_positivity.py
```diff
@@ -56,6 +56,46 @@ def _append_variable_index_level(df: pd.DataFrame, field_name: FieldName) -> Non
     )
 
 
+def _make_result_dataset(
```
ghop02, line 59: fwiw I think `output` might be more clear here.  I'm finding this function a bit difficult to understand what's coming out - taking me a while to translate pandas operations into the resulting dataframe.

libs/datasets/timeseries.py
```diff
@@ -292,6 +292,7 @@ def _timeseries_long(self) -> pd.Series:
             .sort_index()
         )
 
+    @lru_cache(maxsize=None)
```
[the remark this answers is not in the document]
ghop02: I guess it might not be a problem because the underlying dataset is final, but if the underlying dataframe changes it could cause inconsistencies is one thing - just a pattern that I think is worth using very sparingly if there aren't other options available

ghop02 (approved):
Changes are great, thanks!

libs/datasets/timeseries.py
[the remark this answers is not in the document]
TomGoBravo: 100% can only do this when data is immutable. Part of the beauty of immutable objects is greater ease managing data derived from it.

PR #867, opened 2020-12-22
First working version of risk level based on 3 metrics + risk level timeseries

ghop02:
Still need to fix/update tests and docs, but this is my attempt at the risk level timeseries.  I went with the ugly approach - the whole risk level code needs a refactor, but don't think i can do that and get in this feature.


TomGoBravo (approved):
Please add a test

PR #868, opened 2020-12-22
Make recent_days configurable per Method

TomGoBravo:
## This PR

* Removes last of test positivity calculations from top_level_metrics, moving function `calculate_test_positivity` from top_level_metrics to test_positivity.
* Adds a new `test_positivity.Method` called `OldMethod` that reproduces what was done in top_level_metrics
* Makes recent_days configurable per Method because OldMethod used 10 days instead of 14
* Changes `test_positivity.Method.calculate` to return an `MethodOutput` object holding all timeseries for debugging and only recent timeseries as the desired output
* Changes `test_positivity.Method.name` from str to DatasetName because each method is used to produce one dataset
* Adds support for copying provenance from the input dataset to the output dataset in test_positivity
* Adds ugly hack/workaround so recent_days can have a default value in a super class.
* Hacks in OldMethodAll as a fall back when everything else is expired
* Moves `test_top_level_metrics_recent_pos_neg_tests_has_positivity_ratio` to `test_recent_pos_neg_tests_has_positivity_ratio` and use `read_csv...skipinitialspace=True` to make csv easier to read

## Tested

Ran `python ./run.py api generate-test-positivity` and found output identical to master *except* has some regions where the old method produces a timeseries that is all NA but considered valid and recent because the input timeseries is recent. Applying a fix to master in branch `tgb-positivity-old-calc-no-empty` got the same output.

I compared the API output for the state of NY in master and the branch using command such as `python ./run.py api generate-api-v2 -o output/output-ny-positivity-calc-in-methods --state NY output/output-ny-positivity-calc-in-methods` and the diffs looked insignificant.


libs/metrics/test_positivity.py
```diff
@@ -36,19 +55,32 @@ def columns(self) -> Set[FieldName]:
...
             df: DataFrame with rows having MultiIndex of [VARIABLE, LOCATION_ID] and columns with DATE
                 index. Must contain at least one real value for each of self.columns.
             delta_df: DataFrame with rows having MultiIndex of [VARIABLE, LOCATION_ID] and columns with DATE
                 index. Values are 7-day deltas. Must contain at least one real value for each of self.columns.
+            most_recent_date: The most recent date in dataset. Timeseries with a real value for
+                at least one of the dates within recent_days of this are considered recent.
         """
         pass
 
+    def _filter_recent(self, all: MultiRegionDataset, most_recent_date) -> MethodOutput:
```
ghop02, line 73: would prefer not to use `all` as it's a python reserved keyword.  

also, filter is a bit overloaded to me, maybe `remove_series_with_stale_data` or something similar might be more clear

libs/metrics/test_positivity.py
```diff
@@ -62,6 +94,8 @@ def _make_output_dataset(
     default_provenance: str,
     wide_date_df: pd.DataFrame,
     output_metric: FieldName,
+    *,
+    source_columns: Optional[List[FieldName]] = None,
```
ghop02, line 98: add to docstring

libs/metrics/test_positivity.py
```diff
@@ -80,41 +114,64 @@ def _make_output_dataset(
...
+    def name(self) -> timeseries.DatasetName:
         return self._name
 
     @property
     def columns(self) -> Set[FieldName]:
         return {self._numerator, self._denominator}
 
-    def calculate(self, dataset: MultiRegionDataset, diff_days: int) -> MultiRegionDataset:
+    def calculate(
+        self, dataset: MultiRegionDataset, diff_days: int, most_recent_date
```
ghop02, line 173: type on most_recent_date?

libs/metrics/test_positivity.py
```diff
@@ -126,25 +183,40 @@ def calculate(self, dataset: MultiRegionDataset, diff_days: int) -> MultiRegionD
         # DataFrame without the field label so operators such as `/` are calculated for each
         # region/state and date.
         wide_date_df = delta_df.loc[self._numerator, :] / delta_df.loc[self._denominator, :]
-        return _make_output_dataset(dataset, self._name, wide_date_df, CommonFields.TEST_POSITIVITY)
+
+        return self._filter_recent(
+            _make_output_dataset(dataset, self._name, wide_date_df, CommonFields.TEST_POSITIVITY),
```
ghop02, line 188: small nit, but I think it's helpful to have this as its own variable, so i don't have to keep as much context when i'm reading what's happening here

libs/metrics/test_positivity.py
```diff
@@ -155,10 +227,74 @@ def calculate(self, dataset: MultiRegionDataset, diff_days: int) -> MultiRegionD
...
         # to the index which was just taken out. Consider skipping reindexing.
-        return _make_output_dataset(dataset, self._name, wide_date_df, CommonFields.TEST_POSITIVITY)
+
+        return self._filter_recent(
+            _make_output_dataset(dataset, self._name, wide_date_df, CommonFields.TEST_POSITIVITY),
+            most_recent_date,
+        )
+
+
+class OldMethod(Method):
```
ghop02, line 237: nit: don't love the name `OldMethod` - maybe either rename it or add context as to why this method is old?

libs/metrics/test_positivity.py
```diff
@@ -155,10 +227,74 @@ def calculate(self, dataset: MultiRegionDataset, diff_days: int) -> MultiRegionD
...
+    """A method of calculating test positivity using smoothed POSITIVE_TEST and NEGATIVE_TESTS."""
+
+    @property
+    def name(self) -> timeseries.DatasetName:
+        return timeseries.DatasetName("OldMethod")
```
ghop02, line 242: hmm if this is used elsewhere I would lean towards changing it.  hard to know what it is without context

libs/metrics/test_positivity.py
```diff
@@ -155,10 +227,74 @@ def calculate(self, dataset: MultiRegionDataset, diff_days: int) -> MultiRegionD
...
+
+    def calculate(
+        self, dataset: MultiRegionDataset, diff_days: int, most_recent_date
+    ) -> MethodOutput:
+
+        positivity_time_series = {}
+        # To replicate the behavior of the old code, a region is considered to have recent
+        # positivity when the input timeseries (POSITIVE_TESTS and NEGATIVE_TESTS) are recent. The
+        # other subclasses of `Method` filter based on the most recent real value in the output
+        # timeseries.
```
ghop02, line 256: oo very helpful comment, thanks!

libs/metrics/test_positivity.py
```diff
@@ -155,10 +227,74 @@ def calculate(self, dataset: MultiRegionDataset, diff_days: int) -> MultiRegionD
...
+        recent_region = set()
```
ghop02, line 257: `recent_regions`? `regions_with_recent_data`?  also, fwiw do we expect there to ever be overlaps in that for loop? I would guess that as we're grouping by region, it wouldn't happen and we could just use a list

libs/metrics/test_positivity.py
```diff
@@ -216,36 +351,45 @@ def run(
...
+            name: method_output.recent for name, method_output in calculated_dataset_map.items()
         }
+        # HACK: If OldMethod is in the output map (which has Method.recent) then add it again at
+        # the end of the map using Method.all. Remember that dict entries remain in the order
+        # inserted. This makes OldMethod the final fallback for a location if no other Method has
+        # a timeseries for it.
+        old_method_output: Optional[MethodOutput] = calculated_dataset_map.get(
+            timeseries.DatasetName("OldMethod")
+        )
+        if old_method_output is not None:
```
ghop02, line 377: you can just do `if old_method_output` right?

ghop02 (commented):
Overall looks good! I'd like to take a slightly closer look after a meeting, but wanted to get the first round of comments to you :)

libs/metrics/test_positivity.py
TomGoBravo: Technically I'm conflicting with `all` builtin but it isn't reserved or a keyword. Changed to `ds_all`.
Yeah, I found 'filter' confusing too. Changed to `_remove_stale_regions`.

libs/metrics/test_positivity.py
TomGoBravo: done

libs/metrics/test_positivity.py
TomGoBravo: added pd.Timestamp, which seems to work and I think is right most of the time, though maybe str works too.

libs/metrics/test_positivity.py
TomGoBravo: done

libs/metrics/test_positivity.py
TomGoBravo: renamed to SmoothedTests

libs/metrics/test_positivity.py
TomGoBravo: It shows up in the provenance. Renamed.

libs/metrics/test_positivity.py
TomGoBravo: changed to list

libs/metrics/test_positivity.py
TomGoBravo: done

ghop02 (approved):

PR #869, opened 2020-12-23
Adding risk level to csv output

ghop02:
Tested this by running locally and verifying output in csv file - will add a trello ticket to add a real test to this part of the output


TomGoBravo (approved):
Please add something in comments and/or PR desc about how this was tested.

PR #870, opened 2020-12-23
Upgrade docusaurus to try and fix indexing issue

ghop02:
I'm trying to debug an issue where there is a weird flicker from the home page to the requested url on page load, I think this is causing google to think that the home page is every link.  Testing this locally I definitely get a more promising result - so think this version bump may actually work

### DIFF
--- a/api/docs/package.json
+++ b/api/docs/package.json
@@ -11,8 +11,8 @@
     "serve": "docusaurus serve"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


mikelehen (approved):

mikelehen:
LGTM.  If this doesn't fix it and you want any help debugging it, feel free to holler.  I have some familiarity with debugging google crawler issues.

PR #872, opened 2021-01-04
Add test_helpers module and remove _build_one_column_dataset"

TomGoBravo:
This PR
* Adds test_helpers module so it is easier to import without creating import cycles
* Adds `build_default_region_dataset`, a thin wrapper around `build_dataset`
* Removes `_build_one_column_dataset` which is easily replaced by `build_default_region_dataset`


test/test_helpers.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(
+        self, ts_list, *, provenance: str = "",
+    ):
+        super().__init__(ts_list)
+        self.provenance = provenance
+
+
+def build_dataset(
+    metrics: Mapping[Region, Mapping[FieldName, Union[Sequence[float], TimeseriesLiteral]]],
```
ghop02, line 37: I think, `metrics_by_region` might be more helpful as to what the mapping is

test/test_helpers.py
```diff
@@ -0,0 +1,118 @@
...
+    *,
+    start_date="2020-04-01",
+    dropna=True,
+) -> timeseries.MultiRegionDataset:
+    """Returns a dataset for multiple regions and metrics. Each sequence of values represents a
+    timeseries metric with identical length. provenance information can be set for a metric by
+    using a TimeseriesLiteral. Timeseries without any real values are dropped.
```
ghop02, line 44: Can you add args here?

test/test_helpers.py
```diff
@@ -0,0 +1,118 @@
...
+            dtype="str",
+            index=provenance_index,
+            name=PdFields.PROVENANCE,
+        )
+        dataset = dataset.add_provenance_series(provenance_series)
+
+    return dataset
+
+
+def build_default_region_dataset(
```
ghop02, line 88: how would you feel about: 
```
def build_single_region_dataset(metrics: Mapping..., *, region=DEFAULT_REGION, dropna=True)
```

lets you customize the region if you want

ghop02 (approved):
Looks good! couple suggestions on naming

test/test_helpers.py
TomGoBravo: I thought the type make it fairly clear. Changed to `metrics_by_region_then_field_name`

test/test_helpers.py
TomGoBravo: done

test/test_helpers.py
TomGoBravo: I'm down for that YAGNI.

PR #873, opened 2021-01-04
Bath County got too sick, replaced with Highland County

TomGoBravo:
See https://covidactnow.slack.com/archives/C0110QQAPJA/p1609803013425300


BrettBoval (approved):
LGTM. RIP Bath County. I'll miss you dearly.

PR #876, opened 2021-01-05
Move tail filter out of timeseries.py

ghop02:
timeseries.py has grown to be quite the behemoth - I'd like to start separating out code from it so it's easier to browse and comprehend.  I took a shot at the first part with moving the tail filter out into its own module


cli/data.py
```diff
@@ -25,11 +25,15 @@
 from libs.datasets import dataset_utils
 from libs.datasets import combined_datasets
 from libs.datasets.sources import forecast_hub
+from libs.datasets import tail_filter
 from libs.us_state_abbrev import ABBREV_US_UNKNOWN_COUNTY_FIPS
 from pyseir import DATA_DIR
 import pyseir.icu.utils
 from pyseir.icu import infer_icu
 
+TailFilter = tail_filter.TailFilter
```
TomGoBravo, line 34: what's the difference between this doing it with `from libs.datasets.tail_filter import TailFilter` ?

TomGoBravo (approved):
Please merge after I've gotten 868 merged.

cli/data.py
ghop02: Mostly following guidance from the [gsg](https://google.github.io/styleguide/pyguide.html#233-decision)

PR #879, opened 2021-01-07
Replace DC County Data with DC State data

ghop02:
Uses similar aggregation to NYC boroughs to replace DC county data with state data

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -10,6 +10,7 @@ from typing import Dict
 from typing import Iterable
 from typing import List, Optional, Union, TextIO
 from typing import Mapping
+from typing import Set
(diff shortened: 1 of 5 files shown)


libs/datasets/timeseries.py
```diff
@@ -313,6 +313,13 @@ class MultiRegionDataset:
     # DATE. Rows with identical index values may exist.
     tag: pd.Series = _EMPTY_TAG_SERIES
 
+    @property
+    def regions(self) -> Collection[Region]:
```
TomGoBravo, line 317: `timeseries_regions`?
Why not `Set[Region]` ? It looks like docstring is out of sync with method signature before it is merged the first time.

TomGoBravo (approved):

PR #880, opened 2021-01-07
Make test fixes found while pulling in county ICU data.

mikelehen:


libs/build_api_v2.py
```diff
@@ -37,7 +37,7 @@ def _build_actuals(actual_data: dict) -> Actuals:
         hospitalBeds={
             "capacity": actual_data.get(CommonFields.MAX_BED_COUNT),
             "currentUsageCovid": actual_data.get(CommonFields.CURRENT_HOSPITALIZED),
-            "currentUsageTotal": actual_data.get(CommonFields.CURRENT_HOSPITALIZED_TOTAL),
+            "currentUsageTotal": actual_data.get(CommonFields.HOSPITAL_BEDS_IN_USE_ANY),
```
mikelehen, line 40: I opened https://trello.com/c/cyWTrrDs/753-clean-up-commonfields-currenthospitalizedtotal-vs-hospitalbedsinuseany-confusion to clean this up fully, but as-is the currentUsageTotal field in the API is always null because combined data sets only pulls in HOSPITAL_BEDS_IN_USE_ANY.

I noticed this when I plumbed in the HHS data for counties and it didn't show up in the API how I expected.

test/pyseir_end_to_end_test.py
```diff
@@ -41,8 +41,8 @@ def test_pyseir_end_to_end_dc(tmp_path):
     with unittest.mock.patch("pyseir.utils.OUTPUT_DIR", str(tmp_path)):
         regions_dataset = combined_datasets.load_us_timeseries_dataset().get_subset(state="DC")
         regions = [one_region for _, one_region in regions_dataset.iter_one_regions()]
-        region_pipelines: List[OneRegionPipeline] = parallel_utils.parallel_map(
-            OneRegionPipeline.run, regions
+        region_pipelines: List[OneRegionPipeline] = list(
+            parallel_utils.parallel_map(OneRegionPipeline.run, regions)
```
mikelehen, line 45: On macos this was returning an `Iterator`, leading the test to fail on the `len()` call below. So I had to make it explicitly convert to `List`

test/libs/generate_api_v2_test.py
```diff
@@ -54,15 +54,15 @@ def test_build_summary_for_fips(
             negativeTests=nyc_latest["negative_tests"],
             hospitalBeds={
                 "capacity": nyc_latest["max_bed_count"],
-                "currentUsageCovid": None,
-                "currentUsageTotal": None,
+                "currentUsageCovid": nyc_latest["current_hospitalized"],
+                "currentUsageTotal": nyc_latest["hospital_beds_in_use_any"],
```
mikelehen, line 58: These all get populated once we plumb in the HHS data so I had to update the test.

ghop02 (approved):
Lgtm!

PR #881, opened 2021-01-07
Add tag and annotation to OneRegionTimeseriesDataset

TomGoBravo:
This PR is part of https://trello.com/c/Cq2zNk2M/674-annotate-data-outliers-that-weve-flagged-in-the-ui

* Copies all tags associated with a location_id to `OneRegionTimeseriesDataset` when creating it
* Makes `OneRegionTimeseriesDataset.provenance` from the tag DataFrame
* Adds `TagInTimeseries`, a frozen dataclass to represent a tag in the context of a particular timeseries. Use it in production code to iterate through tags and to replace named tuple `AnnotationInTimeseriesLiteral`. Unfortunately it is harder to make a literal dataclass than a tuple so adds `make_tag` to simplify making them in tests.
* Extracts code that turns a list of tags to a DataFrame to `make_tag_df` where it can be reused by a test calling `append_tag_df`


libs/datasets/timeseries.py
```diff
@@ -126,13 +136,28 @@ class OneRegionTimeseriesDataset:
...
-    provenance: Dict[str, str] = dataclasses.field(default_factory=dict)
+    # A default exists for convenience in tests. Non-test code is expected to explicitly set tag.
+    tag: pd.Series = dataclasses.field(
+        default_factory=lambda: _EMPTY_TAG_SERIES.reset_index(CommonFields.LOCATION_ID, drop=True)
+    )
 
-    # TODO(tom): find a way to expose all tags in this class or maybe add `attributes` as tags with
-    # a real date.
+    @property
+    def provenance(self) -> Dict[str, str]:
```
ghop02, line 145: I think eventually this could return a typed pydantic object (that shows up in the api)

ghop02 (approved):
looks good to me!  Couple small comments, but nothing blocking

libs/datasets/timeseries.py
```diff
@@ -114,6 +114,16 @@ class RegionLatestNotFound(IndexError):
     pass
 
 
+@final
+@dataclass(frozen=True)
+class TagInTimeseries:
```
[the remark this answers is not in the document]
TomGoBravo: Yeah, not a great name but I'm trying to be clear that this is only meaningful in the context of a particular timeseries. TimeseriesTag is more likely to include something that identifies the timeseries.

libs/datasets/timeseries.py
TomGoBravo: It can be whatever you want, as long as it is a read-only view of `tag` :-P

PR #883, opened 2021-01-11
test positivity copies tags to provide provenance output

TomGoBravo:
## This PR

* Adds support for writing and reading multiple provenance values for a single timeseries
* Sorts tags by content as well as index, simplifying comparisons between datasets
* Fixes `combined_datasets` bug with missing provenance causing KeyError in loc
* Changes test positivity code to only copy tags from source columns. This removes the clunky way we were copying the method name to provenance
* Removes `wide_dates_df` because all the interesting code is in MultiRegionDataset now.
* Adds test_helper support for multiple provenance and build_default_region_dataset start_date

## Tested

```
git checkout main
python ./run.py api generate-test-positivity --output-dir results/output-main/
git checkout tgb-tag-combine-populate
mkdir results/output-tgb-tag-combine-populate
python ./run.py api generate-test-positivity --output-dir results/output-tgb-tag-combine-populate
```

Diff of `results/output-main/test-positivity-output.csv` is DC county (`iso1:us#iso2:us-dc#fips:11001`) lost its provenance because it uses aggregation which doesn't forward tags. It was `SmoothedTests` which isn't in TestPositivityRatioMethod so likely already broken.

Diff of `results/output-main/test-positivity-all.csv` is a bunch of states (AK, FL, OH, OK, OR, TX, UT, ....) going from positiveTestsViral_totalTestsViral to covid_tracking, likely an improvement. Plus the DC change above.

In main `results/output-main/test-positivity-output-annotations.csv` is empty but with this change it has 128 rows of suspicious observations removed, mostly counties, mostly 1 or 2 a location, lots in AR and TX.


libs/datasets/timeseries.py
```diff
@@ -540,10 +541,24 @@ def from_csv(path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionDataset":
     def read_from_pointer(pointer: dataset_pointer.DatasetPointer) -> "MultiRegionDataset":
         wide_dates_df = pd.read_csv(pointer.path_wide_dates(), low_memory=False)
         wide_dates_df = wide_dates_df.set_index([CommonFields.LOCATION_ID, PdFields.VARIABLE])
+
         # Extract provenance column from wide date DataFrame so all columns are dates.
-        # TODO(tom): support multiple provenance columns
-        provenance_series = wide_dates_df[PdFields.PROVENANCE].dropna()
-        wide_dates_df = wide_dates_df.drop(columns=[PdFields.PROVENANCE])
+        provenance_column_mask = wide_dates_df.columns.str.match(
```
ghop02, line 546: Can you add a comment about why this is a match?

ghop02 (commented):
This looks good to me!  I'm a bit hesitant about the list of provenance values, but not hesitant enough (or don't have a good enough solution for an alternative) to provide any real feedback on it - I think it does a good job of passing the info through that we need!

ghop02 (approved):

libs/datasets/timeseries.py
TomGoBravo: Done

TomGoBravo:
Test failure is test/combined_dataset_test.py::test_combined_county_has_some_timeseries_data[48201] which seems to be because Harris County no longer has positive_tests:

```
        date = "2020-09-04"  # Arbitrary date that both FIPS have data for.
        assert df.loc[date, CommonFields.CASES] > 0
        assert df.loc[date, CommonFields.DEATHS] > 0
>       assert df.loc[date, CommonFields.POSITIVE_TESTS] > 0
E       assert nan > 0
```

PR #884, opened 2021-01-12
Enable filtering out values based on location scale values

ghop02:
Need to go through and clean up tests, but this allows us some flexibility to report values if at least 95% of the population is reporting

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -775,22 +775,14 @@ class MultiRegionDataset:
         tag = self.tag.loc[tag_mask]
         return dataclasses.replace(self, timeseries=timeseries_wide_variables, tag=tag)
 
-    def to_csv(
-        self, path: pathlib.Path, write_timeseries_latest_values=False,
-    ):
+    def to_csv(self, path: pathlib.Path):
         """Persists timeseries to CSV.
 
         Args:
             path: Path to write to.
-            write_timeseries_latest_values: write both explicit set static attributes and timeseries
-              values derived from timeseries. Mostly exists to compare to old files created when latest
-              values were calculated upstream.
         """
-        if write_timeseries_latest_values:
-            latest_data = self.static_and_timeseries_latest_with_fips().reset_index()
-        else:
-            latest_data = self.static.reset_index()
-            _add_fips_if_missing(latest_data)
+        latest_data = self.static.reset_index()
+        _add_fips_if_missing(latest_data)
 
         timeseries_data = self._geo_data.join(self.timeseries).reset_index()
         _add_fips_if_missing(timeseries_data)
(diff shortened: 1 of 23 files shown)


libs/datasets/timeseries.py
```diff
@@ -1098,29 +1119,19 @@ def _aggregate_dataframe_by_region(
...
-    if not ignore_na:
-        # Only keep aggregated values where the count of aggregated values is the same as the
-        # count of input regions.
-        long_agg = long_agg.loc[long_agg["count"] == long_agg["location_id_agg_count"]]
+    if reporting_ratio_required:
+        weighted_reporting_ratio = _calculate_weighted_reporting_ratio(
+            long_all_values, location_id_map, reporting_ratio_location_weights, groupby_columns
+        )
+        is_valid_reporting_ratio = weighted_reporting_ratio >= reporting_ratio_required
+        acceptable_ratios = weighted_reporting_ratio.loc[is_valid_reporting_ratio]
```
TomGoBravo, line 1130: not sure I'm following all the index but not sure you need acceptable_ratios if is_valid_reporting_ratio is a mask of locations with good ratios. How about

long_agg = long_agg.loc[is_valid_reporting_ratio]

libs/datasets/timeseries.py
```diff
@@ -1134,9 +1145,20 @@ def aggregate_regions(
...
 ) -> MultiRegionDataset:
-    """Produces a dataset with dataset_in aggregated using sum or weighted aggregation."""
+    """Produces a dataset with dataset_in aggregated using sum or weighted aggregation.
+
+    Args:
+        dataset_in: Input dataset.
+        aggregate_map: Region mapping input region to aggregate region.
+        aggregations: Sequence of aggregation overrides to apply aggregations other
+            than sum to fields.
+        reporting_ratio_required_to_aggregate: Ratio of locations per aggregate region required
```
TomGoBravo, line 1157: Say something about the ratio of scaling factor metric.

libs/datasets/timeseries.py
```diff
@@ -1134,9 +1145,20 @@ def aggregate_regions(
     aggregate_map: Mapping[Region, Region],
     aggregations: Sequence[StaticWeightedAverageAggregation] = WEIGHTED_AGGREGATIONS,
     *,
-    ignore_na: bool = False,
+    reporting_ratio_required_to_aggregate: Optional[float] = None,
```
TomGoBravo, line 1148: Why optional instead of float with default of 0.0 or 1.0?

TomGoBravo (commented):
I'd like some tests of the edge cases.

libs/datasets/timeseries.py
```diff
@@ -1174,12 +1186,18 @@ def aggregate_regions(
     # ... all other static input values.
     static_in_other_fields = static_in.loc[:, ~scale_fields_mask]
 
+    populations = static_in.loc[:, CommonFields.POPULATION]
```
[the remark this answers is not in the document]
ghop02: I think that's a next step if we want, but largely don't think it'll make that big of a difference

libs/datasets/timeseries.py
ghop02: I wanted to use none to specify that we're disabling filtering by ratio reported but can change it back

TomGoBravo (approved):

PR #889, opened 2021-01-13
Add annotations in API

TomGoBravo:
This PR builds towards https://trello.com/c/Cq2zNk2M/674-annotate-data-outliers-that-weve-flagged-in-the-ui

Chris, Michael and I drafted the API change in https://www.dropbox.com/scl/fi/4l8gcpbv7tsrzfzdqsc9w/Timeseries-annotations-_-tags-in-API.paper?dl=0&rlkey=co034g8saczcgn7nft7pwzo62


api/can_api_v2_definition.py
```diff
@@ -121,6 +133,45 @@ class ActualsTimeseriesRow(Actuals):
     date: datetime.date = pydantic.Field(..., description="Date of timeseries data point")
 
 
+class MetricAnomaly(base_model.APIBaseModel):
```
ghop02, line 136: Since we have a concept of metrics elsewhere do you think `DataAnomaly` or maybe `AnomalyAnnotation` would be more clear?

api/can_api_v2_definition.py
```diff
@@ -121,6 +133,45 @@ class ActualsTimeseriesRow(Actuals):
...
+    date: datetime.date = pydantic.Field(..., description="Date of anomaly")
+    description: str
+
+
+class MetricAnnotations(base_model.APIBaseModel):
+    """Annotations associated with one field."""
+
+    sources: List[MetricSources]
+    anomalies: List[MetricAnomaly]
```
ghop02, line 145: My initial thought was that would all be optional fields - as we add more, I think some metrics definitely won't have annotations? Was this a specific decision?

api/can_api_v2_definition.py
```diff
@@ -261,6 +312,7 @@ class RegionSummary(base_model.APIBaseModel):
     metrics: Metrics = pydantic.Field(...)
     riskLevels: RiskLevels = pydantic.Field(..., description="Risk levels for region.")
     actuals: Actuals = pydantic.Field(...)
+    annotations: Annotations = pydantic.Field(...)
```
ghop02, line 315: What are your thoughts on making annotations here a map of `[FieldName: MetricAnnotations]`? Actually I'm going to read the rest of the PR and see how the rest works first ;)

libs/build_api_v2.py
```diff
@@ -51,11 +61,13 @@ def _build_actuals(actual_data: dict) -> Actuals:
 
 
 def build_region_summary(
-    latest_values: dict,
+    one_region: timeseries.OneRegionTimeseriesDataset,
     latest_metrics: Optional[Metrics],
     risk_levels: RiskLevels,
-    region: pipeline.Region,
+    log,
```
ghop02, line 67: I know I bring this up every once in a while, but do we gain much from passing log through here? I find that passing log through bleeds the intentions of the calling function into sub functions - I don't think that it should be a requirement that if I want to build a region summary I need to have my own logging context set up.

libs/build_api_v2.py
```diff
@@ -73,6 +85,57 @@ def build_region_summary(
...
         locationId=region.location_id,
         url=latest_values[CommonFields.CAN_LOCATION_PAGE_URL],
+        annotations=build_annotations(one_region, log),
+    )
+
+
+def build_annotations(one_region: OneRegionTimeseriesDataset, log) -> Annotations:
+    assert one_region.tag.index.names == [TagField.VARIABLE, TagField.TYPE, TagField.DATE]
+    return Annotations(
+        cases=_build_metric_annotations(one_region.tag, CommonFields.CASES, log),
```
ghop02, line 95: you could also do:
```
fields_for_annotations = [CommonFields.CASES, CommonFields.DEATHS, ...]
annotations = {
    field: _build_metric_annotations(one_region.tag, field) 
    for field in fields_for_annotations
}
return Annotations(**annotations)
```

api/can_api_v2_definition.py
```diff
@@ -19,6 +19,18 @@ class TestPositivityRatioMethod(GetByValueMixin, enum.Enum):
     OTHER = "other"
 
 
+class MetricSources(GetByValueMixin, enum.Enum):
```
ghop02, line 22: How about `MetricSource`?

api/can_api_v2_definition.py
```diff
@@ -19,6 +19,18 @@ class TestPositivityRatioMethod(GetByValueMixin, enum.Enum):
...
+    """Metric source."""
```
ghop02, line 23: This is a bit broader comment, but I'd like to fill out the docstrings here a bit - since these descriptions will show up in the docs, it will show up in the docs and would be good to give people context about what these are

api/can_api_v2_definition.py
```diff
@@ -121,6 +133,45 @@ class ActualsTimeseriesRow(Actuals):
     date: datetime.date = pydantic.Field(..., description="Date of timeseries data point")
 
 
+class MetricAnomaly(base_model.APIBaseModel):
+    date: datetime.date = pydantic.Field(..., description="Date of anomaly")
+    description: str
```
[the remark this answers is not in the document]
ghop02: I wonder if each different tag type should be it's own annotation type

api/can_api_v2_definition.py
ghop02: ah I was thinking of metrics as the `metrics` and `metricsTimeseries` fields

api/can_api_v2_definition.py
ghop02: Ah cool, yeah that's correct - I think it'll be cleanest to not include fields that have have no annotations.  so the actual result would only have the keys of annotations that are there

api/can_api_v2_definition.py
TomGoBravo: I dunno. Changed to `AnomalyAnnotation`.

api/can_api_v2_definition.py
TomGoBravo: Most fields have a source so most will have a `MetricAnnotations`.

api/can_api_v2_definition.py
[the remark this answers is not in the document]
mikelehen: Yeah, in the doc we had sketched out `{ date: ..., originalValue: ..., description: ... }` I think. That still seems reasonable.

api/can_api_v2_definition.py
```diff
@@ -19,6 +19,19 @@ class TestPositivityRatioMethod(GetByValueMixin, enum.Enum):
     OTHER = "other"
 
 
+class MetricSource(GetByValueMixin, enum.Enum):
+    """The data source of a metric. This enumeration lists the places from which CAN fetches
+    data. The source is tracked on a per metric and region timeseries basis."""
```
mikelehen, line 24: In the API we have "metrics" and "actuals" where metrics are our ~5 core metrics (daily new cases, infection rate, test positivity, etc.). So perhaps rename this (and tweak comment) to "FieldSource" ?

api/can_api_v2_definition.py
```diff
@@ -121,6 +134,45 @@ class ActualsTimeseriesRow(Actuals):
     date: datetime.date = pydantic.Field(..., description="Date of timeseries data point")
 
 
+class AnomalyAnnotation(base_model.APIBaseModel):
+    date: datetime.date = pydantic.Field(..., description="Date of anomaly")
+    description: str
+
+
+class MetricAnnotations(base_model.APIBaseModel):
```
mikelehen, line 142: FieldAnnotations?

api/can_api_v2_definition.py
```diff
@@ -121,6 +134,45 @@ class ActualsTimeseriesRow(Actuals):
...
+    )
+    hospitalBeds: Optional[MetricAnnotations] = pydantic.Field(
+        None, description="Annotations for hospitalBeds"
+    )
+    icuBeds: Optional[MetricAnnotations] = pydantic.Field(
+        None, description="Annotations for icuBeds"
+    )
+    newCases: Optional[MetricAnnotations] = pydantic.Field(
+        None, description="Annotations for newCases",
+    )
```
mikelehen, line 173: Where did this list of 8 fields come from?  It seems like a subset of the fields in our API...?

mikelehen (commented):
Couple comments on the API.  Didn't dig into the other details (sorry) but can if it'd be valuable.

Per standup discussion, I think we want to get outlier detection anomalies plumbed in here as well before we would actually ship this on the frontend, but that can be a separate PR if it makes more sense.

mikelehen:
Oh! I see you added the outlier detection tags in https://github.com/covid-projections/covid-data-model/pull/891 so you can ignore that comment. 😁

api/can_api_v2_definition.py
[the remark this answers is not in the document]
TomGoBravo: How do you suggest we represent annotations on disk? Currently they are in a CSV with columns in https://github.com/covid-projections/covid-data-model/blob/main/libs/datasets/timeseries.py#L46-L67
I'm leaning towards having the content be a JSON or bare string depending on the annotation type. The JSON will look kind of ugly when put in a CSV, but lets each annotation type have its own attributes.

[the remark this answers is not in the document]
mikelehen: I'm inclined to defer to @ghop02, but your proposal doesn't sound unreasonable. 😁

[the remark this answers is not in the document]
TomGoBravo: It might be overkill but I created https://github.com/covid-projections/covid-data-model/pull/897

api/can_api_v2_definition.py
TomGoBravo: Done in f7375278

api/can_api_v2_definition.py
TomGoBravo: Done in 98ca0c5e

api/can_api_v2_definition.py
TomGoBravo: From Actuals. Added comment in 41a560c8 to keep them in sync. We could add annotations for metrics too, but I haven't tested that code path. From a glance I think only test positivity will have them.

libs/build_api_v2.py
```diff
@@ -81,9 +92,54 @@ def build_region_summary(
...
+def _build_metric_annotations(
+    tag_series: timeseries.OneRegionTimeseriesDataset, field_name: CommonFields, log
+) -> Optional[FieldAnnotations]:
+
+    sources_enum = []
+    for source_str in tag_series.provenance.get(field_name, []):
+        source_enum = FieldSource.get(source_str)
+        if source_enum is None:
+            source_enum = FieldSource.OTHER
+            log.info(
```
ghop02, line 126: Will this be noisy?

ghop02 (approved):
Lgtm!

libs/build_api_v2.py
TomGoBravo: Maybe, but adding enums is easy.

PR #891, opened 2021-01-13
Add annotations in drop_new_case_outliers

TomGoBravo:
This PR is part of https://trello.com/c/Cq2zNk2M/674-annotate-data-outliers-that-weve-flagged-in-the-ui

* in drop_new_case_outliers add an annotation for each removed value

## Tested

Ran `./run.py data update` which produced a 199k data/multiregion-annotations.csv up from 26k. Looking at `git difftool -t vimdiff --no-prompt main -- data/multiregion-annotations.csv` the diffs look as expected, adding a bunch of zscore_outlier in states and counties. A few regions have multiple outliers removed.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -999,9 +999,12 @@ def drop_new_case_outliers(
     grouped_df = timeseries.groupby_region()
 
     zscores = grouped_df[CommonFields.NEW_CASES].apply(_calculate_modified_zscore)
+    to_exclude = (zscores > zscore_threshold) & (df_copy[CommonFields.NEW_CASES] > case_threshold)
 
     new_tags = []
-    to_exclude = (zscores > zscore_threshold) & (df_copy[CommonFields.NEW_CASES] > case_threshold)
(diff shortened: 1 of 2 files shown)


libs/datasets/timeseries.py
```diff
@@ -999,10 +1000,25 @@ def drop_new_case_outliers(
 
     zscores = grouped_df[CommonFields.NEW_CASES].apply(_calculate_modified_zscore)
 
+    new_tags = []
     to_exclude = (zscores > zscore_threshold) & (df_copy[CommonFields.NEW_CASES] > case_threshold)
-    df_copy.loc[to_exclude, CommonFields.NEW_CASES] = None
+    for idx, _ in to_exclude[to_exclude].iteritems():
```
ghop02, line 1005: Should this be `df_copy[to_exclude]`?

ghop02 (approved):

libs/datasets/timeseries.py
TomGoBravo: Yeah, that looks confusing. Either works. Added comment.

PR #893, opened 2021-01-14
Drop ccd testing and hospitalization data

ghop02:
Removes using old covid county data API so that we are relying on CDC data until we get the new more specific data from the CAN Scrapers


libs/datasets/combined_datasets.py
```diff
@@ -81,21 +81,13 @@ class RegionLatestNotFound(IndexError):
...
-        CovidTrackingDataSource,
-        HHSTestingDataset,
-    ],
-    CommonFields.POSITIVE_TESTS: [
-        CovidCountyDataDataSource,
-        CovidTrackingDataSource,
-        HHSTestingDataset,
-    ],
+    CommonFields.HOSPITAL_BEDS_IN_USE_ANY: [CovidCountyDataDataSource, HHSHospitalDataset],
+    CommonFields.ICU_BEDS: [CovidCountyDataDataSource, HHSHospitalDataset],
```
mikelehen, line 88: Why are we keeping it for hospital fields?

TomGoBravo (approved):

PR #894, opened 2021-01-14
Update pydantic + Exclude unset fields in json serialization

ghop02:
This updates the version of pydantic (seems to shave 1 min of api build) + excluding unset fields, this primarily lets us remove `icuHeadroomDetails` and `testPositivityRatioDetails` from the json output, which shaves about 75MB off of the counties.timeseries.json file. 

Any field that is included but null (which is most fields) will be included in the output


mikelehen (approved):
Nice, LGTM!

PR #895, opened 2021-01-14
Add vaccines to api

ghop02:
This gets vaccination data flowing through to the API. Note that vaccinations completed won't have any data just yet, but now that the data is available, looks like it will sooon!

I also am filtering out including vaccination data before the date of first us vaccination.  

I confirmed that loading the timeseries in pandas works just fine with some records not having all of the fields.

### DIFF
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -113,6 +113,18 @@ Notable exceptions:
  2. Any days with negative new cases are removed.
 """,
     )
+    vaccinesDistributed: Optional[int] = pydantic.Field(
+        ..., description="Number of vaccine doses distributed."
+    )
+    vaccinationsInitiated: Optional[int] = pydantic.Field(
+        ...,
+        description="""
+Number of vaccinations initiated.
+
+This value may vary by type of vaccine, but for Moderna and Pfizer, this indicates
+number of people vaccinated with the first dose.
+""",
+    )
 
 
 class ActualsTimeseriesRow(Actuals):
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -131,7 +131,7 @@ number of people vaccinated with the first dose.
 Number of vaccinations completed.
 
 This value may vary by type of vaccine, but for Moderna and Pfizer, this indicates
-number of people vaccinated with the both first and second dose.
+number of people vaccinated with both the first and second dose.
 """,
     )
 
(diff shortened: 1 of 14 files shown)


api/can_api_v2_definition.py
```diff
@@ -111,6 +111,27 @@ class Actuals(base_model.APIBaseModel):
...
+number of people vaccinated with the first dose.
+""",
+    )
+    vaccinationsCompleted: Optional[int] = pydantic.Field(
+        None,
+        description="""
+Number of vaccinations completed.
+
+This value may vary by type of vaccine, but for Moderna and Pfizer, this indicates
+number of people vaccinated with the both first and second dose.
```
mikelehen, line 134: "the both" => "both the"

libs/build_api_v2.py
```diff
@@ -87,8 +95,16 @@ def build_region_timeseries(
     for row in timeseries.yield_records():
         # Timeseries records don't have population
         row[CommonFields.POPULATION] = region_summary.population
-        actual = _build_actuals(row)
-        timeseries_row = ActualsTimeseriesRow(**actual.dict(), date=row[CommonFields.DATE])
+        actual = _build_actuals(row).dict()
+
+        # Don't include vaccinations in timeseries before first possible vaccination
+        # date to not bloat timeseries.
+        if row[CommonFields.DATE] < USA_VACCINATION_START_DATE:
```
TomGoBravo, line 102: test svp

TomGoBravo (approved):

libs/build_api_v2.py
mikelehen, line 102: Why are we doing this by date instead of just checking for nulls or something?  This feels unnecessarily fragile (if we backfill the vaccine data we'll have to update this).

mikelehen: That said, I don't care too strongly. 😁

ghop02: This is meant to actually handle backfilling - if we were to backfill, any data included after this date will be included (and this date is the first day of vaccination for the US, so there can't be vaccinations from any day earlier). 

If we were just to remove nulls, if a state or county didn't report for a day, (say in the middle of a timeseries) we would remove the field, which to me feels less consistent than saying "this field starts on december 14th"

mikelehen: Ah, right on.  Makes sense! 👍

PR #896, opened 2021-01-14
Remove requirement for required field

ghop02:
We were checking for the is_required field, which means that it wasn't properly typing fields that may be undefined. 

Here's the difference in typescript output from the schemas:
```diff
-vaccinesDistributed?: number;
-vaccinationsInitiated?: number;
-vaccinationsCompleted?: number;
+vaccinesDistributed?: number | null;
+vaccinationsInitiated?: number | null;
+vaccinationsCompleted?: number | null;
```

### DIFF
--- a/libs/base_model.py
+++ b/libs/base_model.py
@@ -26,10 +26,7 @@ class APIBaseModel(pydantic.BaseModel):
 
                 # Checking for fields that allow none (essentially indicating that the
                 # type is Optional[<type>]) and required.
-                # This is possibly more stringent than necessary (field.required may not be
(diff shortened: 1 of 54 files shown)


libs/base_model.py
```diff
@@ -26,10 +26,7 @@ def schema_extra(schema, model):
 
                 # Checking for fields that allow none (essentially indicating that the
                 # type is Optional[<type>]) and required.
-                # This is possibly more stringent than necessary (field.required may not be
```
TomGoBravo, line 29: Comment above this line looks stale since you are removing field.required

TomGoBravo (approved):

PR #897, opened 2021-01-15
Change tag content to JSON

TomGoBravo:
This PR builds towards https://trello.com/c/Cq2zNk2M/674-annotate-data-outliers-that-weve-flagged-in-the-ui

* Moves `date` from a field in the tag/annotation CSV and DataFrame to a field in the JSON content of annotations. This replaces the human readable strings buried in our data pipeline with flexible JSON content. The downside is processing the JSON in pandas is harder. The upside is this makes adding new attributes per annotation *much* easier. Removing the date will also make it easier to include all the annotations in the wide-dates CSV so we can later remove multiregion-annotations.csv.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -1,4 +1,7 @@
+import collections
 import dataclasses
+import json
+from abc import ABC, abstractmethod
 import datetime
 import enum
 import pathlib
@@ -6,6 +9,7 @@ import re
 from dataclasses import dataclass
 from functools import lru_cache
 from typing import Any
+from typing import ClassVar
 from typing import Collection
 from typing import Dict
 from typing import Iterable
@@ -42,6 +46,8 @@ from backports.cached_property import cached_property
 _log = structlog.get_logger()
 
 
+# TODO(tom): Move all this tag stuff to a separate file, maybe called taglib so that code can
+#  import it and still use `tag` as a local name.
 @enum.unique
 class TagField(GetByValueMixin, ValueAsStrMixin, FieldName, enum.Enum):
     """The attributes of a tag, columns in a table with one row per tag."""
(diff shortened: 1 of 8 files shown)


libs/datasets/timeseries.py
```diff
@@ -82,20 +76,146 @@ class TagType(GetByValueMixin, ValueAsStrMixin, str, enum.Enum):
...
+TAG_TYPE_TO_CLASS = {
+    TagType.CUMULATIVE_TAIL_TRUNCATED: CumulativeTailTruncated,
+    TagType.CUMULATIVE_LONG_TAIL_TRUNCATED: CumulativeLongTailTruncated,
+    TagType.ZSCORE_OUTLIER: ZScoreOutlier,
+    TagType.PROVENANCE: ProvenanceTag,
+}
+
+
+@dataclass(frozen=True)
+class TagCollection:
```
ghop02, line 197: Thoughts on throwing this in another module?

test/test_helpers.py
```diff
@@ -58,8 +56,28 @@ def make_tag(
     type: TagType = TagType.CUMULATIVE_TAIL_TRUNCATED,
     date: Union[pd.Timestamp, str] = "2020-04-02",
     content: str = "taggy",
+    **kwargs,
 ) -> timeseries.TagInTimeseries:
-    return timeseries.TagInTimeseries(type, pd.to_datetime(date), content)
+    date: pd.Timestamp = pd.to_datetime(date)
+    if type == TagType.PROVENANCE:
+        assert content != "taggy"
```
ghop02, line 63: What are these checks for?

libs/datasets/timeseries.py
```diff
@@ -82,20 +76,146 @@ class TagType(GetByValueMixin, ValueAsStrMixin, str, enum.Enum):
...
+
+    _location_var_map: MutableMapping[Tuple, List[TagInTimeseries]] = dataclasses.field(
+        default_factory=lambda: collections.defaultdict(list)
+    )
+
+    def add(self, tag: TagInTimeseries, *, location_id: str, variable: CommonFields) -> None:
+        """Adds a tag to this collection."""
+        self._location_var_map[(location_id, variable)].append(tag)
+
+    def _as_records(self) -> Iterable[Mapping]:
```
ghop02, line 209: maybe
```
@property
def _records(self)
```
and 
```
@property
def dataframe(self)
```
?

but don't feel strongly
```

ghop02 (approved):
This looks good to me!

test/test_helpers.py
TomGoBravo: They were leftover from a transition from a previous version of the code. removed.

PR #898, opened 2021-01-18
Remove annotations keys when writing to csv

ghop02:
Adds support for removing entire sub keys and removes annotations from csvs for now

### DIFF
--- a/libs/dataset_deployer.py
+++ b/libs/dataset_deployer.py
@@ -1,4 +1,5 @@
 from typing import List, Optional
+import re
 import enum
 import os
 import pathlib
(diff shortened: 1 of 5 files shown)


libs/dataset_deployer.py
```diff
@@ -78,14 +78,20 @@ def write_nested_csv(
...
     """
+    keys_to_skip = keys_to_skip or []
+
     if not data:
         raise ValueError("Cannot upload a 0 length list.")
-    header = flatten_dict(data[0]).keys()
-    if keys_to_skip:
-        header = [column for column in header if column not in keys_to_skip]
+
+    first_row = data[0]
```
TomGoBravo, line 90: Add a comment, maybe
`# Skip top level keys before passing first_row to flatten_dict`

TomGoBravo (approved):

PR #900, opened 2021-01-19
Fix aggregation required rate

ghop02:
Turns out I wasn't passing the required reporting ratio through 🤦 so it was just aggregating all data regardless if it has missing data or not.  I'm running a snapshot to double check that these changes will turn out to reflect reality

### DIFF
--- a/test/libs/datasets/statistical_areas_test.py
+++ b/test/libs/datasets/statistical_areas_test.py
@@ -1,11 +1,15 @@
 import dataclasses
 
+import pytest
+import pandas as pd
+from covidactnow.datapublic.common_fields import CommonFields
+
 from libs.datasets import statistical_areas
 from libs.datasets.timeseries import MultiRegionDataset
-import pandas as pd
 
 from libs.pipeline import Region
 from test.dataset_utils_test import read_csv_and_index_fips_date
+from test import test_helpers
 
 
 def test_load_from_local_public_data():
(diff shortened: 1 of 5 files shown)


test/libs/datasets/statistical_areas_test.py
```diff
@@ -60,3 +64,29 @@ def test_aggregate():
         pd.to_datetime("2020-05-03"): 6,
         pd.to_datetime("2020-05-04"): 4,
     }
+
+
+@pytest.mark.parametrize("reporting_ratio,expected_na", [(0.8, False), (0.91, True)])
+def test_aggregate_reporting_ratio(reporting_ratio, expected_na):
+    ny_region = Region.from_state("NY")
+    az_region = Region.from_state("AZ")
+    us_region = Region.from_iso1("us")
```
TomGoBravo, line 73: unless there is good reason to use state -> US aggregation please use region types expected by CountyToCBSAAggregator.

TomGoBravo (approved):
I created https://trello.com/c/eXAtVkZO/807-expand-scope-of-lint-static-analysis-checks-to-catch-suspect-code

test/libs/datasets/statistical_areas_test.py
ghop02: The best reason is for creating the test dataset, which expects regions

TomGoBravo: It is really just as easy to make county and CBSA regions: https://github.com/covid-projections/covid-data-model/blob/main/libs/pipeline.py#L108-L132

PR #907, opened 2021-01-22
Add function to backfill vaccinations initiated

ghop02:


libs/datasets/timeseries.py
```diff
@@ -1150,6 +1150,26 @@ def drop_regions_without_population(
     return mrts.get_locations_subset(locations_with_population)
 
 
+def _compute_derived_vaccinations_initiated(dataset: MultiRegionDataset):
+    administered = dataset.timeseries[CommonFields.VACCINES_ADMINISTERED]
+    completed = dataset.timeseries[CommonFields.VACCINATIONS_COMPLETED]
+    derived_initiated = administered - completed
```
ghop02, line 1156: I only want to do this if a region has no data at all for vaccinations_initiated, what's the best way to do that?

libs/datasets/timeseries.py
```diff
@@ -1150,6 +1150,26 @@ def drop_regions_without_population(
...
+    initiated = dataset.timeseries[CommonFields.VACCINATIONS_INITIATED]
+
+    return initiated.fillna(derived_initiated)
```
ghop02, line 1159: I only want to fill for regions that have no vaccinations_initiated data at all. what's the best way to do this?

libs/datasets/timeseries.py
```diff
@@ -1160,6 +1160,66 @@ def drop_regions_without_population(
...
+    """
+    fields = [
+        CommonFields.VACCINES_ADMINISTERED,
+        CommonFields.VACCINATIONS_INITIATED,
+        CommonFields.VACCINATIONS_COMPLETED,
+    ]
+    df = dataset.timeseries_wide_dates().loc[(slice(None), fields), :]
+    df_var_first = df.reorder_levels([PdFields.VARIABLE, CommonFields.LOCATION_ID])
+
+    administered = df_var_first.loc[CommonFields.VACCINES_ADMINISTERED]
```
TomGoBravo, line 1179: is the different indexing without nested [] from next two lines intentional?

TomGoBravo (approved):

PR #908, opened 2021-01-22
Add GA event to signup form + update meta description

ghop02:
Worked on updated meta description with gaia + add in GA event on API signup


pnavarrc (approved):
Looks good!

PR #909, opened 2021-01-22
Snapshot with truncated bulk histories

ghop02:
Here's a feature I've been meaning to add for a while:

Creates a couple more aggregate files `states.timeseries.30d.json`, `counties.timeseries.30d.json`, `counties.timeseries.90d.json`, etc. that truncate the history at 30 days or 90 days.

Then on the api, I added a query arg `daysBack` that lets you specify either 30 or 90 days back.  The cloudfront edge function modifies the s3 uri to request the appropriate file.

so a query of `https://api.covidactnow.org/v2/counties.timeseries.json?daysBack=30` would just return the `counties.timeseries.30d.json` file.

This should help immensely with location summary issues (and i've heard similar issues from other customers).  The counties.timeseries.json file is currently 742 MB, in contrast, the 30d county file is 81MB. 

Currently this is only for the json files, but could make for the csvs as well

### DIFF
--- a/api/awsauth/awsauth/auth_app.py
+++ b/api/awsauth/awsauth/auth_app.py
@@ -1,3 +1,4 @@
+from typing import Optional
 import uuid
 import urllib.parse
 import datetime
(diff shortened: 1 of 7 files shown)


ghop02:
I think I'm going to close this in favor of splitting counties by state

PR #910, opened 2021-01-25
Deploys counties by state

ghop02:
This PR is meant to help make county data more accessible.  Right now, to get all counties for a state you either need to do individual queries for every county or download the bulk counties files.  Unfortunately those are getting prohibitively large (with the counties.timeseries.json getting up to 850MB). 

After this gets in, we can also change the location summary updater to query by state so that we aren't stuck on one version of node

### DIFF
--- a/api/docs/open_api_schema.json
+++ b/api/docs/open_api_schema.json
@@ -594,6 +594,202 @@
         }
       ]
     },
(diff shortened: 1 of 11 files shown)


ghop02:
I think Im' going to close this in favor of splitting counties by state

TomGoBravo (approved):

PR #914, opened 2021-01-25
combined_datasets with dataset objects

TomGoBravo:
## This PR

This is inspired by brainstorms in https://www.dropbox.com/scl/fi/wlchtiqvq0cs9aj5lnbfm/Importing-CAN-scraper-data.paper . It doesn't add a new representation for creating subsets of DataSource objects per region but does let us re-use the existing `get_subset` to add subsets to the per-field lists in combined_datasets.

* changes `timeseries.combined_datasets` to take map with dataset values instead of names. this makes it easier to inject data with different region subsets
* adds a bunch of caching to replace the reuse of values that happened when they were referred to by names

## Tested


libs/datasets/combined_datasets.py
```diff
@@ -53,6 +53,11 @@ class RegionLatestNotFound(IndexError):
 )
 
 
+class HHSHospitalDatasetOnlyTX(HHSHospitalDataset):
+    def multi_region_dataset(self) -> MultiRegionDataset:
+        return super().multi_region_dataset().get_subset(state="TX")
```
mikelehen, line 58: Huh, neat!  I guess this is roughly equivalent to the doc proposal (you can add a [regions+datasource] to the prioritized list of sources for a field), but:
1. This is a lot of boilerplate (creating a class every time we need to block / reprioritize a region+datasource) that people not intimate with this codebase (e.g. Pablo) won't understand if they're trying to fix a snapshot deploy.
2. There's an extra level of indirection (the named `HHSHospitalDatasetOnlyTX` class will probably only be used once and introduces the possibility of the name getting out-of-sync with the definition, especially for more complicated region subsets that don't have a clear name).
3. We don't have an easy `exclude` mechanism (you'd have to do `get_subset(states=[state for state in ALL_STATES if state != "TX"])` or something?
4. FWIW- I can't immediately tell if this is including Texas (fips 48) or all counties in Texas (fips 48xxx) or both.  We should ideally introduce a syntax that's more explicit to avoid confusion.

So I like this but think it's worth implementing in a bit cleaner / more explicit way so that I, Pablo, Brett, etc. can look at the field mapping list and immediately understand what data sources we're using for what regions and make updates as necessary, without having to define new classes or be good at Python. 😁 

Could we do some sort of chained syntax directly off of `DataSource` instead of defining a new subclass? Like if we wanted to use can-scrapers for TX and IL counties, but not Cook County, it would be something like:

`CANScraperStateProviders.include(Counties(state="TX"), Counties(state="IL")).exclude(County(fips="17031"))`

I think this gets us a lot closer to what we suggested in the doc, right?

FWIW- I'd also be open to just using FIPS regexes to make it a bit less type-heavy.

`CANScraperStateProviders.include("48...", "17...").exclude("17031")`

Thoughts?  @ghop02 Want to weigh in too?

mikelehen (commented):
👍  but I wonder if we can make this a little cleaner and closer to the original proposal...

libs/datasets/combined_datasets.py
TomGoBravo: Thank you for taking a look at this PR. It changes `timeseries.combined_datasets` so it is fairly easy to use different subsets of the original DataSource datasets. I'd like to cleanup the DataSource constants before streamlining how we declare what subsets are used where. I get that a big declare structure is probably the easiest way to manipulate how data is combined at a region level, but still don't understand why this seems a better way to understand what data is being used because some fields have many datasources and a casual reader won't know what regions are coming from what source.

TomGoBravo: To be a bit more clear: I think this PR is a useful meaningful step towards what you'd like. A subsequent PR can improve the syntax for declaring `ALL_TIMESERIES_FEATURE_DEFINITION` :crossed_fingers: without changing `timeseries.combined_datasets`. I think getting the syntax in the doc working isn't a huge stretch, it'd involve moving the logic in `HHSHospitalDatasetOnlyTX` to `data.py`, driven by some new tuple/frozen dataclass.

mikelehen: Cool, if we think this is moving us in the right direction, then I'm comfortable with it. Perhaps you, @ghop02, and I can sync at / after standup. 

Regarding:

> still don't understand why this seems a better way to understand what data is being used

I don't think this is about _understanding_ what data is being used. There's a product need to improve how we _control_ what data is being used.

1. Brett has specifically asked for better visibility into what data sources are enabled / disabled and why, without having to hunt across our codebase for one-off hacks.
2. Right now we (me especially maybe) err overly towards disabling things in the frontend because that's easier than figuring out the right place to patch things in the data pipeline.
3. I suspect we're going to need to make priority overrides as soon as we turn on the can-scraper data for test positivity, akin to the one-off hack I had done in https://github.com/covid-projections/covid-data-model/blob/9abf70794ac837298885ed41e010164a352a25ff/libs/datasets/sources/covid_county_data.py#L15 .  I'd rather not repeat that hack, per 1 above.

I think having visibility into what data sources _actually_ got used (based on what data is available from each data source) is also important, but it's not as pressing of a need (i.e. doesn't block us from using can-scraper test positivity)...  and I think we're already headed on a good path for that with the annotation work, right?

cli/data.py
```diff
@@ -83,26 +90,16 @@ def update(aggregate_to_country: bool, state: Optional[str], fips: Optional[str]
...
-        chain(
-            chain.from_iterable(ALL_FIELDS_FEATURE_DEFINITION.values()),
-            chain.from_iterable(ALL_TIMESERIES_FEATURE_DEFINITION.values()),
-        )
-    )
-    data_sources = {
-        data_source_cls.SOURCE_NAME: data_source_cls.local().multi_region_dataset()
-        for data_source_cls in data_source_classes
-    }
+    timeseries_field_datasets = build_field_datasets(ALL_TIMESERIES_FEATURE_DEFINITION)
```
ghop02, line 93: fwiw, I have a hard time parsing `build_field_datasets` - mostly I didn't know what `field_datasets` was (without thinking about it).  How about something like `load_datasets_by_field`, `load_dataset_list_by_field` or `load_multiregion_datasets_by_field`?

cli/data.py
```diff
@@ -237,12 +234,22 @@ def update_case_based_icu_utilization_weights():
         json.dump(output, f, indent=2, sort_keys=True)
 
 
-def build_field_dataset_source(feature_definition_config):
+def build_field_datasets(
+    feature_definition_config: combined_datasets.FeatureDataSourceMap,
```
ghop02, line 238: How about doing the filtering here? instead of in a separate method.

```python
def build_field_datasets(
    feature_definition_config: combined_datasets.FeatureDataSourceMap,
    state=None,
    fips=None
):

    def _load_dataset(data_source_cls):
        dataset = data_source_cls.local().multi_region_dataset()
        if state or fips:
            dataset = dataset.get_subset(state=state, fips=fips)
        return dataset
    
    feature_definition = {
        field_name: list(reversed(list(_load_dataset(cls) for cls in classes)))
        for field_name, classes in feature_definition_config.items()
        if classes
    }
    return feature_definition    
```

cli/data.py
TomGoBravo: Done

cli/data.py
TomGoBravo: I've gone back and forth on this but hadn't tried an internal function. Done.

ghop02 (approved):

TomGoBravo:
[CAN Scraper: Clean up / document mechanism for disabling/overriding data for a problematic (provider, region) pair.](https://trello.com/c/VP9NRpJe/778-can-scraper-clean-up-document-mechanism-for-disabling-overriding-data-for-a-problematic-provider-region-pair)

PR #916, opened 2021-01-27
rm DataSource FIELD_MAP constants

TomGoBravo:
This PR addresses https://trello.com/c/uPQXBuDg/777-remove-unused-datasource-constants

## This PR

This PR changes most DataSource subclasses to be just a few constants without custom code. The exceptions are sources that produce static data or make some custom tweak to the produced dataset.

* Removes `DataSource.local()` which created in intermediate DataFrame that is no longer useful.
* Replaces calls to `data_source_cls.local().multi_region_dataset()` with `data_source_cls.make_dataset()`
* Removes `DataSource.INDEX_FIELD_MAP` and `DataSource.COMMON_FIELD_MAP` which stopped being useful when we started doing the transformations in the covid-data-public repo
* Adds constants `DataSource.EXPECTED_FIELDS` `DataSource.COMMON_DF_CSV_PATH` and `DataSource.IGNORED_FIELDS` that are used in the default `make_dataset` to load timeseries data

## Tested


libs/datasets/data_source.py
```diff
@@ -1,59 +1,65 @@
...
-        if set(self.INDEX_FIELD_MAP.keys()) == set(TIMESERIES_INDEX_FIELDS):
-            dataset = MultiRegionDataset.from_fips_timeseries_df(self.data).add_provenance_all(
-                self.SOURCE_NAME
+    def make_dataset(cls) -> timeseries.MultiRegionDataset:
+        """Default implementation of make_dataset that loads timeseries data from a CSV."""
+        assert cls.COMMON_DF_CSV_PATH, f"No path in {cls}"
+        data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
+        input_path = data_root / cls.COMMON_DF_CSV_PATH
+        data = common_df.read_csv(input_path, set_index=False)
+        expected_fields = pd.Index({*cls.EXPECTED_FIELDS, *TIMESERIES_INDEX_FIELDS})
```
ghop02, line 46: It would be great to trim down the `TIMESERIES_INDEX_FIELDS` to only include fips and date - we don't need most of the fields and we can get aggregate level from the region object itself

ghop02 (approved):
🥳 looks good to me!

TomGoBravo:
[Remove unused DataSource constants](https://trello.com/c/uPQXBuDg/777-remove-unused-datasource-constants)

libs/datasets/data_source.py
TomGoBravo: Created https://trello.com/c/2Pa4DMyu/836-trim-down-the-timeseriesindexfields

PR #917, opened 2021-01-27
move test to tests, fixing conflict with stdlib

TomGoBravo:
This PR helps with https://trello.com/c/jWnCdNG9/822-merge-combined-data-dashboard and is recommended at https://stackoverflow.com/questions/65910240/pytest-breaks-with-pip-install#comment116548215_65910240


ghop02 (approved):
Not for now, but there are a decent amount of old files we probably don't need anymore

PR #918, opened 2021-01-27
Basic data pipeline dashboard

TomGoBravo:
This PR addresses https://trello.com/c/jWnCdNG9/822-merge-combined-data-dashboard

This adds a plotly dash UI for selecting a region and viewing the combined data of that region, with all annotations and provenance information.

Possible future work is described at https://docs.google.com/document/d/1zOrVC3VH0fRRAJVIrTr833qxZMuFfCql2n_nvz4JrUA/edit


dash_app.py
```diff
@@ -0,0 +1,100 @@
...
+external_stylesheets = ["https://codepen.io/chriddyp/pen/bWLwgP.css"]
+
+app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
+
+# Enable offline use.
+app.css.config.serve_locally = True
+app.scripts.config.serve_locally = True
+
+
+ds = combined_datasets.load_us_timeseries_dataset().get_subset(exclude_county_999=True)
```
ghop02, line 26: I realize this more of an example, but I'd like to include as many loading and app setup functions as possible in helpers functions - otherwise on any import (especially on imports in tests) it'll automatically trigger some expensive operations.  

Generally with flask apps, I have a `build_app` function that returns an instance of `app` and handles all of the initialization.  This makes it a bit tricky with the decorator here, even having all of the app building would be helpful to be in a function

requirements.txt
```diff
@@ -54,3 +52,24 @@ xlrd==1.2.0  # For panda's read_excel method
 numexpr==2.7.1
 
 backports.cached-property==1.0.0.post2
+
+
+
```
ghop02, line 57: Maybe add a comment saying these are for dash?

TomGoBravo: Done in cf896f83

dash_app.py
TomGoBravo: Good point. Moved the entire thing into a function. c48b5e49b9

TomGoBravo:
[Merge combined data dashboard](https://trello.com/c/jWnCdNG9/822-merge-combined-data-dashboard)

ghop02 (approved):
LGTM!

PR #919, opened 2021-01-27
Regional subsets in combined_datasets

TomGoBravo:
This PR addresses https://trello.com/c/VP9NRpJe/778 and implements pretty close to the notes in https://www.dropbox.com/scl/fi/wlchtiqvq0cs9aj5lnbfm/Importing-CAN-scraper-data.paper?dl=0&rlkey=zz76hxsiqgg4srkurcutt6pc5

With this change we can add things like `datasource_regions(NYTimesDataset, exclude=ALL_NYC_REGIONS)` or `datasource_regions(CDCTestingDataset, RegionMask(AggregationLevel.COUNTY, states=['TX', 'NY']), exclude=Region.from_fips("06037"))` to the big maps in combined_datasets.py.

* Adds `RegionMask` frozen dataclass, a function to expand it to location_id
* Adds `DataSourceAndRegionMasks`, a DataSource class and include/exclude region masks. Instances of this class can be used in the same places where the DataSource type are used
* Replaces use of the current `NYTimesDataset` class with a new `DataSourceAndRegionMasks` that excludes `ALL_NYC_REGIONS`. Removes the region filter that was in nytimes_dataset.py

## Tested


libs/datasets/combined_datasets.py
```diff
@@ -47,17 +53,82 @@ class RegionLatestNotFound(IndexError):
...
+    include: Optional[RegionMaskOrRegions] = None
+    exclude: Optional[RegionMaskOrRegions] = None
+
+    @property
+    def EXPECTED_FIELDS(self):
+        """Returns the same property of the wrapped DataSource class."""
+        return self.data_source_cls.EXPECTED_FIELDS
+
+    @property
+    def SOURCE_NAME(self):
```
ghop02, line 86: Generally the convention is that only class variables are caps.  Would you be opposed to be lowercase?

libs/datasets/combined_datasets.py
```diff
@@ -47,17 +53,82 @@ class RegionLatestNotFound(IndexError):
...
+        if self.include:
+            dataset = dataset.get_locations_subset(_get_location_ids(self.include))
+        if self.exclude:
+            dataset = dataset.remove_locations(_get_location_ids(self.exclude))
+        return dataset
+
+
+def datasource_regions(
+    data_source_cls: Type[data_source.DataSource],
+    include: Optional[RegionMaskOrRegions] = None,
```
ghop02, line 111: Thoughts on defaulting include to be the equivalent of everything?

libs/datasets/combined_datasets.py
```diff
@@ -47,17 +53,82 @@ class RegionLatestNotFound(IndexError):
...
+                return [r.location_id for r in region_mask_or_regions]
+
+        if self.include:
+            dataset = dataset.get_locations_subset(_get_location_ids(self.include))
+        if self.exclude:
+            dataset = dataset.remove_locations(_get_location_ids(self.exclude))
+        return dataset
+
+
+def datasource_regions(
```
ghop02, line 109: I'm not sure this provides a ton of value over just the class itself?

ghop02 (approved):
Looks good! I think it can be paired down, but overall a fan!

libs/datasets/combined_datasets.py
TomGoBravo: True, not much, but it does make use of this code look more like that in https://www.dropbox.com/scl/fi/wlchtiqvq0cs9aj5lnbfm/Importing-CAN-scraper-data.paper and enforce that callers explicitly list some intention about region.

libs/datasets/combined_datasets.py
TomGoBravo: I prefer it to be explicit, for example `include: Union[str,RegionMaskOrRegions] = "*"`, but even that doesn't seem to be needed at this point.

libs/datasets/combined_datasets.py
ghop02: Ahh I missed this - yeah I think that's alright for this case then - maybe make a comment saying that? This is pretty much implementing an interface

libs/datasets/combined_datasets.py
ghop02: ah yeah - that's what I was thinking!

libs/datasets/combined_datasets.py
ghop02: fine by me - easy enough to change if we desire at any point

TomGoBravo:
[CAN Scraper: Clean up / document mechanism for disabling/overriding data for a problematic (provider, region) pair.](https://trello.com/c/VP9NRpJe/778-can-scraper-clean-up-document-mechanism-for-disabling-overriding-data-for-a-problematic-provider-region-pair)

PR #922, opened 2021-02-01
Enable pagination for query results

ghop02:
good news, we have over 1k API users! The code was only equipped to handle the first page, so this paginates the response and will update with all active users.

Tested by running locally against the athena data that has more than 1000 rows

### DIFF
--- a/libs/update_api_user_metrics.py
+++ b/libs/update_api_user_metrics.py
@@ -90,8 +90,6 @@ def _run_query(database: str, query: str,) -> List[dict]:
     if response["Status"]["State"] != "SUCCEEDED":
         raise CloudWatchQueryError()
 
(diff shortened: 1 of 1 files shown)


libs/update_api_user_metrics.py
```diff
@@ -92,23 +92,34 @@ def _run_query(database: str, query: str,) -> List[dict]:
 
     response_query_result = client.get_query_results(QueryExecutionId=query_id)
```
TomGoBravo, line 93: What happens to response_query_result?

TomGoBravo (approved):
Please add something in PR description about how you tested this.

libs/update_api_user_metrics.py
ghop02: oo i can probably remove this

PR #924, opened 2021-02-02
add timeseries.derive_vaccine_pct and use it when producing the API output.

TomGoBravo:
This is part of https://trello.com/c/PgOjusc7/846-etl-tn-county-vaccine-data-into-api

I ran `make test` but don't have a great way to test API generation locally.


libs/metrics/top_level_metrics.py
```diff
@@ -98,8 +97,12 @@ def calculate_metrics_for_timeseries(
 
     icu_capacity_ratio = icu_capacity.calculate_icu_capacity(data)
 
-    vaccines_initiated_ratio = ratio_vaccinated.calculate_ratio_initiated(timeseries)
-    vaccines_completed_ratio = ratio_vaccinated.calculate_ratio_completed(timeseries)
+    vaccines_initiated_ratio = common_df.get_timeseries(
+        timeseries.date_indexed, CommonFields.VACCINATIONS_INITIATED_PCT, EMPTY_TS
+    )
+    vaccines_completed_ratio = common_df.get_timeseries(
+        timeseries.date_indexed, CommonFields.VACCINATIONS_COMPLETED_PCT, EMPTY_TS
```
ghop02, line 104: I should've asked this before, but are these fields truly percentages or are they actually ratios (22.3 or 0.223)? I would like to be specific so that we don't accidentally have confusion there.

ghop02: If they are actually ratios, I'd prefer to change the common field to say as much

libs/datasets/timeseries.py
```diff
@@ -966,9 +966,13 @@ def write_to_dataset_pointer(self, pointer: dataset_pointer.DatasetPointer):
 
     def drop_column_if_present(self, column: str) -> "MultiRegionDataset":
         """Drops the specified column from the timeseries if it exists"""
-        timeseries_df = self.timeseries.drop(column, axis="columns", errors="ignore")
-        static_df = self.static.drop(column, axis="columns", errors="ignore")
-        tag = self.tag[self.tag.index.get_level_values(PdFields.VARIABLE) != column]
+        return self.drop_columns_if_present([column])
+
+    def drop_columns_if_present(self, columns: List[str]) -> "MultiRegionDataset":
```
ghop02, line 971: should this be `List[CommonFields]`?

libs/metrics/top_level_metrics.py
ghop02: ah it looks like they are all real percentages.  Somewhere along the line we need to scale that value by 100

libs/datasets/timeseries.py
TomGoBravo: Fixed in 978e5e24

libs/metrics/top_level_metrics.py
TomGoBravo: Done

ghop02 (approved):

PR #928, opened 2021-02-03
agg vaccine percentages using population weights

TomGoBravo:
This PR addresses issue #<insert number>

### Please Check
- [ ] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/main/api/schemas) updated?
- [ ] Are tests passing?

PR #929, opened 2021-02-03
Data update with fixed vaccine aggregations

TomGoBravo:
This PR addresses https://trello.com/c/WYdvQgl4/865-metro-vaccinated-incorrect

Data update follow up to https://github.com/covid-projections/covid-data-model/pull/928 after failure in github action that I don't need (https://github.com/covid-projections/covid-data-public/runs/1825313387).


TomGoBravo:
Instead doing this: image (1)

PR #930, opened 2021-02-08
Add tag type to anomaly annotations

ghop02:
passes through the tag type to the annotations to help inform frontend about what sort of anomaly it is


tests/libs/api_v2_pipeline_test.py
```diff
@@ -167,7 +165,7 @@ def test_annotation(rt_dataset, icu_dataset):
 
     assert timeseries_for_region.annotations.deaths.sources == []
     assert timeseries_for_region.annotations.deaths.anomalies == [
-        AnomalyAnnotation(date="2020-04-01", original_observation=10.0)
+        AnomalyAnnotation(date="2020-04-01", original_observation=10.0, tag_type=tag.TAG_TYPE)
```
TomGoBravo, line 168: use the instance attribute `type` instead of class attribute

libs/build_api_v2.py
```diff
@@ -130,14 +130,16 @@ def _build_metric_annotations(
 
     anomalies = tag_series.annotations(field_name)
     anomalies = [
-        AnomalyAnnotation(date=t.date, original_observation=t.original_observation)
-        for t in anomalies
+        AnomalyAnnotation(
+            date=tag.date, original_observation=tag.original_observation, tag_type=tag.TAG_TYPE
```
TomGoBravo, line 134: use the instance attribute `type` instead of class attribute

api/docs/open_api_schema.json
```diff
@@ -1622,10 +1622,22 @@
...
+      "TagType": {
+        "title": "TagType",
+        "enum": [
+          "cumulative_tail_truncated",
+          "cumulative_long_tail_truncated",
+          "zscore_outlier",
+          "provenance"
+        ],
+        "type": "string",
+        "description": "Values that may appear in TagField.TYPE.\n\nCurrently 'annotation' refers to tag types that must have a specific real value for DATE. Other\ntags (currently only PROVENANCE, but may be expanded for things such as storing a processing\nstep name) do not have a date field. Putting the type of all tags in a single enum\nmakes it easy to represent all tags in the same structure but makes the concept of an\n'annotation type' less explicit in our code."
```
TomGoBravo, line 1634: This desc could be confusing in the API because all the stuff about PROVENANCE doesn't apply. Duplicating the enum for the API is kind of annoying. I'd be fine with editing the desc to match what we want in the API docs and leaving internal notes in a comment. Or dup the enum and use a unit test to check that they are in sync.

TomGoBravo (approved):

PR #933, opened 2021-02-08
Move annotations into wide-dates file

TomGoBravo:
This PR prepares for https://trello.com/c/lbiZpKwH/849-proposal-improve-the-provenance-link-so-that-we-can-show-a-source-urls-instead-of-a-data-class by moving all tags into the wide dates file. Before this change provenance was the only tag written to the wide-dates file. Reducing the number of files to represent a MultiRegionDataset will also simplify reading it directly from a git history.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -500,26 +500,24 @@ class MultiRegionDataset:
         wide_dates_df = pd.read_csv(pointer.path_wide_dates(), low_memory=False)
         wide_dates_df = wide_dates_df.set_index([CommonFields.LOCATION_ID, PdFields.VARIABLE])
 
-        tag_columns = pd.Series(False, index=wide_dates_df.columns)
+        # Iterate through all known tag types. The following are populated while iterating.
+        tag_columns_mask = pd.Series(False, index=wide_dates_df.columns)
         tag_df_to_concat = []
         for tag_type in TagType:
-            # Extract provenance columns from wide date DataFrame so all columns are dates. This
-            # parses the column names created in `timeseries_rows`.
-            provenance_column_mask = wide_dates_df.columns.str.match(
-                r"\A" + tag_type + r"(-\d+)?\Z"
-            )
-            if provenance_column_mask.any():
-                provenance_columns = wide_dates_df.loc[:, provenance_column_mask]
-                provenance_series = (
-                    provenance_columns.stack().reset_index(-1, drop=True).rename(TagField.CONTENT)
-                )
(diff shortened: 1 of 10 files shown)


libs/datasets/timeseries.py
```diff
@@ -725,22 +730,24 @@ def timeseries_rows(self) -> pd.DataFrame:
...
-                # for a single column. Extra columns are named "provenance-1", "provenance-2", ...
-                # These column names are parsed in `read_from_pointer`.
-                .rename(columns=lambda i: TagType.PROVENANCE + ("" if i == 0 else f"-{i}"))
-            )
-            return pd.concat([provenance_columns, wide_dates], axis=1)
-        else:
-            return wide_dates
+        # Each element of output_series will be a column in the returned DataFrame. There is at
+        # least one column for each tag type in self.tag. If a timeseries has multiple tags with
+        # the same type additional columns are added.
```
ghop02, line 735: Instead of having one column per annotation, could we just combine them into a single json list? 

thinking about it a bit more - i think this approach would be easier if we represented tags in the dataframe as actual `TagInTimeseries`  objects rather than json encoded strings.  Is there much of an advantage to keeping it as strings in the pandas objects?

ghop02 (approved):
This looks alright to me as it is - I do think we're overcomplicating some of the code by keeping the `content` as strings rather than annotation objects in the pandas series/dataframes.  

Seems like we could get away with a simple "annotations_json" column for each field that was a json serialized list of all annotations for that field.  Would be easier to do if provenance was also a json object - imagining it similar to: `[{"tag_type": "provenance", "value": "something"}, {"tag_type": "zscore_outlier": "original_observation": 10000", "date": "2020-12-12"}]`

But that's probably not for this PR - just something to think about as we continue forward with passing source info through.

libs/datasets/timeseries.py
TomGoBravo: Fair point. There is a tradeoff between trying to keep data in structures that are well supported by pandas vs customized to exactly what we want. Values in panda DataFrame cells can be operated on with native fast panda operations instead of us needing to write code. I like being able to do most of joining and making subsets of datasets with panda operations. Keeping the disk and memory representations pretty close simplifies the reading and writing code.

TomGoBravo: Another advantage of using separate columns of strings in CSVs on disk is that I currently depend pretty heavily on csvkit for debugging data issues, which will be less effective with heavier use of JSON in CSVs. I'm hoping standing up a dashboard will reduce how much time we spend handling CSVs and give us more flexibility to use other representations.

PR #934, opened 2021-02-09
read and transform parquet file during data update

TomGoBravo:
This PR will help with https://trello.com/c/lbiZpKwH/849-proposal-improve-the-provenance-link-so-that-we-can-show-a-source-urls-instead-of-a-data-class because it lets us skip adding source_url in covid-data-public CSVs. It also simplifies the process of adding new fields in general to combined_datasets.

* Adds a new (and hacky) `CanScraperBase` that caches the CAN parquet file after it has been loaded
* Changes `CANScraperStateProviders`, `CDCTestingDataset` and `CDCVaccinesDataset` to depend on their scripts in the covid-data-public repo. After the scripts are turned off we can move the source to this repo
* Adds a pip dependency on python-snappy to fix  ..../fastparquet/compression.py line 144, in decompress_data RuntimeError: Decompression 'SNAPPY' not available. 

## Tested

In `main` ran `time ./run.py data update` and got
```
real	14m53.280s
user	10m39.442s
sys	3m46.207s
```

then checked out this branch, ran the same and got
```
real	9m43.823s
user	7m48.892s
sys	1m39.757s
```

with the same output in `data` except for the covid-data-model repo state.


libs/datasets/data_source.py
```diff
@@ -36,13 +38,17 @@ class DataSource(object):
     IGNORED_FIELDS = (CommonFields.COUNTY, CommonFields.COUNTRY, CommonFields.STATE)
 
     @classmethod
-    @lru_cache(None)
-    def make_dataset(cls) -> timeseries.MultiRegionDataset:
-        """Default implementation of make_dataset that loads timeseries data from a CSV."""
+    def _get_data(cls) -> pd.DataFrame:
```
ghop02, line 41: load data?

libs/datasets/sources/can_scraper_state_providers.py
```diff
@@ -1,11 +1,13 @@
 from covidactnow.datapublic.common_fields import CommonFields
+from scripts import update_can_scraper_state_providers
```
ghop02, line 2: I'm a bit hesitant to import the `scripts` package directly.  I don't think it's clear that it comes from `covid-data-public` and probably more important is that I don't think it's obvious that if you change code in covid-data-public/scripts you'll potentially be breaking upstream code in covid-data-model.  

I'd be happy to meet in the middle somewhere - maybe we can provide some base class that provides a well defined transform interface that we lean on a bit more?  Thoughts on that concern?

libs/datasets/data_source.py
```diff
@@ -63,3 +69,23 @@ def make_dataset(cls) -> timeseries.MultiRegionDataset:
...
+# TODO(tom): Clean up the mess that is subclasses of DataSource and
+#  instances of DataSourceAndRegionMasks
+class CanScraperBase(DataSource):
+    # The method called to transform the DataFrame returned by CovidCountyDataset into what is
+    # consumed by DataSource.make_dataset. Must be set in subclasses.
+    TRANSFORM_METHOD: Callable[[pd.DataFrame], pd.DataFrame]
+
+    @staticmethod
+    @lru_cache(None)
+    def _get_covid_county_dataset() -> ccd_helpers.CovidCountyDataset:
```
ghop02, line 84: Ah - this is probably a todo, but we should rename this to be `CanScraperDataset` or something similar - `CovidCountyDataset` name is a bit outdated

libs/datasets/sources/can_scraper_state_providers.py
ghop02: Also I think the only reason that we can even see the scripts directory is an artifact of the way we install the package.  If it were a proper pip package (or we installed it another way than referring to the path of covid-data-public directly) i don't think it would work

TomGoBravo: We want to move ccd_helpers from covid-data-public to covid-data-model. I don't see much value in refactoring to support sharing this code between the repos because I don't see anything blocking the move. I see merge this today and move ccd_helpers tomorrow.

libs/datasets/data_source.py
TomGoBravo: 'dataset' is pretty overloaded. Added TODO with suggestion  `CanScraperLoader`

ghop02 (approved):
Fine with this approach - I think it'd be helpful to note above the `scripts` imports that they are in `covid-data-public` and will be changed shortly

TomGoBravo:
> Fine with this approach - I think it'd be helpful to note above the `scripts` imports that they are in `covid-data-public` and will be changed shortly

Cool. This gives me the chance to remove 3 more TODOs soon :-)

PR #935, opened 2021-02-09
Add source_url tag

TomGoBravo:
This PR is a big part of https://trello.com/c/lbiZpKwH/849-proposal-improve-the-provenance-link-so-that-we-can-show-a-source-urls-instead-of-a-data-class

* Adds `class SourceUrl(TagInTimeseries)` and an property to get them in `OneRegionTimeseriesDataset`
* Changes `test_helper.TimeseriesLiteral` to a frozen dataclass with some type conversion in `__init__`
* Adds `dataclass_with_default_init`, backport of code in Python 3.9

Instead of doing type conversion in `test_helper.TimeseriesLiteral` i could leave the attributes as one of 3 types and added properties provenance_as_list and source_url_as_list. Either is fine but I had to pick one.

### DIFF
--- a/libs/datasets/taglib.py
+++ b/libs/datasets/taglib.py
@@ -98,22 +98,28 @@ class ProvenanceTag(TagInTimeseries):
 
     @classmethod
     def make_instance(cls, *, content: str) -> "TagInTimeseries":
-        return ProvenanceTag(source=content)
+        return cls(source=content)
 
     @property
     def content(self) -> str:
         return self.source
 
 
+class UrlStr(str):
+    """"""
+
+    pass
+
+
 @dataclass(frozen=True)
 class SourceUrl(TagInTimeseries):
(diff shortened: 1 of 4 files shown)


ghop02 (commented):
Seems alright for now - I was however initially imagining this as a property on the provenance annotation, i.e. `{"data_source": "CANScraper", "url": "<source url>"}`.  I think the source url does fall under the scope of the provenance of that timeseries.  What do you think?

TomGoBravo:
> Seems alright for now - I was however initially imagining this as a property on the provenance annotation, i.e. `{"data_source": "CANScraper", "url": "<source url>"}`. I think the source url does fall under the scope of the provenance of that timeseries. What do you think?

Good idea. Added as a TODO.

ghop02 (approved):

PR #936, opened 2021-02-10
Switch to can scraper usa facts

ghop02:


ghop02:
I ran the update and verified that case data for 36061 didn't change - hopefully it does after the scraper successfully runs

TomGoBravo (approved):

PR #937, opened 2021-02-10
Move parquet transforms to covid-data-model

TomGoBravo:

It moves the files that are run in the this repo and will be modified to load source_url, into this repo.

* Move parquet transforms from covid-data-public repo to covid-data-model without changing the structure or names much
* Import the copied module as ccd_helpers to avoid changing the transform functions at all
* rename CovidCountyDataset to CanScraperLoader

## Tested

Deleted the data from my local covid-data-public repo
```
	deleted:    data/can-scrapers-state-providers/timeseries-common.csv
	deleted:    data/testing-cdc/timeseries-common.csv
	deleted:    data/vaccines-cdc/timeseries-common.csv
	deleted:    scripts/ccd_helpers.py
	deleted:    scripts/update_can_scraper_state_providers.py
	deleted:    scripts/update_cdc_test_data.py
	deleted:    scripts/update_cdc_vaccine_data.py
```
then ran `./run.py data update` and checked that identical data was produced.

### DIFF
--- a/libs/datasets/data_source.py
+++ b/libs/datasets/data_source.py
@@ -75,15 +75,15 @@ class DataSource(object):
 # TODO(tom): Clean up the mess that is subclasses of DataSource and
 #  instances of DataSourceAndRegionMasks
 class CanScraperBase(DataSource):
-    # The method called to transform the DataFrame returned by CovidCountyDataset into what is
+    # The method called to transform the DataFrame returned by CanScraperLoader into what is
     # consumed by DataSource.make_dataset. Must be set in subclasses.
     TRANSFORM_METHOD: Callable[[pd.DataFrame], pd.DataFrame]
 
     @staticmethod
     @lru_cache(None)
-    def _get_covid_county_dataset() -> ccd_helpers.CovidCountyDataset:
-        # TODO(tom): Rename CovidCountyDataset to CanScraperLoader or the something like that.
-        return ccd_helpers.CovidCountyDataset.load()
+    def _get_covid_county_dataset() -> ccd_helpers.CanScraperLoader:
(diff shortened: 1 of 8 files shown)


libs/datasets/data_source.py
```diff
@@ -7,7 +7,7 @@
 import structlog
 from covidactnow.datapublic import common_df
 from covidactnow.datapublic.common_fields import CommonFields
-from scripts import ccd_helpers
+from libs.datasets.sources import can_scraper_helpers as ccd_helpers
```
ghop02, line 10: Are you waiting to push the can_scraper_helpers name down? Seems we can just use that everywhere

libs/datasets/sources/can_scraper_state_providers.py
```diff
@@ -1,15 +1,149 @@
+from libs.datasets import data_source
 from covidactnow.datapublic.common_fields import CommonFields
+from libs.datasets.sources import can_scraper_helpers as ccd_helpers
 
-# TODO(tom): Remove this really ugly import from the covid-data-public repo.
-from scripts import update_can_scraper_state_providers
 
-from libs.datasets import data_source
+def transform(dataset: ccd_helpers.CanScraperLoader):
```
ghop02, line 6: 🥳 this is exciting!!

ghop02 (approved):

libs/datasets/data_source.py
TomGoBravo: Yes, I'm waiting for something that will take less screen space than hundreds of `can_scraper_helpers.ScraperVariable` with all those keyword args. Also trying to preserve the existing code during the transition in case there are merge conflicts to resolve etc.

PR #939, opened 2021-02-11
Move variables to a CanScraperBase attribute

TomGoBravo:

replace CanScraperBase.TRANSFORM_METHOD with VARIABLES and a method that can be overridden. This will make it easier to have query_multiple_variables return source_url values that can be added to the MultiRegionDataset.

## Tested

Ran `data update` and checked that identical output was produced.

### DIFF
--- a/libs/datasets/sources/can_scraper_state_providers.py
+++ b/libs/datasets/sources/can_scraper_state_providers.py
@@ -3,8 +3,28 @@ from covidactnow.datapublic.common_fields import CommonFields
 from libs.datasets.sources import can_scraper_helpers as ccd_helpers
 
 
-def transform(dataset: ccd_helpers.CanScraperLoader):
-    variables = [
+class CANScraperStateProviders(data_source.CanScraperBase):
+    SOURCE_NAME = "CANScrapersStateProviders"
+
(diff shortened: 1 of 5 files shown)


libs/datasets/sources/can_scraper_state_providers.py
```diff
@@ -3,8 +3,28 @@
 from libs.datasets.sources import can_scraper_helpers as ccd_helpers
 
 
-def transform(dataset: ccd_helpers.CanScraperLoader):
-    variables = [
+class CANScraperStateProviders(data_source.CanScraperBase):
+    SOURCE_NAME = "CANScrapersStateProviders"
+
+    EXPECTED_FIELDS = [
```
ghop02, line 9: In the future, for Can scrapers, could we just generate this from `VARIABLES`?

ghop02 (approved):

libs/datasets/sources/can_scraper_state_providers.py
TomGoBravo: Yes, later. Maybe hard to believe :-P but I'm trying to stick to the min refactoring to get the current task done.

PR #941, opened 2021-02-11
Copy source_url to combined_dataset

TomGoBravo:

* Changes `CanScraperLoader.query_multiple_variables` to return a tuple of data and source_urls, instead of just one DataFrame
* Changes `CanScraperBase.make_dataset` to copy the most recent source_url of each timeseries to the MultiRegionDataset it returns
* Changes `build_can_scraper_dataframe` to support returning different URLs for each observation to help verify that above is working correctly
* Adds `MultiRegionDataset.append_fips_tag_df`, tiny helped to do fips->location_id transformation.
* Reverts "fix zscore_outlier data in generate_api_v2_test (#938)". Apparently it isn't very stable. Longer term fix is beyond scope of current work.

## Tested

Ran data update and saw a bunch of source_url get added. See https://covidactnow.slack.com/archives/C015V4JS952/p1613073818001400


ghop02 (approved):
Lgtm!

PR #942, opened 2021-02-12
source_url in API

TomGoBravo:

* Adds ￼source_url to API FieldAnnotations
* Adds annotations for vaccines which were in the API but never populated.
* Adds a tests to make sure all FieldAnnotations are populated

## Tested

Changed to branch `git checkout tgb-849-add-source-url`, ran `./run.py data update` then switched back to this branch with updated data. Then ran `python pyseir/cli.py build-all --generate-api-v2 -s FL --fips 12011` and checked with `jq . output/county/12011.timeseries.json` that some URLs were added.

### DIFF
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -194,6 +194,7 @@ class FieldAnnotations(base_model.APIBaseModel):
 
     sources: List[FieldSource]
     anomalies: List[AnomalyAnnotation]
+    source_url: str = pydantic.Field(None)
 
 
 class Annotations(base_model.APIBaseModel):
--- a/libs/build_api_v2.py
+++ b/libs/build_api_v2.py
@@ -27,7 +27,7 @@ from libs.datasets.timeseries import OneRegionTimeseriesDataset
 
 
 METRIC_SOURCES_NOT_FOUND_MESSAGE = "Unable to find provenance in FieldSource enum"
-
+METRIC_MULTIPLE_SOURCE_URLS_MESSAGE = "More than one source_url for a field"
 
 USA_VACCINATION_START_DATE = datetime(2020, 12, 14)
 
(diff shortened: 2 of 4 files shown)


ghop02 (approved):
This looks alright for now! I think we'll get value out of combining the sources and source_url into a single annotation, but maybe that'll be for another time. 

I think provenance is intended to show the source where it came from and right now it's just surfacing the `DataSource` that it comes from. An idea i was thinking of was:
```
sources: [{
  "name": "New York Times",
  "url": <nytimes url>,
  # this field maybe isn't needed - or maybe its more useful in combined datasets but not the API
  "data_source": "NYTimesDataSource"
}]
```

I guess with this in mind - do you think its worth publishing this in the API right now or holding off until we refactor that part a bit?

TomGoBravo:
> This looks alright for now! I think we'll get value out of combining the sources and source_url into a single annotation, but maybe that'll be for another time.
> 
> I think provenance is intended to show the source where it came from and right now it's just surfacing the `DataSource` that it comes from. An idea i was thinking of was:
> 
> ```
> sources: [{
>   "name": "New York Times",
>   "url": <nytimes url>,
>   # this field maybe isn't needed - or maybe its more useful in combined datasets but not the API
>   "data_source": "NYTimesDataSource"
> }]
> ```
> 
> I guess with this in mind - do you think its worth publishing this in the API right now or holding off until we refactor that part a bit?

That's a reasonable idea. Are you thinking we can change the type of elements in `sources` (from string to an object) now because it doesn't have any big consumers yet?

ghop02:
> That's a reasonable idea. Are you thinking we can change the type of elements in `sources` (from string to an object) now because it doesn't have any big consumers yet?

I tend to think it's alright - I don't think too many people are using it and it's not yet used on the website anywhere.  alternatively we could create a new field and deprecate this one

PR #944, opened 2021-02-17
API with source and source_url combined

TomGoBravo:

* Changes API output to merge source and source_url. `source` is now an object instead of a string taken from an enum.

before
```
{
  "sources": ["CANScrapersStateProviders"], 
  "anomalies": [],
  "source_url": "https://www.dshs.state.tx.us/coronavirus/immunize/vaccine.aspx"
}
```

after
```
{
  "sources": [{
    "type": "CANScrapersStateProviders",
    "url": "https://www.dshs.state.tx.us/coronavirus/immunize/vaccine.aspx",
    "name": null
  }], 
  "anomalies": []
}
```

## Tested

Looked at 
```
python pyseir/cli.py build-all --generate-api-v2 -s TX --fips 48507
jq .  < output/county/48507.timeseries.json | less
```


api/can_api_v2_definition.py
```diff
@@ -189,12 +189,21 @@ class AnomalyAnnotation(base_model.APIBaseModel):
     )
 
 
+class FieldSource(base_model.APIBaseModel):
+    type: Optional[FieldSourceType] = pydantic.Field(
+        None, description="The type of data source from a CAN list of data source types"
```
ghop02, line 194: Should one of these be required? It seems incorrect that a field source annotation would have all 3 None.

api/can_api_v2_definition.py
```diff
@@ -189,12 +189,21 @@ class AnomalyAnnotation(base_model.APIBaseModel):
...
+    )
+    url: Optional[str] = pydantic.Field(
+        None, description="URL of a webpage containing the data at the source"
+    )
+    name: Optional[str] = pydantic.Field(None, description="A human readable name of the source")
```
ghop02, line 199: Thoughts on having the fields be required, but the values be optional?
i.e. `name: Optional[str] = pydantic.Field(...)`.  This would mean that all json fields would have `{"type": <something>, "name":  null}` rather than `{"type": <something>}`.  

I think this might be slightly more clear - even if for now all names are null.  However, I don't feel super strongly

ghop02 (commented):
Couple small questions but looking good!

api/can_api_v2_definition.py
TomGoBravo: Agreed that I'd expect at least one to be set but any of them can be None. I'm not sure what to do.

api/can_api_v2_definition.py
TomGoBravo: Following the principal of being strict in what you produce and lenient with what you consume I'd expect consumers to behave the same if a field is missing or null, so why pass around null? I don't think we want consumers to crash if a field is missing.

TomGoBravo: Looking at 
```
python pyseir/cli.py build-all --generate-api-v2 -s TX --fips 48507
jq .  < output/county/48507.timeseries.json | less
```
with `name: Optional[str] = pydantic.Field(None, desc..)` I see 

````
    "vaccinationsInitiated": {
      "sources": [
        {
          "type": "CANScrapersStateProviders",
          "url": "https://www.dshs.state.tx.us/coronavirus/immunize/vaccine.aspx",
          "name": null
        }
      ],
```

Maybe the problem is that the way we test API output makes it hard to see what is actually produced. :-/

PR #945, opened 2021-02-17
data pipeline dashboard production configuration

TomGoBravo:
All being well, this PR finishes https://trello.com/c/s0ObL5ii/860-deploy-dashapppy-the-data-pipeline-dashboard

This change adds configurations for uwsgi, nginx and webhook running at https://data-pipeline-dashboard.covidactnow.org/ so that it updates when changes are committed to `main`.


dash_app.py
```diff
@@ -102,6 +102,8 @@ def update_figure(selected_rows):
     return app
 
 
+app = init_dashboard()
```
TomGoBravo, line 105: @sglyon I moved `app` from a module-level variable to something only created when in __main__ to avoid creating it if this module is imported. That doesn't seem to play well with the way uwsgi wants to run it. What do you suggest?

services/data-pipeline-dashboard/uwsgi.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=uWSGI instance to serve data-pipeline-dashboard
+After=network.target
+
+[Service]
+User=tom
```
ghop02, line 6: Should we have a more generic user name?

services/webhook/secret.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+Environment="PULL_COVID_DATA_MODEL_SECRET=PUT_SECRET_HERE"
```
ghop02, line 2: How does this get modified?

dash_app.py
ghop02: on other projects I've separated out the code that creates the app from the code that imports/runs it (and making it clear that importing that run file will cause it to execute) - maybe that could work here?

services/webhook/secret.conf
TomGoBravo: Manually. I mention this in the README.

dash_app.py
TomGoBravo: You mean something like moving all the stuff in functions to another module (that can be imported without running the app) and leaving `dash_app` as a module that has the `app` as a module level variable so importing it runs the code?

services/data-pipeline-dashboard/uwsgi.service
TomGoBravo: I've banged my head enough to get it this far. Not sure it is worth more effort to get a machine account.

dash_app.py
ghop02: yeah that's what I was thinking!

TomGoBravo: Can you recommend a structure for organizing this flask app in our repo?

TomGoBravo: Addressed in https://github.com/covid-projections/covid-data-model/pull/948

ghop02 (approved):
lgtm!

PR #946, opened 2021-02-17
Add tail testing truncation

ghop02:
Wanted to run initial approach by @TomGoBravo before putting finishing touches/tests on


libs/datasets/timeseries.py
```diff
@@ -996,6 +1001,60 @@ def drop_new_case_outliers(
...
+    """Identifies and drops outliers from the new case series.
+
+    Args:
+        timeseries: Timeseries.
+        zscore_threshold: Z-score threshold.  All new cases with a zscore greater than the
+            threshold will be removed.
+
+    Returns: timeseries with outliers removed from test_positivity_7d
+    """
+    df_copy = timeseries.timeseries.copy()
```
TomGoBravo, line 1018: Can you use timeseries.timeseries_wide_dates() instead of df_copy and grouped_df? Something like
```
test_pos = ts_wide_dates.loc[:, CommonFields.TEST_POSITIVITY_7D]
zscores = test_pos.apply(process_series, axis=1)
test_positivity_diffs = test_pos.diff(axis=1).abs()
```
?

libs/datasets/timeseries.py
```diff
@@ -996,6 +1001,60 @@ def drop_new_case_outliers(
     return new_timeseries
 
 
+def drop_tail_positivity_outliers(
+    timeseries: MultiRegionDataset,
```
TomGoBravo, line 1005: timeseries is overloaded (also a module name and the object may have static data). Instead dataset or ds or ds_input or ...

libs/datasets/timeseries.py
```diff
@@ -1001,6 +1006,57 @@ def drop_new_case_outliers(
...
 
+def drop_tail_positivity_outliers(
+    dataset: MultiRegionDataset,
+    zscore_threshold: float = 10.0,
+    diff_threshold_ratio: float = 0.015,
+) -> MultiRegionDataset:
+    """Identifies and drops outliers from the new case series.
+
+    Args:
+        dataset: Timeseries.
```
TomGoBravo, line 1017: This line doesn't help the reader and seems unnecessary. What's the downside of removing it?

tests/libs/datasets/timeseries_test.py
```diff
@@ -1612,3 +1612,28 @@ def test_write_read_dataset_pointer_with_source_url(tmpdir):
     assert source_url_read[CommonFields.ICU_BEDS] == [url_str1]
     # Copy to a set because the order of the URLs in the source_url may change.
     assert set(source_url_read[CommonFields.CASES]) == {url_str2, url_str3}
+
+
+@pytest.mark.parametrize("last_value,is_outlier", [(0.02, False), (0.045, True)])
+def test_remove_test_positivity_outliers(last_value, is_outlier):
+    values = [0.015] * 7 + [last_value]
+    dataset = test_helpers.build_default_region_dataset({CommonFields.TEST_POSITIVITY_7D: values})
+    dataset = timeseries.drop_tail_positivity_outliers(dataset)
```
TomGoBravo, line 1621: call these dataset_in and dataset_out to avoid reusing the name. Then in the else block `test_helpers.assert_dataset_like(dataset_in, dataset_out)`

libs/datasets/timeseries.py
```diff
@@ -1001,6 +1006,57 @@ def drop_new_case_outliers(
...
+        zscore_threshold: Z-score threshold.  All new cases with a zscore greater than the
+            threshold will be removed.
+        diff_threshold_ratio: Minimum difference required for value to be outlier.
+
+    Returns: timeseries with outliers removed from test_positivity_7d
+    """
+    df_copy = dataset.timeseries.copy()
```
TomGoBravo, line 1024: Add TODO about being more consistent in how this accesses the data so readers know to not copy this pattern.

tests/libs/datasets/timeseries_test.py
```diff
@@ -1612,3 +1612,28 @@ def test_write_read_dataset_pointer_with_source_url(tmpdir):
     assert source_url_read[CommonFields.ICU_BEDS] == [url_str1]
     # Copy to a set because the order of the URLs in the source_url may change.
     assert set(source_url_read[CommonFields.CASES]) == {url_str2, url_str3}
+
+
+@pytest.mark.parametrize("last_value,is_outlier", [(0.02, False), (0.045, True)])
```
TomGoBravo, line 1617: Add TODO about making this test stronger. The code it tests has at least 3 cases of zscore or diff being over or under a threshold.

TomGoBravo (approved):
Add something in PR desc about how this was tested.

PR #957, opened 2021-02-19
Deploy api on serverless

ghop02:
This takes advantage of github environments to do an automated deploy of dev environment -> run integration tests -> prod.

Right now I set the prod deploy to require my manual approval, but we can modify to add others as well - just like the idea that  we're not accidentally deploy to prod right now


mikelehen (commented):
Yay!  Though I'm a bit curious about the way this is all structured.  Might be worth a chat on Monday.

1. There seems to be no way to just deploy to dev?  It seems like the point of a dev environment is that you can deploy to it separately from prod and verify everything works as expected.
2. We are duplicating all of the deploy steps separate for dev vs prod, which means that they can get out-of-sync and the dev environment is no longer a safe way to test prod, since they're potentially doing different steps.

I think I'd prefer a single workflow that you can run against either dev or prod, kinda' like the way the alert emails work (single workflow that you can target at different environments).  https://github.com/covid-projections/covid-projections/blob/7979f5dfee1c20802cdc66afd861459a5faf6e77/.github/workflows/vaccination-alerts.yml#L33

ghop02:
> Yay! Though I'm a bit curious about the way this is all structured. Might be worth a chat on Monday.
> 
> 1. There seems to be no way to just deploy to dev?  It seems like the point of a dev environment is that you can deploy to it separately from prod and verify everything works as expected.
> 2. We are duplicating all of the deploy steps separate for dev vs prod, which means that they can get out-of-sync and the dev environment is no longer a safe way to test prod, since they're potentially doing different steps.
> 
> I think I'd prefer a single workflow that you can run against either dev or prod, kinda' like the way the alert emails work (single workflow that you can target at different environments). https://github.com/covid-projections/covid-projections/blob/7979f5dfee1c20802cdc66afd861459a5faf6e77/.github/workflows/vaccination-alerts.yml#L33

Sounds good! Yeah I think part of the confusion here is that dev is actually more like a staging environment here

Deploying to production actually requires manual approval right now, so it stops before actually deploying to prod - but I do somewhat agree about that process.

I do like the idea of more automated CI/CD process (deploy to dev, run tests, tests pass -> deploy to prod automatically) but definitely agree about the prospects of the two environments diverging (and agree it should look different).

mikelehen:
I see...  So if you were working on adding a new feature and wanted to test it (manually), you'd just kick off a deploy and just not it, so it would get deployed to dev (where you can test it) but not to prod?  That's a bit different than any staging / prod workflow I've dealt with in the past, but I also don't care too strongly and this will be your thing to maintain 99% of the time, so I'm okay deferring. 😁

ghop02:
haha I agree that it's not a great workflow (or any that i've really worked on) - maybe creating an explicit staging environment that we then run the tests on would be the better solution.  

I do think a "purely dev" environment is handy because of the difficulty of really testing locally.

.github/workflows/deploy_aws_api_infrastructure_dev.yml
```diff
@@ -1,17 +1,93 @@
...
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        CLOUDFRONT_DISTRIBUTION_ID: ${{ secrets.CLOUDFRONT_DISTRIBUTION_ID }}
+        DOTENV: ${{ secrets.DOTENV }}
+        AWS_REGION: us-east-1
+      run: tools/deploy-api-infrastructure.sh dev
+
+  TestDev:
+    if: github.event.inputs.deploy_to_prod == 'true'
```
ghop02, line 48: TODO: I think i need to update this to only check if this is from a worfklow event

mikelehen: Why not just always test dev? 🤷

.github/workflows/deploy_aws_api_infrastructure_dev.yml
```diff
@@ -1,17 +1,93 @@
...
+    needs: [DeployDev]
```
mikelehen, line 49: FWIW- If we just always ran TestDev, you could combine TestDev and DeployDev jobs and speed up the github actions runtime (avoid checking out / installing python redundantly, etc.).

.github/workflows/deploy_aws_api_infrastructure_dev.yml
```diff
@@ -1,17 +1,93 @@
...
+
     runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python ${{ matrix.python-version }}
```
mikelehen, line 55: Looks like we're always installing 3.7.

api/awsauth/README.md
```diff
@@ -13,46 +13,28 @@ npm install -g serverless
 
 ## Deploy
 
-Currently we have a `dev` and `prod` stage.
+Currently we have a `dev` and `prod` stage deployed via Github Actions.
```
mikelehen, line 16: "stage" ?  Is that a common term?  Maybe "environment"?

api/awsauth/README.md
```diff
@@ -13,46 +13,28 @@ npm install -g serverless
...
 
-Deploys are set up to look for environment variables in a `.env` file.
+A .env file is used to configure most of the environment variables for deploy.
```
mikelehen, line 18: I don't quite understand how 'prod' and 'dev' end up doing different things.  Does the generated .env file have separate configurations for them?  What specifically is different to differentiate prod and env?  Mostly the CLOUDFRONT_DISTRIBUTION_ID?

If it's easy to add a bit more context about how the two environments work, that'd be nice, because it's a bit opaque just looking at the breadcrumbs I can see in this PR at least.

.github/workflows/deploy_aws_api_infrastructure_dev.yml
```diff
@@ -1,17 +1,93 @@
...
+        python-version: 3.7
+    - name: Install Serverless Framework and NPM dependencies
+      run: |
+        npm install -g serverless
+        npm install
+    - name: Deploy
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        CLOUDFRONT_DISTRIBUTION_ID: ${{ secrets.CLOUDFRONT_DISTRIBUTION_ID }}
```
mikelehen, line 90: How does it work that dev and prod use the same cloudfront ID?  I guess it's served from the same CDN but has a different URL or something?  Is that documented somewhere? 

FWIW- I'm not sure if the cloudfront distribution ID needs to be considered a secret.  If it's always the same, you could probably just hardcode it in the deploy script. 🤷

api/awsauth/README.md
```diff
@@ -13,46 +13,28 @@ npm install -g serverless
...
+date variables.
 
-# Cloudfront distribution ID number for data deploy
-CLOUDFRONT_DISTRIBUTION_ID=
+## Deploying code to production
 
-# AWS Kinesis Firehose table name for metric collections
-FIREHOSE_TABLE_NAME=
+When code in api/awsauth is merged, the deploy process is automatically kicked off.
+It will deploy to dev and then run tests.  One the tests succeed, the deploy waits for a
```
mikelehen, line 28: nit: One => Once

api/awsauth/README.md
```diff
@@ -13,46 +13,28 @@ npm install -g serverless
...
+manual approval to deploy. This is limited right now to make sure that someone is around to
```
mikelehen, line 29: "to deploy" => "to deploy to prod"

api/awsauth/README.md
```diff
@@ -13,46 +13,28 @@ npm install -g serverless
...
-object with the contents of the .env file.  Refer to the class for the most up to
-date variables.
+If you are making changes that require the .env file to be chnaged, follow these steps:
 
-```
-sls deploy --stage {dev,prod}
-```
+ 1. Request access to the [API .env content doc](https://docs.google.com/document/d/1lHD2cG0FYKoOc5Q17ugg20m-WcucD7UfkiTSChpmUhQ)
+ 2. Update value for dev and prod
+ 3. Update Github action secrets
```
mikelehen, line 39: Is there more than one or is it just DOTENV?

mikelehen (approved):
Few minor questions / comments, but looks good!  Thanks for getting this automated.

api/awsauth/README.md
ghop02: ah stage was the term used in the serverless framework - can change this to environment

.github/workflows/deploy_aws_api_infrastructure_dev.yml
ghop02: This takes advantage of environment level secrets - so there are different values per secret

api/awsauth/README.md
mikelehen: Feel free to leave it if it's their terminology!

PR #959, opened 2021-02-21
use loc to set aggregate_level

ghop02:
I'm not 100% sure why, but I think the change https://github.com/covid-projections/covid-data-public/pull/208/files caused https://sentry.io/organizations/covidactnow/issues/2229338105/?project=5203044&query=is%3Aunresolved.  

changing the way we set location_id here seems to do the trick - not sure if we'll see more of these pop up or not


TomGoBravo (approved):

TomGoBravo:
addressed by https://github.com/covid-projections/covid-data-public/pull/210 and https://github.com/covid-projections/covid-data-model/pull/960

PR #961, opened 2021-02-22
refactoring for adding source tag

TomGoBravo:
This PR is prep for https://trello.com/c/hSJN82y9/927-pipe-through-source-names-to-fe

* Extracts API code that makes `FieldSource` so it can be more easily replaced
* Renames ` TagInTimeseries` field `type` to `tag_type` so that `Source` can have `type` property.
* Adds `OneRegionTimeseriesDataset.tag_objects_series` for easier access to these, by type

### DIFF
--- a/libs/build_api_v2.py
+++ b/libs/build_api_v2.py
@@ -23,7 +23,6 @@ from api.can_api_v2_definition import AnomalyAnnotation
 from api.can_api_v2_definition import FieldSource
 from api.can_api_v2_definition import FieldSourceType
 from libs.datasets import timeseries
-from libs.datasets.taglib import TagType
 from libs.datasets.tail_filter import TagField
 from libs.datasets.timeseries import OneRegionTimeseriesDataset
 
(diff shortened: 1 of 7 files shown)


libs/build_api_v2.py
```diff
@@ -129,30 +129,33 @@ def _build_metric_annotations(
...
+        for tag in anomalies
+    ]
+
+    if not sources and not anomalies:
+        return None
+
+    return FieldAnnotations(sources=sources, anomalies=anomalies)
+
+
+def _sources_from_provenance_and_source_url(field_name, log, tag_series):
```
ghop02, line 148: types? also log feels like it would go better at the end of the args list

libs/datasets/timeseries.py
```diff
@@ -109,10 +109,17 @@ def source_url(self) -> Mapping[CommonFields, List[UrlStr]]:
...
 
     def annotations(self, metric: FieldName) -> List[taglib.AnnotationWithDate]:
-        return_value = []
-        for _, row in self.tag.loc[[metric], ANNOTATION_TAG_TYPES].reset_index().iterrows():
-            return_value.append(taglib.AnnotationWithDate.make(row.tag_type, content=row.content))
-        return return_value
+        return self.tag_objects_df.loc[[metric], ANNOTATION_TAG_TYPES].to_list()
+
+    @cached_property
+    def tag_objects_df(self) -> pd.Series:
```
ghop02, line 115: is this being returned as a dataframe or a series? Type seems inconsistent with the name

ghop02 (approved):
Couple small comments but looks good!

PR #964, opened 2021-02-23
attempt at fixing another KeyError

TomGoBravo:
Replace direct access to `OneRegionTimeseriesDataset.tag_objects_series` with a `sources(field)` method. Added tests in timeseries_test.py but not the api code.

https://sentry.io/organizations/covidactnow/issues/2232048499/?referrer=slack

libs/pipelines/api_v2_pipeline.py in build_timeseries_for_region at line 160

location_id | iso1:us#iso2:us-ny#fips:3651000

`KeyError: "None of [Index(['source'], dtype='object')] are in the [columns]"`

https://covidactnow.slack.com/archives/C011Z21ST8V/p1614040552007500

### DIFF
--- a/libs/build_api_v2.py
+++ b/libs/build_api_v2.py
@@ -24,7 +24,6 @@ from api.can_api_v2_definition import AnomalyAnnotation
 from api.can_api_v2_definition import FieldSource
 from api.can_api_v2_definition import FieldSourceType
 from libs.datasets import timeseries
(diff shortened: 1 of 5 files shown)


ghop02 (approved):
LGTM - I added back in the current icu field - which I think is why that test was failing, but not sure if you removed it for a specific reason or not

PR #965, opened 2021-02-23
Update schemas with source object

ghop02:
Just running schema update to get the new source object in

```
./run.py api update-schemas
```


TomGoBravo (approved):
Please put command or explicit link to doc used to do this in the PR desc.

PR #966, opened 2021-02-23
speed up CanScraperLoader._get_rows

TomGoBravo:
This PR is part of https://trello.com/c/hSJN82y9/927-pipe-through-source-names-to-fe

I got tired of waiting for tests to run on all the data and have seen this is a big chunk of combined dataset update too.

Benchmark taken with 

```
py-spy record --subprocesses -r 10 -o output/`date --iso-8601=minutes --utc`.svg -- pytest tests/libs/datasets/data_source_test.py
```

then looking at the count of samples in changed code.


ghop02 (approved):
Thanks!

PR #968, opened 2021-02-23
Safer pivot in CAN Loader

TomGoBravo:
Fix for problem tested by https://github.com/covid-projections/covid-data-model/pull/967 found while working on https://trello.com/c/hSJN82y9/927-pipe-through-source-names-to-fe.

* Adds failure when a field is added to the CAN parquet file that isn't in Fields so we can make sure adjustments don't need to be made
* Adds failure when a duplicate observation is found
* Changes aggregation from pivot_table to unstack
* Factors out `taglib.Source.attribute_df_to_json_series` so it can be used in a subsequent PR
* Removes `aggregate_level` from `query_multiple_variables` return value because it wasn't used

## Tested

`./run.py data update` output identical data.

### DIFF
--- a/libs/datasets/sources/can_scraper_helpers.py
+++ b/libs/datasets/sources/can_scraper_helpers.py
@@ -161,7 +161,6 @@ class CanScraperLoader:
                 :, [CommonFields.FIPS, CommonFields.DATE, PdFields.VARIABLE, Fields.SOURCE_URL]
             ].rename(columns={Fields.SOURCE_URL: taglib.TagField.CONTENT})
         else:
-            # TODO(tom): Make a better empty tag DataFrame
             combined_source_urls = pd.DataFrame([])
 
         indexed = long_df.set_index(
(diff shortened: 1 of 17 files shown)


libs/datasets/sources/can_scraper_helpers.py
```diff
@@ -139,47 +142,45 @@ def query_multiple_variables(
                     unit_counts=str(more_data[Fields.UNIT].value_counts().to_dict()),
                 )
 
-            # Rename fields if common field name exists
-            if variable.common_field:
-                data.loc[:, Fields.VARIABLE_NAME] = variable.common_field
+            data.loc[:, Fields.VARIABLE_NAME] = variable.common_field
```
ghop02, line 145: is `variable.common_field` always set?

ghop02 (approved):
one question but otherwise looks good

libs/datasets/sources/can_scraper_helpers.py
TomGoBravo: added comment in f5e7006

PR #976, opened 2021-02-24
copy source tag to source_url

TomGoBravo:
This PR is for dashboard PR https://github.com/covid-projections/covid-data-model/pull/958 and is one possible way to make a `pd.Series` of URLs in `source` tags.

* adds tag_objects_series to MultiRegionDataset
* fixes tag_objects_series handling of empty tag DF
* adds make_source_url_tags

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -52,6 +52,8 @@ TAG_INDEX_FIELDS = [
     TagField.TYPE,
 ]
 
+TAG_DF_COLUMNS = TAG_INDEX_FIELDS + [TagField.CONTENT]
+
 ANNOTATION_TAG_TYPES = [
     TagType.CUMULATIVE_LONG_TAIL_TRUNCATED,
     TagType.CUMULATIVE_TAIL_TRUNCATED,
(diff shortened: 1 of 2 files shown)


libs/datasets/timeseries.py
```diff
@@ -1673,3 +1694,17 @@ def make_source_tags(ds_in: MultiRegionDataset) -> MultiRegionDataset:
     )
 
     return ds_in.replace_tag_df(pd.concat([source_df, other_tags_df]))
+
+
+def make_source_url_tags(ds_in: MultiRegionDataset) -> MultiRegionDataset:
+    """Make source_url tags from source tags"""
+    assert TagType.SOURCE_URL not in ds_in.tag.index.get_level_values(TagField.TYPE)
```
ghop02, line 1701: I guess I don't quite understand the intent here.  Why do we need source_urls as their own tag type? shouldn't having `source` cover that case?

TomGoBravo: I want to display count the URLs for the dashboard so need a way to get them out of all the `source` JSONs. This is just one way to make that happen.

TomGoBravo: An alternative for https://github.com/covid-projections/covid-data-model/pull/958 is putting code that does roughly the same thing in the dashboard.

ghop02: It seems like that data would better live in the dashboard - or generally maybe some helper functions for accessing json data stored in the content (especially since I think all of the annotations store json now?)

TomGoBravo: I added a TODO about cleaning this up when we know more about what we want.

ghop02 (approved):

PR #977, opened 2021-02-24
Add static name and URL to non-CAN Scraper 

TomGoBravo:
This PR completes :crossed_fingers: https://trello.com/c/rI3EQlS6/976-add-nytimes-source-name-to-annotations

* Adds source name and URL to older data sources that are not part of the new CAN Scraper system

### DIFF
--- a/libs/datasets/sources/nytimes_dataset.py
+++ b/libs/datasets/sources/nytimes_dataset.py
@@ -5,7 +5,7 @@ from libs.datasets.taglib import UrlStr
 
 class NYTimesDataset(data_source.DataSource):
     SOURCE_TYPE = "NYTimes"
-    SOURCE_NAME = "New York Times"
+    SOURCE_NAME = "The New York Times"
     SOURCE_URL = UrlStr("https://github.com/nytimes/covid-19-data")
 
     COMMON_DF_CSV_PATH = "data/cases-nytimes/timeseries-common.csv"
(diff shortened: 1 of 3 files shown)


libs/datasets/sources/nytimes_dataset.py
```diff
@@ -1,9 +1,12 @@
 from covidactnow.datapublic.common_fields import CommonFields
 from libs.datasets import data_source
+from libs.datasets.taglib import UrlStr
 
 
 class NYTimesDataset(data_source.DataSource):
     SOURCE_TYPE = "NYTimes"
+    SOURCE_NAME = "New York Times"
```
ghop02, line 8: I think this should be "The New York Times” https://github.com/nytimes/covid-19-data#license-and-attribution

TomGoBravo: Fixed in feb9b826

TomGoBravo:
[Add NYTimes source name to annotations](https://trello.com/c/rI3EQlS6/976-add-nytimes-source-name-to-annotations)

BrettBoval:
Those names look fine to me.

ghop02 (approved):

mikelehen:
FWIW- I think cms_testing_dataset.py would ideally be in here too.

Name would be "The Centers for Medicare & Medicaid Services" and the URL would be https://data.cms.gov/stories/s/COVID-19-Nursing-Home-Data/bkwz-xpvg

PR #978, opened 2021-02-24
Add metric annotations

TomGoBravo:
This PR addresses https://trello.com/c/g0fSP3MT/974-change-annotation-names-to-match-metric-names

* Adds Annotations for metrics
* Adds some test that will hopefully make maintaining these less dependent on us noticing problems
* Changes `build_annotations` to use two maps of constants instead of long list of hard-coded function calls
* Changes `test_annotation` to put duplicate URLs in new_cases instead of cases. Currently `cases` is used multiple times which produced multiple log warning calls and broke the log event check in the test.

### DIFF
--- a/libs/build_api_v2.py
+++ b/libs/build_api_v2.py
@@ -116,14 +116,14 @@ ACTUALS_NAME_TO_COMMON_FIELD = {
 
 METRICS_NAME_TO_COMMON_FIELD = {
     "contactTracerCapacityRatio": CommonFields.CONTACT_TRACERS_COUNT,
-    "caseDensity": CommonFields.CASES,  # TODO(chris): Fix this
-    "infectionRate": CommonFields.CASES,  # TODO(chris): Fix this
+    "caseDensity": CommonFields.CASES,
+    "infectionRate": CommonFields.CASES,
     "testPositivityRatio": CommonFields.TEST_POSITIVITY,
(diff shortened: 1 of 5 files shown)


TomGoBravo:
[Change annotation names to match metric names](https://trello.com/c/g0fSP3MT/974-change-annotation-names-to-match-metric-names)

libs/build_api_v2.py
```diff
@@ -99,31 +99,45 @@ def build_region_summary(
...
+    "vaccinationsCompleted": CommonFields.VACCINATIONS_COMPLETED,
+}
+
+
+METRICS_NAME_TO_COMMON_FIELD = {
+    "contactTracerCapacityRatio": CommonFields.CONTACT_TRACERS_COUNT,
+    "caseDensity": CommonFields.CASES,  # TODO(chris): Fix this
+    "infectionRate": CommonFields.CASES,  # TODO(chris): Fix this
+    "testPositivityRatio": CommonFields.TEST_POSITIVITY,
+    "icuHeadroomRatio": CommonFields.CASES,  # TODO(chris): Fix this
```
ghop02, line 122: Should probably be `CommonFields.CURRENT_ICU_TOTAL`

ghop02 (approved):
I think this looks good! Only one field I think needs to change for right now.

libs/build_api_v2.py
TomGoBravo: Fixed in f4ef93a

PR #980, opened 2021-02-25
add more counts to Regions table

TomGoBravo:
Add count of timeseries, url and the difference (ie timeseries without url) to the Regions table. This helps to expose places where we can add more URLs.

Example output:
image

There are some counties with more URLs than timeseries, for example the following have a URL but no real value:
```
iso1:us#iso2:us-nv#fips:32009,vaccinations_completed
iso1:us#iso2:us-nv#fips:32009,vaccinations_initiated
iso1:us#iso2:us-nv#fips:32009,vaccines_administered
```

### DIFF
--- a/dash_app/dashboard.py
+++ b/dash_app/dashboard.py
@@ -1,3 +1,5 @@
+from dataclasses import dataclass
+
 import dash
 import dash_core_components as dcc
 import dash_html_components as html
@@ -12,6 +14,7 @@ from dash.dependencies import Input
 from dash.dependencies import Output
 from pandas.core.dtypes.common import is_numeric_dtype
 from plotly import express as px
+from typing_extensions import final
 
 from libs import pipeline
 from libs.datasets import AggregationLevel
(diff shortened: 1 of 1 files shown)


dash_app/dashboard.py
```diff
@@ -102,14 +92,32 @@ def init(server):
...
     counties = ds.get_subset(aggregation_level=AggregationLevel.COUNTY)
     # TODO(tom): update to work with Source tag
     has_url = counties.tag.loc[:, :, TagType.SOURCE_URL].unstack(PdFields.VARIABLE).notnull()
     county_variable_population_ratio = population_ratio_by_variable(counties, has_url)
 
+    # A table of regions in a DataFrame.
+    static_columns = pd.Index(
+        [CommonFields.LOCATION_ID, CommonFields.COUNTY, CommonFields.STATE, CommonFields.POPULATION]
+    ).intersection(ds.static.columns)
+    df_regions = ds.static.loc[:, static_columns].copy()
```
ghop02, line 107: seems like this function will be growing in size, might be good to pull the `df_regions` generation logic into it's own function

ghop02 (approved):

dash_app/dashboard.py
TomGoBravo: I tried to got a step further with a032a38 and created a dataclass to hold all the per region variables.

PR #982, opened 2021-02-28
Also exclude IA counties from NYTimes dataset

ghop02:
Removes IA counties from NYTimes dataset so we fall back on USA facts

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -14,6 +14,7 @@ from covidactnow.datapublic.common_fields import CommonFields
 from covidactnow.datapublic.common_fields import FieldName
 from typing_extensions import final
 
+from libs.datasets import AggregationLevel
 from libs.datasets import dataset_utils
 from libs.datasets import data_source
 from libs.datasets import dataset_pointer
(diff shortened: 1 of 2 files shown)


libs/datasets/combined_datasets.py
```diff
@@ -88,25 +91,33 @@ def SOURCE_TYPE(self):
...
+            return region_mask_to_location_ids(region_mask_or_regions)
+        elif isinstance(region_mask_or_regions, Region):
+            return [region_mask_or_regions.location_id]
+        else:
+            location_ids = []
+            for mask_or_region in region_mask_or_regions:
+                if isinstance(mask_or_region, RegionMask):
+                    location_ids.extend(region_mask_to_location_ids(mask_or_region))
+                else:
+                    location_ids.append(mask_or_region.location_id)
```
mikelehen, line 106: nit: You could probably replace the above 4 lines with a recursive call?

mikelehen (approved):

libs/datasets/combined_datasets.py
TomGoBravo: Or changing `RegionMaskOrRegions` to be only `Collection[Union[Region, RegionMask]]`, which is just a convenience so we don't need to wrap single regions in a list with one element.

PR #983, opened 2021-03-01
Dashboard browse by variable group and add timeseries county population ratio

TomGoBravo:
This PR
* Adds enums for how regions and variables are aggregated and use them to parameterize `_agg_wide_var_counts`
* Add `AggregatedStats` to hold stats after they have been aggregated
* Expand population ratio URL coverage table to include, for each variable, portion of population that has some county data
* Add dropdown in Region section at bottom to include only a subset of variables
* Factor out dash_table_from_data_frame


ghop02 (approved):
Haven't looked too deeply in this, but happy to approve to get better looking data up there!

PR #984, opened 2021-03-01
fix test_build_summary_for_fips by adding anomalies

TomGoBravo:

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/tests/libs/generate_api_v2_test.py
+++ b/tests/libs/generate_api_v2_test.py
@@ -83,7 +83,16 @@ def test_build_summary_for_fips(
             vaccinationsCompleted=nyc_latest.get("vaccinations_completed"),
         ),
         annotations=Annotations(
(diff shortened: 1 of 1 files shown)


tests/libs/generate_api_v2_test.py
```diff
@@ -125,11 +134,38 @@ def test_build_summary_for_fips(
...
+                sources=[field_source_usafacts],
+                anomalies=[
+                    {
+                        "date": datetime.date(2021, 2, 25),
+                        "original_observation": 102666.0,
+                        "type": "cumulative_tail_truncated",
+                    }
+                ],
+            ),
+            infectionRateCI90=FieldAnnotations(
```
ghop02, line 159: hmm interesting case - I'm not sure we want to pass the anomalies down to these fields.  Probably only want the source

ghop02 (approved):

PR #985, opened 2021-03-01
Split up timeseries

ghop02:
I've been meaning to split up the timeseries file a bit.  There's a lot going on, and I find the quite large file pretty difficult to navigate.  I think by splitting it up it'll some more code re-use/help compartmentalize some of what we do to different modules.


TomGoBravo (approved):
I haven't looked at this with any detail because I don't care much what file stuff is in... which is how we ended up with so much in one file :-P Please add something in PR desc of how you tested it.

ghop02:
Tested this by running `./run.py data update` and various unit tests

PR #986, opened 2021-03-02
Adding aggregate level to dataset if not included

ghop02:
This will fix the breakage in the data update by making sure that all data has an aggregate level.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -193,6 +193,16 @@ def _add_state_if_missing(df: pd.DataFrame):
         )
 
 
+def _add_aggregate_level_if_missing(df: pd.DataFrame):
+    """Adds the aggregate level column if missing, in place."""
+    assert CommonFields.LOCATION_ID in df.columns
+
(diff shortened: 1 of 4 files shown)


TomGoBravo (commented):
What is the breakage that this fixes? I don't see anything in the notification channels or recent suspicious commits in the git log.

ghop02:
> What is the breakage that this fixes? I don't see anything in the notification channels or recent suspicious commits in the git log.

This fixes: https://github.com/covid-projections/covid-data-model/actions/workflows/update_repo_datasets.yml

Also, it would fix any `get_subset` queries on datasets that don't have aggregate level

TomGoBravo (approved):
This'll do for now. It'd be nice to have better test coverage and some way to have subset query by level work for datasets not created via fips-related methods like from_fips_timeseries_df

ghop02:
> This'll do for now. It'd be nice to have better test coverage and some way to have subset query by level work for datasets not created via fips-related methods like from_fips_timeseries_df

Thanks, yup definitely agree!

PR #987, opened 2021-03-02
Buckets from CanScraperLoader

TomGoBravo:

* Changes CanScraperLoader to return rows of observations identified by location_id. It returns all buckets but getting buckets that are not "all" to work further downstream is a WIP
* Adds a check that `location_id` in parquet file matches what is computed from FIPS. This fails right now so added work around.
* Updates `test_helpers.assert_dataset_like` to compare `timeseries_bucketed` instead of `timeseries`

## Tested

Ran `data update` many many times, fixing various problems until it was identical. This is the expected state because stages of the pipeline after loading the data copy `timeseries` instead of `timeseries_bucketed`, dropping all timeseries that are not in bucket "all".


libs/datasets/sources/can_scraper_helpers.py
```diff
@@ -83,105 +82,147 @@ def _fips_from_int(param: pd.Series):
...
+                Fields.PROVIDER,
+                Fields.VARIABLE_NAME,
+                Fields.MEASUREMENT,
+                Fields.UNIT,
+                Fields.LOCATION_ID,
+            ]
+            + DEMOGRAPHIC_FIELDS
+            + [Fields.DATE]
+        ).sort_index()
+        all_ares = (
```
ghop02, line 120: What does ares refer to?

libs/datasets/sources/can_scraper_helpers.py
```diff
@@ -83,105 +82,147 @@ def _fips_from_int(param: pd.Series):
...
 
-        selected_data = []
-        for variable in variables:
-            # Check that `variable` agrees with stuff in the ScraperVariable docstring.
-            if variable.common_field is None:
-                assert variable.measurement == ""
-                assert variable.unit == ""
-                continue
+        # Split `variables` into lists of variables dropped and those returned.
+        variables_drop, variables_return = [], []
```
ghop02, line 157: Fwiw I think you could initialize these explicitly 

'''
variables_to_drop = [variable for variable in variables if not variable.common_field]
common_field_variables = [variable for variable in variables if variable.common_field]
'''

Just makes it a bit easier to read the loops

ghop02 (approved):

PR #988, opened 2021-03-02
Freeze column ordering for CSV + adding vaccination initiated/completed ratio to output

ghop02:


libs/dataset_deployer.py
```diff
@@ -89,8 +93,11 @@ def write_nested_csv(
     first_row = data[0]
     first_row = {key: value for key, value in first_row.items() if key not in keys_to_skip}
 
-    header = flatten_dict(first_row).keys()
-    header = [column for column in header if column not in keys_to_skip]
+    if columns:
```
TomGoBravo, line 96: how about making columns required and removing the branch?

ghop02: I do like the interface of a specified header being non-required, but I did simplify the branch a lot

libs/dataset_deployer.py
```diff
@@ -71,26 +71,23 @@ def flatten_dict(data: dict, level_separator: str = ".") -> dict:
...
         raise ValueError("Cannot upload a 0 length list.")
 
     first_row = data[0]
-    first_row = {key: value for key, value in first_row.items() if key not in keys_to_skip}
+    first_row = {key: value for key, value in first_row.items()}
 
-    header = flatten_dict(first_row).keys()
-    header = [column for column in header if column not in keys_to_skip]
+    all_columns = list(flatten_dict(first_row).keys())
+    header = header or all_columns
```
TomGoBravo, line 90: not fooling me, you've just hidden the branch :-P More seriously, why maintain functionality that isn't used in production? when refactoring I want to quickly see all the ways a function is called by non-test code. It is nice to get that information from the function prototype without manually inspecting each call site. If keeping `header` as `Optional` is useful for tests or future development please document it.

libs/dataset_deployer.py
```diff
@@ -71,26 +71,23 @@ def flatten_dict(data: dict, level_separator: str = ".") -> dict:
...
     if not data:
         raise ValueError("Cannot upload a 0 length list.")
 
     first_row = data[0]
-    first_row = {key: value for key, value in first_row.items() if key not in keys_to_skip}
+    first_row = {key: value for key, value in first_row.items()}
 
-    header = flatten_dict(first_row).keys()
-    header = [column for column in header if column not in keys_to_skip]
+    all_columns = list(flatten_dict(first_row).keys())
```
TomGoBravo, line 89: when will we find out that all_columns and header have diverged? a unit test would be nice, with an assert during production run is okayish, during QA is bad.

ghop02: I guess I'm leaning on the fact that this is not very likely - this would mean we've changed the names of the fields in the pydantic api field definitions.  In that case, we would be breaking the API for everyone using the JSON files - something I don't think we intend to do.

libs/dataset_deployer.py
ghop02: I definitely see your point - I think i prefer the optional argument here because this function is primarily meant to just write data in a list of dicts to a csv.  If i were looking at that function, I would expect to be able to just write the csv by passing the dictionary and a file path.  I think if we make the `header` (or could change it to `header_override`) required, I would change the name of the function to indicate that that's write nested csv in specific heading order.  

I dont' think that's really necessary, so I don't see it as much as additional functionality - more as api fluency.

libs/dataset_deployer.py
TomGoBravo: What about a typo in `csv_column_ordering.py`?

libs/dataset_deployer.py
TomGoBravo: I'd rather see the awkward reality instead of a more elegant alternative.... though I have a hunch that isn't how you see the two choices here.

TomGoBravo (approved):
Thank you for fixing the lint errors.

libs/dataset_deployer.py
ghop02: ah I think i'll add a test to make sure that these match properties in the matching schemas.

PR #989, opened 2021-03-02
can_scraper_helpers_test using make_dataset

TomGoBravo:

* change can_scraper_helpers_test to call make_dataset instead of query_multiple_variables, so the value returned by query_multiple_variables can be modified without changing the tests.
* call _get_covid_county_dataset with cls so it can be overridden in subclasses, removing need to mock it
* change CanScraperBase to compute `EXPECTED_FIELDS` from `VARIABLES` with some funky metaclass stuff.
* add some unit tests checking the constants in subclasses of CanScraperBase

## Tested

Tests pass except for tests/libs/generate_api_v2_test.py::test_build_summary_for_fips, which is expected to be fixed by https://github.com/covid-projections/covid-data-model/pull/988

### DIFF
--- a/tests/libs/datasets/data_source_test.py
+++ b/tests/libs/datasets/data_source_test.py
@@ -119,9 +119,16 @@ def test_can_scraper_class_unique_variable_names():
         assert len(variables_not_none) == len(set(v.common_field for v in variables_not_none))
 
 
-def test_can_scraper_class_variable_set_expected_fields_unset():
+def test_can_scraper_class_variable_set():
     cls: data_source.CanScraperBase
     for cls in test_helpers.get_concrete_subclasses(data_source.CanScraperBase):
-        # EXPECTED_FIELDS is empty. It is computed from VARIABLES.
-        assert not cls.EXPECTED_FIELDS
         assert cls.VARIABLES
+
+
+def test_can_scraper_class_expected_fields_set():
+    cls: data_source.DataSource
+    for cls in test_helpers.get_concrete_subclasses(data_source.DataSource):
+        assert cls.EXPECTED_FIELDS
+        assert isinstance(cls.EXPECTED_FIELDS, list)
+        assert cls.SOURCE_TYPE
+        assert isinstance(cls.SOURCE_TYPE, str)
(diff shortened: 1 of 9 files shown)


tests/libs/datasets/data_source_test.py
```diff
@@ -47,11 +46,12 @@ class CANScraperForTest(can_scraper_usafacts.CANScraperUSAFactsProvider):
         # observations are not sorted by increasing date.
         input_data = input_data.iloc[::-1]
 
-    data = ccd_helpers.CanScraperLoader(input_data)
+    class CANScraperForTest(can_scraper_usafacts.CANScraperUSAFactsProvider):
+        @staticmethod
+        def _get_covid_county_dataset():
```
ghop02, line 51: If this is a classmethod, you could probably pretty easily mock out this using mocks rather than needing to create a test class. 

but this method is fine as well!

tests/libs/datasets/data_source_test.py
```diff
@@ -98,3 +98,30 @@ class NYTimesForTest(nytimes_dataset.NYTimesDataset):
         static=region_static,
     )
     test_helpers.assert_dataset_like(dataset_expected, dataset_read)
+
+
+def test_can_scraper_class_single_provider():
+    cls: data_source.CanScraperBase
```
ghop02, line 104: I haven't seen this sort of syntax before, is this the recommended way to show typing for this case?

ghop02 (approved):

tests/libs/datasets/data_source_test.py
TomGoBravo: Ha, I thought the subclass and override was preferable to mocking. Getting this working without breaking the cache was a bit fiddly. classmethod doesn't seem to play with lru_cache as well. Cleaning up all this inheritance/class stuff is an exercise for another day.

tests/libs/datasets/data_source_test.py
TomGoBravo: Yes. https://www.python.org/dev/peps/pep-0526/#where-annotations-aren-t-allowed

PR #1002, opened 2021-03-05
Add MultiRegionDataset.timeseries_bucketed

TomGoBravo:

* Replaces property `MultiRegionDataset.timeseries` with `timeseries_bucketed` that has an index level for the demographic bucket. Access to the old attribute is preserved with a `@cached_property` method.
* Adds a custom `MultiRegionDataset.__init__` that transforms `timeseries` to `timeseries_bucketed` when an object is created. Even calls to `dataclasses.replace` go through this. This constructor raises in the ambiguous case of both parameters being not None.
* Changes existing places that create a `MultiRegionDataset` using `dataclasses.replace` to include `timeseries_bucketed=None` . Subsequent PRs will replace each of these code paths with something that preserves the bucket information.

## Tested

`tests/libs/generate_api_v2_test.py::test_build_summary_for_fips` is known to be failing.


libs/datasets/timeseries.py
```diff
@@ -1099,7 +1153,9 @@ def process_series(series):
 
     df_copy.loc[to_exclude[to_exclude].index, CommonFields.TEST_POSITIVITY_7D] = np.nan
 
-    return dataclasses.replace(dataset, timeseries=df_copy).append_tag_df(new_tags.as_dataframe())
+    return dataclasses.replace(dataset, timeseries=df_copy, timeseries_bucketed=None).append_tag_df(
```
ghop02, line 1156: Why does this need `=None`? isn't this removing any existing timeseries_bucketed objects?

ghop02: I guess I'm worried that bucketed data will just get thrown out very quickly

TomGoBravo: The new `__init__` enforces that exactly one of timeseries_bucketed and timeseries is not None. Whatever was passed in, it stores timeseries_bucketed. Without `timeseries_bucketed=None` replace will pass both to `__init__`. Next I'll deprecate and replace use of timeseries.

ghop02 (approved):
LGTM!

PR #1004, opened 2021-03-08
Moving new cases + adding new deaths to combined datasets

ghop02:
Adds new deaths to combined datasets + moves and refactors the new cases logic a bit


tests/libs/datasets/new_cases_and_deaths_test.py
```diff
@@ -0,0 +1,134 @@
...
+            "iso1:us#fips:4,,,\n"
+        )
+    )
+
+    timeseries_after = new_cases_and_deaths.add_new_cases(mrts_before)
+    test_helpers.assert_dataset_like(mrts_expected, timeseries_after)
+
+
+def test_calculate_new_deaths():
+    mrts_before = timeseries.MultiRegionDataset.from_csv(
```
TomGoBravo, line 47: When adding new tests I think moving away from hard coded CSVs to the test_helper stuff is more important than consistency with surrounding code.

TomGoBravo (approved):

tests/libs/datasets/new_cases_and_deaths_test.py
ghop02: 👍  will update that

PR #1008, opened 2021-03-09
fix CAN Scraper tests

ghop02:
After I added the aggregate level addition, needed to fix the expected output of this test


TomGoBravo (approved):
What do you think of making state, fips and level be not loaded and written like normal variables and instead be something that MultiRegionDataset joins with every dataset when it is created and drops before writing?

PR #1011, opened 2021-03-09
MultiRegionDataset.geo_data

TomGoBravo:
This PR addresses https://trello.com/c/cmh0RhsA/1071-remove-geo-data-from-multiregiondatasetsstatic

A MultiRegionDataset needs level,state (aka geo data columns) etc for making subsets of it to work consistently. Before this change subset may or may not work depending on how the MultiRegionDataset was created. This PR fixes that be making sure every MultiRegionDataset has geo data.

* Adds `data/geo-data.csv`, a location_id indexed table of values (FIPS, county name, state, country, level) that we can maintain manually because they are even more static than the scalars in `ALL_FIELDS_FEATURE_DEFINITION`.
* Drops geo data columns when creating a MultiRegionDataset, warning when values conflict with geo-data.csv.
* Adds geo_data property that dynamically makes a table of the geo-data for location in the dataset when needed
* Moves most of the logic for interpreting RegionMaskOrRegion from combined_datasets to timeseries now that the later has consistent access to geo data.
* Fixes a bunch of tests that had bogus FIPS and location_id

## Tested

`./run.py data update` then 
`git cat-file blob main:data/multiregion-static.csv| git lfs smudge | csvcut -C fips,state,country,county,aggregate_level > data/multiregion-static-main.csv` to get the existing data without columns expected to be removed.
`diff data/multiregion-static.csv data/multiregion-static-main.csv` shows no diff and `git status` lists only the expected json and static data changes.


libs/datasets/timeseries.py
```diff
@@ -398,6 +399,20 @@ def timeseries(self) -> pd.DataFrame:
                 dtype="float",
             )
 
+    def location_ids(self) -> pd.Index:
```
ghop02, line 402: I think this should be a property

libs/datasets/timeseries.py
```diff
@@ -398,6 +399,20 @@ def timeseries(self) -> pd.DataFrame:
...
+        return self.static.index.unique(CommonFields.LOCATION_ID).union(
+            self.timeseries_bucketed.index.unique(CommonFields.LOCATION_ID)
+        )
+
+    @cached_property
+    def geo_data(self) -> pd.DataFrame:
```
ghop02, line 408: hmm, how do you feel about loading geo_data on init, or providing a helper static function that loads it and passes it to init

libs/datasets/timeseries.py
```diff
@@ -398,6 +399,20 @@ def timeseries(self) -> pd.DataFrame:
...
+        location_ids = self.location_ids()
+        geo_data = dataset_utils.get_geo_data()
+        missing_location_id = location_ids.difference(geo_data.index)
+        if not missing_location_id.empty:
+            raise KeyError(f"location_id not in data/geo-data.csv:\n{missing_location_id}")
+        return dataset_utils.get_geo_data().reindex(index=location_ids)
```
ghop02, line 414: the `get_geo_data` function is called twice here, probably should make it a variable?

tests/test_helpers.py
```diff
@@ -77,9 +64,9 @@ def __init__(
         # let the dataclasses code initialize `data`.
         self.__default_init__(  # pylint: disable=E1101
             *args,
-            provenance=_to_list(provenance),
-            source_url=_to_list(source_url),
-            source=_to_list(source),
+            provenance=to_list(provenance),
```
ghop02, line 67: nit: i think importing this as `combined_datasets` and calling `combined_datasets.to_list` here is a bit more clear when reading through but low priority

libs/datasets/timeseries.py
TomGoBravo: Done in ff9dd015

tests/test_helpers.py
TomGoBravo: Done in 7a6c56e0

libs/datasets/timeseries.py

ghop02 (approved):

libs/pipeline.py
```diff
@@ -25,35 +29,19 @@ class BadFipsWarning(UserWarning):
 
 def fips_to_location_id(fips: str) -> str:
```
ghop02, line 30: I noticed there were a couple other use cases of this which are calling this function `.apply(fips_to_location_id)` I think it's worth converting those calls to map rather than apply - seemed faster locally

ghop02 (approved):

PR #1012, opened 2021-03-10
change can_scraper_helpers BadLocationId warning to raises

TomGoBravo:
This PR follows up on https://github.com/covid-projections/covid-data-model/pull/987 and https://covidactnow.slack.com/archives/C015V4JS952/p1615408720113400 

## Tested

`./run.py data update` produced identical `data/multiregion-wide-dates.csv`

### DIFF
--- a/libs/datasets/sources/can_scraper_helpers.py
+++ b/libs/datasets/sources/can_scraper_helpers.py
@@ -1,6 +1,5 @@
 """Helpers to access and query data loaded from the CAN Scraper parquet file.
 """
-import warnings
 from typing import List
 import enum
 import dataclasses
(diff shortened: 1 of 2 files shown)


libs/datasets/sources/can_scraper_helpers.py
```diff
@@ -136,10 +135,6 @@ def _make_indexed_df(all_df: pd.DataFrame) -> pd.DataFrame:
             .rename(PdFields.DEMOGRAPHIC_BUCKET)
         )
         CanScraperLoader._check_location_id(all_df)
-        # Hack fix for bad location_ids.
-        all_df[Fields.LOCATION_ID] = _fips_from_int(all_df[Fields.LOCATION]).apply(
-            pipeline.fips_to_location_id
```
sglyon, line 141: Is this a function that can be deleted now?

sglyon (changes requested):

libs/datasets/sources/can_scraper_helpers.py
TomGoBravo: No, it is used in _check_location_id. "Trust, but verify" :-P

sglyon (approved):

PR #1016, opened 2021-03-11
Chris fixes for PR

ghop02:
Adds fixes for PR static value aggregation + adds a helper flag that skips all dataset filters

### DIFF
--- a/cli/data.py
+++ b/cli/data.py
@@ -66,41 +66,10 @@ def main():
     pass
 
 
-@main.command()
-@click.option("--filename", default="external_forecasts.csv")
-def update_forecasts(filename):
-    """Updates external forecasts to the current checked out covid data public commit"""
-    path_prefix = dataset_utils.DATA_DIRECTORY.relative_to(dataset_utils.REPO_ROOT)
-    data_root = dataset_utils.LOCAL_PUBLIC_DATA_PATH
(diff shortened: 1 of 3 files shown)


cli/data.py
```diff
@@ -66,6 +66,37 @@ def main():
     pass
 
 
+def _run_dataset_filters(
```
TomGoBravo, line 69: Good enough for now. Longer term I'd rather have a way to keep a copy of before and after the filters, than be able to see what changed at those points of the pipeline as input data and/or code changes.

ghop02: 👍 i could imagine a world where this is 3 steps with separate outputs - combine, filter, aggregate

tests/libs/datasets/custom_aggregations_test.py
```diff
@@ -78,3 +78,28 @@ def test_calculate_puerto_rico_bed_occupancy_rate():
         )
     )
     test_helpers.assert_dataset_like(actual, expected)
+
+
+def test_pr_aggregation():
+    dataset = combined_datasets.load_us_timeseries_dataset()
```
TomGoBravo, line 84: i'll likely exclude these from the fast tests but *much* better than nothing.

TomGoBravo (approved):
Thank you for adding these.

PR #1018, opened 2021-03-15
Time-series bucketed long

TomGoBravo:
This PR is part of https://trello.com/c/lvRFxC5y/932-get-dei-vaccination-data-into-api. It makes bucketed and not bucketed more explicit.

* Changes method `timeseries_wide_dates()` to property `timeseries_not_bucketed_wide_dates` and adds `timeseries_bucketed_wide_dates` which will gradually replace it.
* Changes `test_timeseries_long` to use test_helpers instead of CSV in a string to create input dataset
* Removes unused `_timeseries_long`

## Tested

### DIFF
--- a/tests/libs/datasets/timeseries_test.py
+++ b/tests/libs/datasets/timeseries_test.py
@@ -346,16 +346,14 @@ def test_append_regions_duplicate_region_raises():
 
 
 def test_timeseries_long():
-    ts = timeseries.MultiRegionDataset.from_csv(
-        io.StringIO(
-            "location_id,date,county,aggregate_level,m1,m2\n"
-            "iso1:us#cbsa:10100,2020-04-02,,,,2\n"
-            "iso1:us#cbsa:10100,2020-04-03,,,,3\n"
-            "iso1:us#cbsa:10100,,,,,3\n"
-            "iso1:us#fips:97111,2020-04-02,Bar County,county,2,\n"
-            "iso1:us#fips:97111,2020-04-04,Bar County,county,4,\n"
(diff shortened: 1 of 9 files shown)


tests/libs/datasets/timeseries_test.py
```diff
@@ -346,40 +346,86 @@ def test_append_regions_duplicate_region_raises():
...
+        io.StringIO(
+            "       location_id,demographic_bucket,      date,variable,value\n"
+            "iso1:us#cbsa:10100,               all,2020-04-02,      m2,    2\n"
+            "iso1:us#cbsa:10100,               all,2020-04-03,      m2,    3\n"
+            "iso1:us#fips:97111,           age:0-9,2020-04-02,      m1,    4\n"
+            "iso1:us#fips:97111,           age:0-9,2020-04-03,      m1,    5\n"
+            "iso1:us#fips:97111,           age:0-9,2020-04-04,      m1,    6\n"
+            "iso1:us#fips:97111,         age:10-19,2020-04-04,      m1,    7\n"
+            "iso1:us#fips:97111,               all,2020-04-02,      m1,    2\n"
+            "iso1:us#fips:97111,               all,2020-04-04,      m1,    4\n".replace(" ", "")
```
ghop02, line 413: Do we need all of these rows to test out all this functionality?

ghop02 (approved):

tests/libs/datasets/timeseries_test.py
TomGoBravo: Not with the current implementation but I like having some coverage of one location-variable with one time-series and distributions with one time-series and more than one time-series.

PR #1020, opened 2021-03-16
add bucket to tags

TomGoBravo:
This PR is part of https://trello.com/c/lvRFxC5y/932-get-dei-vaccination-data-into-api.

* Adds `DEMOGRAPHIC_BUCKET` in the index of most tag DataFrame and Series representations
* Adds cached property `tag_all_bucket` that is just the tags in the bucket "all" and use it in annotations_all_bucket, sources_all_bucket etc
* Changes some calls to `index.get_level_values` with `index.unique` where we don't need a copy of the value for each row. (https://stackoverflow.com/a/50364903)

## Tested

### DIFF
--- a/libs/datasets/data_source.py
+++ b/libs/datasets/data_source.py
@@ -127,6 +127,8 @@ class CanScraperBase(DataSource, abc.ABC, metaclass=_CanScraperBaseMeta):
                 .last()
                 .reset_index()
                 .drop(columns=[CommonFields.DATE])
+                .copy()
             )
+            timeseries._tag_df_add_all_bucket_in_place(source_tag_df)
             ds = ds.append_tag_df(source_tag_df)
         return ds
--- a/libs/datasets/data_source.py
+++ b/libs/datasets/data_source.py
@@ -80,7 +80,7 @@ class DataSource(object):
         input_path = data_root / cls.COMMON_DF_CSV_PATH
         data = common_df.read_csv(input_path, set_index=False)
         data = cls._check_data(data)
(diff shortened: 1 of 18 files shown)


libs/datasets/data_source.py
```diff
@@ -127,6 +127,8 @@ def make_dataset(cls) -> timeseries.MultiRegionDataset:
                 .last()
                 .reset_index()
                 .drop(columns=[CommonFields.DATE])
+                .copy()
```
ghop02, line 130: Why is the copy needed here?

libs/datasets/data_source.py
```diff
@@ -127,6 +127,8 @@ def make_dataset(cls) -> timeseries.MultiRegionDataset:
...
             )
+            timeseries._tag_df_add_all_bucket_in_place(source_tag_df)
```
ghop02, line 132: since we're accessing this outside of the class, would prefer if it was just `tag_df_...`

libs/datasets/timeseries.py
```diff
@@ -346,6 +365,17 @@ def _check_timeseries_wide_vars_structure(wide_vars_df: pd.DataFrame, *, buckete
     assert numeric_columns.all()
 
 
+def _tag_add_all_bucket(tag: pd.Series) -> pd.Series:
+    tag_bucketed = pd.concat(
+        {DemographicBucket("all"): tag}, names=[PdFields.DEMOGRAPHIC_BUCKET]
```
ghop02, line 370: DemographicBucket.ALL?

ghop02 (approved):

PR #1021, opened 2021-03-17
add support for reading and writing with bucket.

TomGoBravo:
This PR is part of https://trello.com/c/lvRFxC5y/932-get-dei-vaccination-data-into-api.

* Changes `timeseries_bucketed_wide_dates` to return bucket as inner/right index level to be more consistent with idea that a set of time-series form a distribution within one metric/variable.
* Changes `read_from_pointer` to load the bucket if the column is present and assume `all` if not
* Changes `timeseries_rows` (which is called by `write_to_dataset_pointer`) to include bucket in the index

## Tested

`./run.py data update` adds demographic_bucket column always containing "all". 

Confirmed that nothing else changed with
```
mv data/multiregion-wide-dates.csv data/multiregion-wide-dates-demo.csv
csvcut -C demographic_bucket data/multiregion-wide-dates-demo.csv | perl -np -e 's/,+$//' > data/multiregion-wide-dates.csv
git status
```

### DIFF
--- a/libs/datasets/tail_filter.py
+++ b/libs/datasets/tail_filter.py
@@ -19,7 +19,7 @@ TagField = taglib.TagField
 # The Series.name in apply is a tuple copied from the index. Make an assert fail when the index
 # names change so we know to update access to the tuple elements.
 # https://stackoverflow.com/questions/26658240/getting-the-index-of-a-row-in-a-pandas-apply-function
-_EXPECTED_INDEX_NAMES = [CommonFields.LOCATION_ID, PdFields.DEMOGRAPHIC_BUCKET, PdFields.VARIABLE]
+_EXPECTED_INDEX_NAMES = [CommonFields.LOCATION_ID, PdFields.VARIABLE, PdFields.DEMOGRAPHIC_BUCKET]
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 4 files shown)


libs/datasets/tail_filter.py
```diff
@@ -132,8 +132,8 @@ def _filter_one_series(self, series_in: pd.Series) -> pd.Series:
                 ),
                 # `name` is a tuple with elements _EXPECTED_INDEX_NAMES
                 location_id=series_in.name[0],
-                variable=series_in.name[2],
-                bucket=series_in.name[1],
+                variable=series_in.name[1],
+                bucket=series_in.name[2],
```
ghop02, line 136: this might be helpful as a namedtuple

ghop02 (approved):
LGTM!

libs/datasets/tail_filter.py
TomGoBravo: You are not the first with that idea: https://stackoverflow.com/questions/26658240/getting-the-index-of-a-row-in-a-pandas-apply-function#comment110869633_26658301 and https://github.com/pandas-dev/pandas/issues/8574#issuecomment-59735379 says there is an issue but I can't find it.

PR #1025, opened 2021-03-17
[WIP] Prototype of demographic data in API

ghop02:
This is just a first pass at trying to pass the demographic data through to the API.

There's a bit of duplication with latest and bucketed_latest in one region, but I think we kinda need it for now until static has buckets.

### DIFF
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import List, Optional, Dict
 import enum
 
 from libs.datasets.dataset_utils import AggregationLevel
@@ -191,6 +191,8 @@ This value may vary by type of vaccine, but for Moderna and Pfizer this indicate
 number of people vaccinated with both the first and second dose.
 """,
     )
+    casesByAge: Optional[Dict[str, int]] = pydantic.Field(None)
+
     # When adding a new "actual" field here remember to add a `FieldAnnotations` in `Annotations`.
 
 
@@ -345,6 +347,7 @@ class Metrics(base_model.APIBaseModel):
             infectionRate=None,
             infectionRateCI90=None,
             icuHeadroomRatio=None,
+            icuCapacityRatio=None,
         )
 
 
@@ -388,6 +391,18 @@ class RiskLevels(base_model.APIBaseModel):
     infectionRate: RiskLevel = pydantic.Field(..., description="Infection rate risk level.")
     icuHeadroomRatio: RiskLevel = pydantic.Field(..., description="ICU headroom ratio risk level.")
     icuCapacityRatio: RiskLevel = pydantic.Field(..., description="ICU capacity ratio risk level.")
+
+    @classmethod
+    def empty(cls) -> "RiskLevels":
+        return RiskLevels(
+            overall=RiskLevel.LOW,
+            testPositivityRatio=RiskLevel.LOW,
+            caseDensity=RiskLevel.LOW,
+            contactTracerCapacityRatio=RiskLevel.LOW,
+            infectionRate=RiskLevel.LOW,
+            icuHeadroomRatio=RiskLevel.LOW,
+            icuCapacityRatio=RiskLevel.LOW,
+        )
 
 
 # Additional class used for bulk timeseries where we are not including all risk levels
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -60,7 +60,7 @@ class NonCovidPatientsMethod(enum.Enum):
     ESTIMATED_FROM_TOTAL_ICU_ACTUAL = "estimated_from_total_icu_actual"
 
 
-class Demographics(base_model.APIBaseModel):
+class DemographicDistributions(base_model.APIBaseModel):
 
     age: Optional[Dict[str, int]] = pydantic.Field(None)
     race: Optional[Dict[str, int]] = pydantic.Field(None)
@@ -202,7 +202,7 @@ number of people vaccinated with both the first and second dose.
     vaccinesAdministered: Optional[int] = pydantic.Field(
         None, description="Total number of vaccine doses administered."
     )
-    vaccinesAdministeredDemographics: Optional[Demographics] = pydantic.Field(None)
+    vaccinesAdministeredDemographics: Optional[DemographicDistributions] = pydantic.Field(None)
     # When adding a new "actual" field here remember to add a `FieldAnnotations` in `Annotations`.
 
 
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -235,7 +235,73 @@ class FieldAnnotations(base_model.APIBaseModel):
     anomalies: List[AnomalyAnnotation]
 
 
-Annotations = NewType("Annotations", Dict[str, Optional[FieldAnnotations]])
+class Annotations(base_model.APIBaseModel):
+    """Annotations for each field."""
+
+    # Keep this list of fields in sync with the fields in `Actuals`
+    cases: Optional[FieldAnnotations] = pydantic.Field(None, description="Annotations for cases")
+    deaths: Optional[FieldAnnotations] = pydantic.Field(None, description="Annotations for deaths")
+    positiveTests: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for positiveTests"
+    )
+    negativeTests: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for negativeTests"
+    )
+    contactTracers: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for contactTracers"
+    )
+    hospitalBeds: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for hospitalBeds"
+    )
+    icuBeds: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for icuBeds"
+    )
+    newCases: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for newCases"
+    )
+    newDeaths: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for newDeaths"
+    )
+    vaccinesDistributed: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for vaccinesDistributed"
+    )
+    vaccinationsInitiated: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for vaccinationsInitiated"
+    )
+    vaccinationsCompleted: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for vaccinationsCompleted"
+    )
+    vaccinesAdministered: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for vaccinesAdministered"
+    )
+    # Keep this list of fields in sync with the fields in `Metrics`
+    testPositivityRatio: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for testPositivityRatio"
+    )
+    caseDensity: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for caseDensity"
+    )
+    contactTracerCapacityRatio: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for contactTracerCapacityRatio"
+    )
+    infectionRate: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for infectionRate"
+    )
+    infectionRateCI90: Optional[FieldAnnotations] = pydantic.Field(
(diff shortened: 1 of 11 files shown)


libs/build_api_v2.py
```diff
@@ -35,17 +35,28 @@
 USA_VACCINATION_START_DATE = datetime(2020, 12, 14)
 
 
-def _build_actuals(actual_data: dict) -> Actuals:
+def _select_category(category: str, data: dict):
```
TomGoBravo, line 38: Add comment referring to `make_short_name` which generated the string this method parses. Or even better, move them to the same module.

libs/datasets/timeseries.py
```diff
@@ -539,15 +550,23 @@ def timeseries_not_bucketed_wide_dates(self) -> pd.DataFrame:
...
         long = self.timeseries.stack().droplevel(CommonFields.DATE)
         # `long` has MultiIndex with LOCATION_ID and VARIABLE (added by stack). Keep only the last
         # row with each index to get the last value for each date.
         unduplicated_and_last_mask = ~long.index.duplicated(keep="last")
+
         return long.loc[unduplicated_and_last_mask, :].unstack()
 
+    @cached_property
+    def _timeseries_bucketed_latest_values(self) -> pd.DataFrame:
+        long_bucketed = self.timeseries_bucketed.stack().droplevel(CommonFields.DATE)
```
TomGoBravo, line 566: How about adding a sort_index() between stack and droplevel to make sure it is sorted by date?

libs/datasets/timeseries.py
```diff
@@ -539,15 +550,23 @@ def timeseries_not_bucketed_wide_dates(self) -> pd.DataFrame:
...
+        unduplicated_bucketed_and_last_mask = ~long_bucketed.index.duplicated(keep="last")
+        return long_bucketed.loc[unduplicated_bucketed_and_last_mask, :].unstack()
```
TomGoBravo, line 568: add level name in `unstack` to avoid depending on order of levels.

tests/libs/generate_api_v2_test.py
```diff
@@ -194,3 +202,35 @@ def test_generate_timeseries_for_fips(nyc_region, nyc_rt_dataset, nyc_icu_datase
     # Double checking that serialized json does not contain NaNs, all values should
     # be serialized using the simplejson wrapper.
     assert "NaN" not in region_timeseries.json()
+
+
+def test_multiple_distributions():
+    """All time-series within a variable are treated as a unit when combining"""
+    m1 = FieldName("cases")
```
TomGoBravo, line 209: s/m1/cases/g
similarly for m2

api/can_api_v2_definition.py
```diff
@@ -227,70 +235,7 @@ class FieldAnnotations(base_model.APIBaseModel):
...
-    icuCapacityRatio: Optional[FieldAnnotations] = pydantic.Field(
-        None, description="Annotations for icuCapacityRatio"
-    )
-    vaccinationsInitiatedRatio: Optional[FieldAnnotations] = pydantic.Field(
-        None, description="Annotations for vaccinationsInitiatedRatio"
-    )
-    vaccinationsCompletedRatio: Optional[FieldAnnotations] = pydantic.Field(
-        None, description="Annotations for vaccinationsCompletedRatio"
-    )
+Annotations = NewType("Annotations", Dict[str, Optional[FieldAnnotations]])
```
TomGoBravo, line 238: I thought we had the explicit fields in Annotations, while kind of annoying to maintain, to make the JSON structure documented. Why change it?

api/can_api_v2_definition.py
```diff
@@ -60,6 +60,14 @@ class NonCovidPatientsMethod(enum.Enum):
     ESTIMATED_FROM_TOTAL_ICU_ACTUAL = "estimated_from_total_icu_actual"
 
 
+class Demographics(base_model.APIBaseModel):
```
TomGoBravo, line 63: I don't think `Demographics` describes what this holds very clearly. How about calling this DemographicDistribution or Distributions or ByDemographicBucket?

api/can_api_v2_definition.py
```diff
@@ -191,7 +199,7 @@ class Actuals(base_model.APIBaseModel):
 number of people vaccinated with both the first and second dose.
 """,
     )
-    casesByAge: Optional[Dict[str, int]] = pydantic.Field(None)
+    vaccinesAdministeredDemographics: Optional[Demographics] = pydantic.Field(None)
```
TomGoBravo, line 202: Hmm, well here `vaccinesAdministeredDemographics` seems like an okay name, I think because here Demographics is helping to describe a quantity but when it is a class name on its own I'm expecting it to be a noun that describes the stuff in it.

api/can_api_v2_definition.py
ghop02: So the main reason why I was playing around with changing it here was because I wasn't totally sure how I was going to represent the source annotations in the API response, so I was thinking this would be a string `vaccinesAdministeredDemographics.age`.  But I might just be able to lump them all together for `vaccinesAdministeredDemographics` as they should be the same

for instance, "vaccinesAdministeredDemographics.age"

api/can_api_v2_definition.py
ghop02: Yeah I'll think of a better name for it!

libs/build_api_v2.py
```diff
@@ -35,17 +39,52 @@
 USA_VACCINATION_START_DATE = datetime(2020, 12, 14)
 
 
-def _build_actuals(actual_data: dict) -> Actuals:
+_logger = structlog.get_logger()
+
+
+def _filter_demographic_field_and_remove_prefix(demographic_field: str, data: Dict):
```
TomGoBravo, line 45: What is data?

libs/build_api_v2.py
```diff
@@ -35,17 +39,52 @@
...
+def _build_demographic_data_for_field(
+    field_bucket: Dict[str, int]
+) -> Optional[DemographicDistributions]:
+
+    data = {
+        field: _filter_demographic_field_and_remove_prefix(field, field_bucket) or None
+        for field in can_scraper_helpers.DEMOGRAPHIC_FIELDS
+    }
+
+    # If there is no demographic data, do not create a DemographicDistributions
```
TomGoBravo, line 62: What about the default, `all`?

libs/build_api_v2.py
```diff
@@ -141,7 +180,7 @@ def build_annotations(one_region: OneRegionTimeseriesDataset, log) -> Annotation
         annotations_name: _build_metric_annotations(one_region, field_name, log)
         for annotations_name, field_name in name_and_common_field
     }
-    return Annotations(**annotations)
+    return {k: v for k, v in annotations.items() if v is not None}
```
TomGoBravo, line 183: do you plan to merge with a dict or `Annotations`?

libs/build_api_v2.py
```diff
@@ -35,17 +39,52 @@
...
 
-def _build_actuals(actual_data: dict) -> Actuals:
+_logger = structlog.get_logger()
+
+
+def _filter_demographic_field_and_remove_prefix(demographic_field: str, data: Dict):
+    return {
+        key.replace(demographic_field + ":", ""): value
+        for key, value in data.items()
+        if key.startswith(demographic_field)
```
TomGoBravo, line 49: Oh, is this where the "all" bucket is skipped instead of getting output with the other distributions?

tests/libs/build_api_v2_test.py
```diff
@@ -195,3 +202,36 @@ def test_generate_timeseries_for_fips(nyc_region, nyc_rt_dataset, nyc_icu_datase
     # Double checking that serialized json does not contain NaNs, all values should
     # be serialized using the simplejson wrapper.
     assert "NaN" not in region_timeseries.json()
+
+
+def test_multiple_distributions():
+    """All time-series within a variable are treated as a unit when combining"""
+    vaccines_administered = FieldName("vaccines_administered")
```
TomGoBravo, line 209: Use CommonFields instead making a new FieldName

tests/libs/datasets/timeseries_test.py
```diff
@@ -1738,6 +1742,19 @@ def test_combine_demographic_data_multiple_distributions():
     test_helpers.assert_dataset_like(combined, ds_expected)
 
 
+def test_bucketed_latest_missing_location_id(nyc_region: Region):
+    dataset = test_helpers.build_default_region_dataset({CommonFields.CASES: [1, 2, 3]})
+    # nyc_region = Region.from_fips("97222")
+    output = dataset._bucketed_latest_for_location_id(nyc_region.location_id)
+    expected = pd.DataFrame(
```
TomGoBravo, line 1749: is `assert output.empty` enough to test that it did what we want?

libs/build_api_v2.py
```diff
@@ -96,7 +135,7 @@ def build_region_summary(
         riskLevels=risk_levels,
         lastUpdatedDate=datetime.utcnow(),
         locationId=region.location_id,
-        url=latest_values[CommonFields.CAN_LOCATION_PAGE_URL],
+        url=latest_values.get(CommonFields.CAN_LOCATION_PAGE_URL),
```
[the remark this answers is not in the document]
ghop02: basically I wanted to write testing functions on a subset of the data - and of the direct access prevents from writing targeted testing... I've probably advocated against this in the past, but since we have the test on real data i feel somewhat comfortable?

libs/build_api_v2.py
ghop02: Ah i should revert this

libs/build_api_v2.py
ghop02: that's not included in the demographic specific data - essentially the normal `actuals` data is all of the all data

libs/build_api_v2.py
```diff
@@ -35,17 +39,52 @@
 USA_VACCINATION_START_DATE = datetime(2020, 12, 14)
 
 
-def _build_actuals(actual_data: dict) -> Actuals:
+_logger = structlog.get_logger()
+
+
+def _filter_demographic_field_and_remove_prefix(demographic_field: str, data: Dict):
+    return {
+        key.replace(demographic_field + ":", ""): value
```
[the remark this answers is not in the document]
ghop02: yeah I guess I could make a function that outputs something similar to the structure of the demographicdistributions and then just map from distribution name to the key

libs/datasets/timeseries.py
```diff
@@ -612,15 +623,30 @@ def timeseries_not_bucketed_wide_dates(self) -> pd.DataFrame:
...
         # timeseries is already sorted by DATE with the latest at the bottom.
         long = self.timeseries.stack().droplevel(CommonFields.DATE)
         # `long` has MultiIndex with LOCATION_ID and VARIABLE (added by stack). Keep only the last
         # row with each index to get the last value for each date.
         unduplicated_and_last_mask = ~long.index.duplicated(keep="last")
+
         return long.loc[unduplicated_and_last_mask, :].unstack()
 
+    @cached_property
+    def _timeseries_bucketed_latest_values(self) -> pd.DataFrame:
```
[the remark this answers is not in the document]
ghop02: done!

tests/libs/datasets/timeseries_test.py
ghop02: nope :) I'll add another test

libs/datasets/demographics.py
```diff
@@ -0,0 +1,49 @@
+from typing import Optional, List, NewType, Dict
+import dataclasses
+
+import pandas as pd
+
+
+Distribution = NewType("Distribution", Dict[str, float])
```
TomGoBravo, line 7: ScalarDistribution to differentiate from the things in a MultiRegionDataset which are timeseries distributions.

TomGoBravo, line 7: I think the str is just a bucket name, not DistributionBucket. Add comment to clarify.

libs/datasets/sources/can_scraper_helpers.py
```diff
@@ -87,14 +88,8 @@ def _fips_from_int(param: pd.Series):
 
 def make_short_name(row: pd.Series) -> str:
     """Transform a Series of demographic values into a single string such as 'age:0-9;sex:female'"""
```
TomGoBravo, line 90: update doc string

libs/datasets/demographics.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@dataclasses.dataclass
```
TomGoBravo, line 10: make it frozen.

libs/datasets/demographics.py
```diff
@@ -4,10 +4,11 @@
 import pandas as pd
 
 
-Distribution = NewType("Distribution", Dict[str, float])
+# Key for ScalarDistributions are just bucket names, such as "10-19" in a distribution for age.
+ScalarDistribution = NewType("Distribution", Dict[str, float])
```
TomGoBravo, line 8: update `"Distribution"` to match type name.

TomGoBravo (approved):

PR #1027, opened 2021-03-17
adds vaccine_backfills bucket support

TomGoBravo:
This PR addresses https://trello.com/c/lvRFxC5y/932-get-dei-vaccination-data-into-api

## Tested
    

ghop02 (approved):
snazzy!

PR #1036, opened 2021-03-22
Migrate hhs hospital importer to covid-data-model.

mikelehen:
Tested via `./run.py data update` and `git difftool`.  The resulting multiregion dataset files were identical except for some rounding errors and the (desired) inclusion of source url / name information.

### DIFF
--- a/libs/datasets/sources/hhs_hospital_dataset.py
+++ b/libs/datasets/sources/hhs_hospital_dataset.py
@@ -26,7 +26,7 @@ CUSTOM_START_DATES = {
     "53": "2020-10-25",  # Washington
 }
 
-# Mapping of CAN scraper fields to common fields.
+# Mapping of ScraperVariable.variable_name to common fields.
 FIELD_MAPPING = {
     "adult_icu_beds_capacity": CommonFields.ICU_BEDS,
     "adult_icu_beds_in_use": CommonFields.CURRENT_ICU_TOTAL,
@@ -104,15 +104,13 @@ def modify_dataset(ds: MultiRegionDataset) -> MultiRegionDataset:
 def filter_early_data(df):
     """Filters out early data for each location based on DEFAULT_START_DATE and
     CUSTOM_START_DATES."""
-    keep_rows = df.index.get_level_values(CommonFields.DATE.value) >= pd.to_datetime(
-        DEFAULT_START_DATE
-    )
+    keep_rows = df.index.get_level_values(CommonFields.DATE) >= pd.to_datetime(DEFAULT_START_DATE)
     df = df.loc[keep_rows]
 
     for (fips, start_date) in CUSTOM_START_DATES.items():
         location_id = pipeline.fips_to_location_id(fips)
-        keep_rows = (df.index.get_level_values(CommonFields.LOCATION_ID.value) != location_id) | (
-            df.index.get_level_values(CommonFields.DATE.value) >= pd.to_datetime(start_date)
+        keep_rows = (df.index.get_level_values(CommonFields.LOCATION_ID) != location_id) | (
+            df.index.get_level_values(CommonFields.DATE) >= pd.to_datetime(start_date)
         )
         df = df.loc[keep_rows]
 
--- a/libs/datasets/sources/hhs_hospital_dataset.py
+++ b/libs/datasets/sources/hhs_hospital_dataset.py
@@ -49,11 +49,11 @@ def make_hhs_variable(can_scraper_field, common_field, measurement):
     )
 
 
-# NOTE: HHSHospitalStateDataset and HHSHospitalCountyDataset must be separate classes because they
-# use different measurements ("current" vs "rolling_average_7_day") on the same variables (e.g.
-# "adult_icu_beds_capacity"). CanScraperBase operates on a per-variable basis (instead of
-# per-variable+measurement) and so will not merge them properly if we try to use both in the same
-# dataset.
+# TODO(michael): HHSHospitalStateDataset and HHSHospitalCountyDataset must be separate classes
+# because they both have ScraperVariables that target the same CommonFields. This happens becaues
+# the state level and county level data use the same variable names but different measurements
+# ("current" and "rolling_average_7_day"). This causes CanScraperBase to overwrite itself as it's
+# extracting the variables. Using separate Dataset classes solves this.
 
 
 class HHSHospitalStateDataset(data_source.CanScraperBase):
--- a/libs/datasets/sources/hhs_hospital_dataset.py
+++ b/libs/datasets/sources/hhs_hospital_dataset.py
@@ -1,24 +1,119 @@
+import dataclasses
 from functools import lru_cache
+from libs import pipeline
+
 from covidactnow.datapublic.common_fields import CommonFields
+import pandas as pd
+from libs.datasets.sources import can_scraper_helpers as ccd_helpers
 from libs.datasets import data_source
 from libs.datasets.timeseries import MultiRegionDataset
 
 
-class HHSHospitalDataset(data_source.DataSource):
-    SOURCE_TYPE = "HHSHospital"
+# Early data is noisy due to lack of reporting, etc. so we just trim it off.
+DEFAULT_START_DATE = "2020-09-01"
+CUSTOM_START_DATES = {
+    "02": "2020-10-06",  # Alaska
+    "04": "2020-09-02",  # Arizona
+    "15": "2020-10-10",  # Hawaii
+    "16": "2020-10-17",  # Idaho
+    "19": "2020-09-05",  # Iowa
+    "21": "2020-10-15",  # Kentucky
+    "28": "2020-11-11",  # Mississippi
+    "34": "2020-09-01",  # New Jersey
+    "38": "2020-11-03",  # North Dakota
+    "46": "2020-11-02",  # South Dakota
+    "47": "2020-09-20",  # Tennessee
+    "53": "2020-10-25",  # Washington
+}
 
-    COMMON_DF_CSV_PATH = "data/hospital-hhs/timeseries-common.csv"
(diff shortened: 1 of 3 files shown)


libs/datasets/sources/hhs_hospital_dataset.py
```diff
@@ -1,24 +1,119 @@
...
+        unit="beds",
+        common_field=common_field,
+    )
+
+
+# TODO(michael): HHSHospitalStateDataset and HHSHospitalCountyDataset must be separate classes
+# because they both have ScraperVariables that target the same CommonFields. This happens becaues
+# the state level and county level data use the same variable names but different measurements
+# ("current" and "rolling_average_7_day"). This causes CanScraperBase to overwrite itself as it's
+# extracting the variables. Using separate Dataset classes solves this.
```
mikelehen, line 56: The problematic code is here: https://github.com/covid-projections/covid-data-model/blob/a4f365631ca0770cfca8a73bcb986f7514521442/libs/datasets/sources/can_scraper_helpers.py#L222

If you think it's worth trying to make it handle this, let me know, though it may introduce some ambiguity (which takes precedence if multiple `ScraperVariable`s that target the same `CommonField` have valid data?).  By treating them as separate datasets, we can use our combined datasets priority merging and there's no ambiguity.  🤷

libs/datasets/sources/hhs_hospital_dataset.py
```diff
@@ -1,24 +1,119 @@
...
+    """Post-processes the can scraper data as necessary."""
+    ts_copy = ds.timeseries.copy()
+
+    # Filter out some of the early noisy data.
+    filtered_ds = dataclasses.replace(
+        ds, timeseries=filter_early_data(ts_copy), timeseries_bucketed=None
+    )
+
+    # Ensure ICU_BEDS is included in the static fields as well.
+    return filtered_ds.latest_in_static(field=CommonFields.ICU_BEDS)
```
mikelehen, line 101: FWIW- having to explicitly move/copy the latest timeseries data into static data was a surprise to me and wasted a couple hours of building / debugging.  I wonder if CanScraperBase could / should do this automatically somehow?  Or are there cases where we don't want the latest timeseries data automatically promoted to static data?

TomGoBravo: For sure a developer spending hours trying to get a new data source working is a problem. The fix I'm angling for is faster tests that cover the complex parts of our pipeline.
Since there is non-trivial code here a unittest would be nice. You could call `modify_dataset` directly from your test. Or do something like `test_query_source_url` or `test_can_scraper_returns_source_url`.
About this particular issue: long ago we copied the latest timeseries value into the static data somewhere around here but it made the "static" (maybe "scalar" is a better word) data change a lot which made understanding what is happening with it harder. Also we want to get the latest values from timeseries data after the timeseries has gone through all the filtering/aggregation etc.

libs/datasets/sources/hhs_hospital_dataset.py
```diff
@@ -1,24 +1,119 @@
...
+    "21": "2020-10-15",  # Kentucky
+    "28": "2020-11-11",  # Mississippi
+    "34": "2020-09-01",  # New Jersey
+    "38": "2020-11-03",  # North Dakota
+    "46": "2020-11-02",  # South Dakota
+    "47": "2020-09-20",  # Tennessee
+    "53": "2020-10-25",  # Washington
+}
+
+# Mapping of CAN scraper fields to common fields.
```
TomGoBravo, line 29: Why refer to ScraperVariable.variable_name as "CAN scraper fields"? "CAN scraper variable_name" is more consistent. Same for the local var name in make_hhs_variable.

libs/datasets/sources/hhs_hospital_dataset.py
TomGoBravo: There is a check for multiple elements of VARIABLES with the same `common_field` in the unit test test_can_scraper_class_unique_variable_names. Do you want this to be a run-time assert too?

libs/datasets/sources/hhs_hospital_dataset.py
ghop02: One other point about this is that the whole reason we had this in the first place is because we're getting "static" icu beds from covid care map or the latest value from HHS - since HHS data is complete I don't think we should be using Covid care map at all and we may be able to remove the static data for icu beds

libs/datasets/sources/hhs_hospital_dataset.py
mikelehen: 1. Oh, sorry. It's possible I missed that I broke a test since tests were broken in `main` when I was working on this. 😬 
2. As a general rule, yes a runtime assert sounds great.  Anything that can be asserted at runtime should be asserted at runtime, and asserts are better than tests (ideally tests will exercise the asserts and allow you to write less test code that has to be maintained separately).

libs/datasets/sources/hhs_hospital_dataset.py
mikelehen: @TomGoBravo I don't think either of the tests you suggested would have helped catch this issue, but I can try to add a general dataset test if you want.  Can you point me to a good example to follow?

@ghop02  Yeah.  I was thinking about removing covid care map data as a follow-on PR.

TomGoBravo: A tests where you compare the dataset produced by the hhs class(es) given a small bit of input to an expected dataset could verify that it produced values in `MultiRegionDataset.static`. That doesn't exercise the combined data code that snagged you but does make sure the data source is doing the right thing.
OTOH, if we can remove the dependency on  CovidCareMapBeds then we can stop doing the combined_dataset thing for static hospital/bed capacity values, have the data sources only produce timeseries and let the code downstream pick the latest value. That'd let us remove the call to `latest_in_static` and simplify what is tested.
Argh, so much code for such a simple process :-/

mikelehen (commented):
Sorry for the delay.  I made a couple tweaks.  I know a test would be nice, but I'm not psyched about writing the first test for a Dataset class and I couldn't find any examples to work off of.  Let me know if I missed something.  Thanks!

libs/datasets/sources/hhs_hospital_dataset.py
TomGoBravo: https://github.com/covid-projections/covid-data-model/blob/1af9ab5b5a4fa492a238c0e88b4e28ce5bfaff8d/tests/libs/datasets/sources/can_scraper_helpers_test.py#L168-L197
* makes some input data
* defines a class with `_get_covid_county_dataset` overridden to return the input data wrapped in a CanScraperLoader
* calls make_dataset() on the newly defined class
* makes an expected dataset
* compares the expected dataset to the dataset returned by make_dataset.

You want something very similar except the class definition will look more like that at https://github.com/covid-projections/covid-data-model/blob/1af9ab5b5a4fa492a238c0e88b4e28ce5bfaff8d/tests/libs/datasets/data_source_test.py#L51-L54 
```
class CANScraperForTest(HHSHospitalCountyDataset):
        @staticmethod
        def _get_covid_county_dataset():
            return ccd_helpers.CanScraperLoader(input_data)
```
(the method name _get_covid_county_dataset is totally irrelevant today).

Here are two more examples of setting up input data, defining a test class, calling make_dataset on the test class, checking the returned value: https://github.com/covid-projections/covid-data-model/blob/1af9ab5b5a4fa492a238c0e88b4e28ce5bfaff8d/tests/libs/datasets/data_source_test.py#L101-L150

If you aren't already using it, I strongly recommend the debugger in pycharm.

mikelehen: Thank you for the pointers, that was immensely helpful.  Sorry I didn't dig harder for these examples myself. I probably should have been able to find them. 😁   Added a couple tests that exercise the date trimming logic as well as validation that we correctly add the data to `latest`.  let me know what you think!

tests/libs/datasets/sources/hhs_hospital_dataset_test.py
```diff
@@ -0,0 +1,93 @@
...
+    icu_beds = test_helpers.TimeseriesLiteral(
+        [20, 30], source=taglib.Source(type="HHSHospitalState", url=source_url)
+    )
+    expected_ds = test_helpers.build_default_region_dataset(
+        {CommonFields.ICU_BEDS: icu_beds},
+        region=pipeline.Region.from_fips("36"),
+        start_date="2020-09-01",
+    )
+
+    expected_ds = expected_ds.latest_in_static(field=CommonFields.ICU_BEDS)
```
TomGoBravo, line 47: I have slight preference for putting constant `static={CommonFields.ICU_BEDS: 30}` in build_default_region_dataset parameters to simplify the path from what is written in the source and what is compared by assert_dataset_like.

tests/libs/datasets/sources/hhs_hospital_dataset_test.py
```diff
@@ -0,0 +1,93 @@
...
+    icu_beds = test_helpers.TimeseriesLiteral(
+        [20, 30], source=taglib.Source(type="HHSHospitalState", url=source_url)
+    )
+    expected_ds = test_helpers.build_default_region_dataset(
+        {CommonFields.ICU_BEDS: icu_beds},
+        region=pipeline.Region.from_fips("02"),
+        start_date="2020-10-06",
+    )
+
+    expected_ds = expected_ds.latest_in_static(field=CommonFields.ICU_BEDS)
```
TomGoBravo, line 91: same as above, inline constant

TomGoBravo (approved):
Nice test! Thank you for bearing with me.

tests/libs/datasets/sources/hhs_hospital_dataset_test.py
mikelehen: Good point, better to be explicit in the test!

PR #1037, opened 2021-03-23
Truncate old dates. Add test_helpers.read_csv_str.

TomGoBravo:
* Expands `DataSource._check_data` to truncate data from before 2020-01-01 and rename it to `_check_and_removed_unexpected_data`
* Expands DataSource.IGNORED_FIELDS to include more fields that are okay to be ignored now that geo data is loaded from elsewhere
* Adds `test_helpers.read_csv_str` to produce data for testing _check_data and use it in timeseries_test

This PR includes an updated `multiregion-wide-dates.csv` because the current main has data going back to 1970 and uses > 9GB of memory.

## Tested

Ran `./run.py data update`, saw the log output of data being dropped and checked for sensible header row in `data/multiregion-wide-dates.csv`

### DIFF
--- a/libs/datasets/data_source.py
+++ b/libs/datasets/data_source.py
@@ -42,7 +42,14 @@ class DataSource(object):
     # Fields that are ignored when warning about missing and extra fields. By default some fields
     # that contain redundant information about the location are ignored because cleaning them up
     # isn't worth the effort.
-    IGNORED_FIELDS = (CommonFields.COUNTY, CommonFields.COUNTRY, CommonFields.STATE)
+    IGNORED_FIELDS = (
+        CommonFields.COUNTY,
+        CommonFields.COUNTRY,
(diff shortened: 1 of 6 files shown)


libs/datasets/data_source.py
```diff
@@ -51,6 +58,21 @@ def source_tag(cls) -> taglib.Source:
 
     @classmethod
     def _check_data(cls, data: pd.DataFrame):
```
ghop02, line 60: I think `_check_data` is a bit misleading - I don't think I would expect this to modify the data - maybe `_check_and_remove_unexpected_data` would be a better name?

ghop02 (approved):
One naming comment, but otherwise looks good!

libs/datasets/data_source.py
TomGoBravo: Done

PR #1041, opened 2021-03-24
Add Google conversion event + hubspot pixel

ghop02:


api/docs/static/scripts/gtag_conversion_event.js
```diff
@@ -1,4 +1,17 @@
-// Track page view conversion event. 
+// Track page view conversion event.
 // Default gtag plugin doesn't support conversion events - this
 // script is included in head to track conversion events.
 gtag("event", "conversion", { send_to: "AW-527465415/lRyqCM2a9-kBEMf3wfsB" });
+
+function gtag_report_conversion(url) {
```
mikelehen, line 6: Not sure what's up with the `url` param and we don't seem to use it? 

Maybe rename and/or add a comment to explain what this is for.

mikelehen (approved):

PR #1042, opened 2021-03-25
Fix case where values for a bucket dont exist

ghop02:
https://sentry.io/organizations/covidactnow/issues/2295671173/?project=5203044&referrer=slack was a bit surprising to me, but it turns out that because latest is stored in a wide variables way, it somewhat pollutes the bucket name space.

The proper fix is to make `_timeseries_bucketed_latest_values` work from wide dates, but this is a quicker fix that should be alright for now.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -180,7 +180,7 @@ class OneRegionTimeseriesDataset:
                 bucket = DistributionBucket.from_str(short_name)
                 result[field][bucket.distribution][bucket.name] = value
 
-        return dict(result)
+        return result
 
     @cached_property
     def tag_objects_series(self) -> pd.Series:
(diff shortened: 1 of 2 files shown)


libs/datasets/timeseries.py
```diff
@@ -157,20 +157,30 @@ def sources_all_bucket(self, field_name: FieldName) -> List[taglib.Source]:
...
+                # variables with different buckets for a given distribution, including none's
+                # will mix buckets.  To properly fix, consider passing latest in a long format
+                # rather than a wide variables format.
+                if value is None:
+                    continue
                 bucket = DistributionBucket.from_str(short_name)
                 result[field][bucket.distribution][bucket.name] = value
 
-        return result
+        return dict(result)
```
TomGoBravo, line 183: Please add concern my shirt why you need to make a non default dict.

TomGoBravo (approved):

PR #1043, opened 2021-03-25
Updating descriptions for vaccination demographic data

ghop02:
This updates docs for vaccination demographic data.  Best two files to review are `api/can_api_v2_definition.py` and `api/docs/docs/updates.md`.  The rest is autogenerated.

`./run.py api update-sources`


api/can_api_v2_definition.py
```diff
@@ -61,6 +61,13 @@ class NonCovidPatientsMethod(enum.Enum):
 
 
 class DemographicDistributions(base_model.APIBaseModel):
+    """Distributions of demographic data.
+
+    Note that because different regions may have different demographic distributions for
```
TomGoBravo, line 66: rm "because"

api/can_api_v2_definition.py
```diff
@@ -61,6 +61,13 @@ class NonCovidPatientsMethod(enum.Enum):
...
+    the same field.  For instance, how age is collected is not uniform.
```
TomGoBravo, line 67: 'collected' makes me think of how ages are record from individuals but this is likely a decision made during some aggregation stage. How about "For instance, the age ranges are not uniform." ?

api/can_api_v2_definition.py
```diff
@@ -61,6 +61,13 @@ class NonCovidPatientsMethod(enum.Enum):
...
+
+    The data provided matches the source distributions.
```
TomGoBravo, line 69: We may filter the values so the values may not match our upstream sources exactly. How about "Distributions match those provided by upstream data sources."

TomGoBravo (approved):

PR #1045, opened 2021-03-25
Add additional checks on API infra deploy

ghop02:
This adds in a couple sanity checks before and after the lambda deploy.

I think this could potentially belong in a separate step, but because they are lightweight and broad checks, it doesn't feel like it steps on the toes of other types of testing.  

Checks added:
 * Unit test to load config object - this protects against a bad .env file
 * Simple request checking that api endpoints work


mikelehen (approved):
Nice, looks great!

PR #1046, opened 2021-03-25
Submit hubspot form instead of normal contact creation

ghop02:
This is a first step to improving the signup form - we will be using a hubspot form to pass additional signup data through.  

This will then be added to the contact created in hubspot.

As we add in more data that will be passed up through the sign up form, we can add more properties here

### DIFF
--- a/api/awsauth/awsauth/hubspot_client.py
+++ b/api/awsauth/awsauth/hubspot_client.py
@@ -1,3 +1,4 @@
+from typing import Optional
 import requests
 from awsauth.config import Config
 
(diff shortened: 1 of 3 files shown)


api/awsauth/awsauth/hubspot_client.py
```diff
@@ -7,7 +8,7 @@ class HubSpotAPICallFailed(Exception):
 
 
 class HubSpotClient:
-    def create_contact(self, email: str) -> dict:
+    def create_contact(self, email: str) -> Optional[dict]:
```
TomGoBravo, line 11: I don't see other create_contact calls. Can it be deleted?

TomGoBravo (approved):

api/awsauth/awsauth/hubspot_client.py
ghop02: yeah I'll go ahead and delete - almost went ahead and deleted it

PR #1048, opened 2021-03-29
Adding hubspot token to form

ghop02:
This lets us capture the hubspot token cookie and pass to form submission.  

Also adds in a bit of refactoring - when the more complicated reg form is added there will be a separate endpoint for retrieving an existing token.  This breaks up the `_get_api_key` and `_create_new_user` functions and adds a small class to parse/handle errors for input

### DIFF
--- a/api/awsauth/awsauth/auth_app.py
+++ b/api/awsauth/awsauth/auth_app.py
@@ -1,4 +1,4 @@
-from typing import Tuple, Dict, Optional
+from typing import Dict, Optional
 import uuid
 import dataclasses
 import urllib.parse
@@ -151,7 +151,9 @@ def _create_new_user(args: RegistrationArguments) -> str:
 
     # attempt to add hubspot contact, but don't block reg on failure.
     try:
-        registry.hubspot_client.submit_reg_form(email)
+        registry.hubspot_client.submit_reg_form(
+            email, hubspot_token=args.hubspot_token, page_uri=args.page_uri
+        )
     except hubspot_client.HubSpotAPICallFailed:
         _logger.error("HubSpot call failed")
         sentry_sdk.capture_exception()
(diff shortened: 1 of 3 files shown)


api/awsauth/awsauth/auth_app.py
```diff
@@ -200,16 +239,21 @@ def register_edge(event, context):
...
-    if "email" not in data:
-        return {"status": 400, "errorMessage": "Missing email parameter", "headers": headers}
 
-    email = data["email"]
-    if not re.match(EMAIL_REGEX, email):
-        return {"status": 400, "errorMessage": "Invalid email", "headers": headers}
+    try:
+        args = RegistrationArguments.make_from_json_body(data)
+    except ValueError as exn:
+        return {"status": 400, "errorMessage": str(exn), "headers": headers}
```
TomGoBravo, line 246: dumping str(exn) into a message that may be returned to the client seems way more dangerous than a static string. I don't know much about the context here or security but enough to be paranoid.  I'd make an exception type per message or have some attribute of an exception that is known to be safe.

api/docs/src/components/SignupForm.jsx
```diff
@@ -12,6 +12,19 @@ import {
...
+const HUBSPOT_COOKIE_NAME = "hubspotutk";
+
+// Taken from
+// https://gist.github.com/joduplessis/7b3b4340353760e945f972a69e855d11
+function getCookie(name) {
+  const value = "; " + document.cookie;
+  const parts = value.split("; " + name + "=");
+
+  if (parts.length == 2) {
+    return parts.pop().split(";").shift();
```
TomGoBravo, line 24: Did you see the possible improvements at https://gist.github.com/joduplessis/7b3b4340353760e945f972a69e855d11#gistcomment-3668361 ?

TomGoBravo (approved):

api/docs/src/components/SignupForm.jsx
ghop02: ah yeah - we're not yet using typescript yet :'( but planning on changing that soon

PR #1053, opened 2021-03-29
combined data gets 2 more sig fig

TomGoBravo:
This PR is part of https://trello.com/c/EyISOtVY/623-usa-data-aggregation-in-pipeline-doesnt-work

This increases combined data from about 47MB to about 48MB.

## Tested

`./run.py data update` and `git difftool --no-prompt -t vimdiff main -- data/multiregion-wide-dates.csv` confirms that lots of rows with float values changed. int values were not modified.


libs/datasets/timeseries.py
```diff
@@ -1165,8 +1165,9 @@ def write_to_dataset_pointer(self, pointer: dataset_pointer.DatasetPointer):
         """Writes `self` to files referenced by `pointer`."""
         wide_df = self.timeseries_rows()
 
-        # 7 significant digits of precision seems like enough.
-        csv_buf = wide_df.to_csv(index=True, float_format="%.7g")
+        # The number of significant digits needs to be enough that 100M vaccines are formatted as
+        # an integer and not changed to exponential format.
```
ghop02, line 1169: What is the issue with exponential format? does pandas not handle it properly?

ghop02 (approved):

libs/datasets/timeseries.py
TomGoBravo: Added comment.

PR #1062, opened 2021-03-31
Add country to the API with hacky fips='0'

TomGoBravo:
This PR addresses https://trello.com/c/dckJRBsA/574-add-usa-aggregation-to-api

Schema auto-generated files updated with `./run.py api update-schemas`

## Tested

This change has a unittest which passes. I also ran `python pyseir/cli.py build-all --generate-api-v2 --location-id-matches 'iso1:us$'` and browsed the files I found afterwards using `find output`.


libs/pipelines/api_v2_paths.py
```diff
@@ -88,7 +92,12 @@ def bulk_flattened_timeseries_data(self, file_type, state: Optional[str] = None)
     def single_summary(self, region_summary: can_api_v2_definition.RegionSummary, file_type):
         if self.level is AggregationLevel.STATE:
             return self.region_subdir / f"{region_summary.state}.{file_type.suffix}"
-        if self.level in (AggregationLevel.COUNTY, AggregationLevel.CBSA, AggregationLevel.PLACE):
+        if self.level in (
+            AggregationLevel.COUNTY,
+            AggregationLevel.CBSA,
+            AggregationLevel.PLACE,
+            AggregationLevel.COUNTRY,
```
ghop02, line 99: I think we can use the country code for the key - I think /country/US.json is clearer and potentially more stable if we end up changing the key

ghop02 (changes requested):
I'd like to change the key for the individual country but other than that looks good!

libs/pipelines/api_v2_paths.py
TomGoBravo: Done in ccfc14cb

ghop02 (approved):
This looks good to me! 🚢

PR #1064, opened 2021-04-05
read/write to paths instead of DatasetPointer

TomGoBravo:

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -867,6 +867,7 @@ class MultiRegionDataset:
 
     @staticmethod
     def read_from_pointer(pointer: dataset_pointer.DatasetPointer) -> "MultiRegionDataset":
+        # TODO(tom): Deprecate use of DatasetPointer and remove this method
         return MultiRegionDataset.from_wide_dates_csv(
             pointer.path_wide_dates()
         ).add_static_csv_file(pointer.path_static())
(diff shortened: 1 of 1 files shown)


libs/datasets/timeseries.py
```diff
@@ -853,25 +853,23 @@ def read_from_pointer(pointer: dataset_pointer.DatasetPointer) -> "MultiRegionDa
...
-            dataset = dataset.append_tag_df(tag_df)
+        ).append_tag_df(tag_df)
 
-        return dataset
+    def add_static_csv_file(self, path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionDataset":
+        static_df = pd.read_csv(path_or_buf, dtype={CommonFields.FIPS: str}, low_memory=False)
+        return self.add_static_values(static_df)
+
+    @staticmethod
+    def read_from_pointer(pointer: dataset_pointer.DatasetPointer) -> "MultiRegionDataset":
```
ghop02, line 869: Are we keeping these in the meantime?

ghop02 (approved):

libs/datasets/timeseries.py
TomGoBravo: Yes, removing them is more than I want to bite off when trying to debug something else.

PR #1065, opened 2021-04-05
More uses of manual_filter. Ignore tests when checking subclasses.

TomGoBravo:
This PR was originally motivated by  https://github.com/covid-projections/covid-data-model/pull/1065 but wasn't used to resolve it.

* adds manual_filter support for field_group
* adds datasource_regions manual_filter
* Unrelated fix for tests/libs/datasets/data_source_test.py::test_can_scraper_class_expected_fields_set by changing get_concrete_subclasses to get_concrete_subclasses_not_in_tests, exempting test subclasses from checks

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -12,7 +12,6 @@ import pandas as pd
 import structlog
 
 from covidactnow.datapublic.common_fields import CommonFields
-from covidactnow.datapublic.common_fields import FieldGroup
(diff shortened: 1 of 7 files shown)


mikelehen:
I think this looks good and would get us through the week, but I think it's introducing a test failure (though it's hard to tell since our tests aren't passing right now 😿 ).

```
FAILED tests/libs/datasets/combined_dataset_utils_test.py::test_include_exclude_regions
```

mikelehen (approved):
LGTM, thanks!

PR #1066, opened 2021-04-06
Long term fix for build_api_v2_test

TomGoBravo:
This PR adds a small stable subset of timeseries data to use in tests that have in the past used the entire combined_datasets. It build on https://github.com/covid-projections/covid-data-model/pull/1064 that made reading and writing to different file paths easier.

With this change running all the non-slow tests (as done by `make test-fast`) takes 23 seconds instead of almost a minute and the longest duration non-slow test dropped from 28 seconds to 1.1 second.

This change will prevent data updates breaking our fast tests. The slow tests can still be broken but these are generally more stable.

`./run.py data update-test-combined-data` updates the data.


tests/libs/build_api_v2_test.py
```diff
@@ -52,7 +52,11 @@ def test_build_summary_for_fips(
     risk_levels = top_level_metric_risk_levels.calculate_risk_level_from_metrics(latest_metric)
     assert latest_metric
     summary = build_api_v2.build_region_summary(fips_timeseries, latest_metric, risk_levels, log)
-    field_source_hhshospital = FieldSource(type=FieldSourceType.HHSHospital)
+    field_source_hhshospital = FieldSource(
+        name="Department of Health and Human Services",
+        type=FieldSourceType.OTHER,
+        url="https://healthdata.gov/Hospital/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/anag-cw7u",
```
mikelehen, line 58: Hrm.  Why did this change?

libs/datasets/combined_datasets.py
```diff
@@ -231,6 +231,13 @@ def load_us_timeseries_dataset(
     return MultiRegionDataset.read_from_pointer(pointer)
 
 
+@functools.lru_cache(None)
+def load_test_dataset() -> MultiRegionDataset:
+    wide_dates_path = dataset_utils.REPO_ROOT / "tests/data/test-combined-wide-dates.csv"
```
mikelehen, line 236: Could this path be in a constant somewhere and shared with update_test_combined_data() ?

mikelehen (approved):
THANK YOU!  🎉 

TBH I didn't even know about `make test-fast`.  Can you add that to https://github.com/covid-projections/covid-data-model/blob/main/SETUP.md#run-tests and also document the new `./run.py data update-test-combined-data` command?

libs/datasets/combined_datasets.py
TomGoBravo: Done

libs/datasets/combined_datasets.py
```diff
@@ -231,6 +231,13 @@ def load_us_timeseries_dataset(
     return MultiRegionDataset.read_from_pointer(pointer)
 
 
+@functools.lru_cache(None)
+def load_test_dataset() -> MultiRegionDataset:
```
ghop02, line 244: Suggestion to move this in `tests/test_helpers`. prefer to not have code only used for tests outside of the `tests` folder

TomGoBravo: https://github.com/covid-projections/covid-data-model/pull/1067

tests/libs/build_api_v2_test.py
mikelehen: Oh, thanks!  I'm not sure how I missed that. 😦

PR #1069, opened 2021-04-06
Add territories to country aggregation

TomGoBravo:
This PR addresses https://trello.com/c/idSo6rgo/1215-add-territories-to-usa-aggregation-and-investigate-any-diff-from-nyt

This PR does not fix the problem with stalled cumulative tails getting truncated on state level timeseries causing the country level aggregated timeseries having unexpectedly low or NA values for recent dates.

## Tested

```
./run.py data update
```
shows small increase in country level stats and no other changes.


cli/data.py
```diff
@@ -150,10 +150,14 @@ def update(aggregate_to_country: bool, state: Optional[str], fips: Optional[str]
     multiregion_dataset = multiregion_dataset.append_regions(cbsa_dataset)
     multiregion_dataset.print_stats("CountyToCBSAAggregator")
 
+    # TODO(tom): Add a clean to store intermediate values instead of commenting out code like this:
```
mikelehen, line 153: missing a word ("clean way"?)

cli/data.py
```diff
@@ -175,13 +179,18 @@ def aggregate_cbsa(output_path: pathlib.Path):
...
+    dataset = timeseries.MultiRegionDataset.from_wide_dates_csv(
+        pathlib.Path("data/pre-agg-wide-dates.csv")
+    ).add_static_csv_file(pathlib.Path("data/pre-agg-static.csv"))
+    dataset = region_aggregation.aggregate_regions(
+        dataset,
+        pipeline.us_states_and_territories_to_country_map(),
+        reporting_ratio_required_to_aggregate=DEFAULT_REPORTING_RATIO,
+    )
+    dataset.write_to_wide_dates_csv(
+        pathlib.Path("data/post-agg-wide-dates.csv"), pathlib.Path("data/post-agg-static.csv")
```
mikelehen, line 192: Is this code useless without the commented-out code above? Seems like it at least warrants a TODO comment explaining the state of affairs and how / when you'd use this command.

mikelehen (approved):

cli/data.py
TomGoBravo: I added a `help` attribute with some info so it appears in `./run.py data --help` and `./run.py data aggregate-states-to-country --help` and the source.

cli/data.py
TomGoBravo: Fixed.

PR #1076, opened 2021-04-12
add update_test_combined_data command line options

TomGoBravo:
I've been using a subset of data created with 

```
./run.py data update-test-combined-data --no-truncate-dates --state SD
cp tests/data/test-combined-wide-dates.csv data/multiregion-wide-dates.csv
cp tests/data/test-combined-static.csv data/multiregion-static.csv
```

to browse SD data while working on https://trello.com/c/KdcwKHwK/1241-several-south-dakota-counties-have-fully

### DIFF
--- a/cli/data.py
+++ b/cli/data.py
@@ -280,7 +280,11 @@ def update_case_based_icu_utilization_weights():
         json.dump(output, f, indent=2, sort_keys=True)
 
 
-@main.command()
+@main.command(
+    help="Regenerate the test combined data. The options can be used to produce data for "
+    "testing or experimenting with particular subsets of the entire dataset. Use "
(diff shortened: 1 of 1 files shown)


cli/data.py
```diff
@@ -281,24 +281,39 @@ def update_case_based_icu_utilization_weights():
 
 
 @main.command()
-def update_test_combined_data():
+@click.option(
+    "--truncate-dates/--no-truncate-dates",
```
ghop02, line 285: What's the motivation behind adding this functionality? Will there be multiple test outputs with different data sets of test data? 

Would be good to indicate somewhere why/when this should be used

ghop02 (approved):
Code looks fine - but not sure I understand the intent

PR #1077, opened 2021-04-12
tag vaccine backfill as derived

TomGoBravo:
This PR is part of https://trello.com/c/jeoMFnbb/1243-revisit-backfillvaccinationinitiated

* Adds taglib.Derived that for now has no attributes so is encoded as an empty JSON `"{}"`
* Adds `MultiRegionDataset.add_tag_to_subset` making it easy to add a tag to a subset of time series
* Changes how backfill_vaccination_initiated merges calculated and existing time series without changing behavior

## Tested

Ran `./run.py data update` and checked `git difftool -t vimdiff --no-prompt main -- data/multiregion-annotations.csv` only adds a column.

05eecac asserted that the old and new way of merging the backfill produced the same output. This worked with the current combined data.

### DIFF
--- a/libs/datasets/taglib.py
+++ b/libs/datasets/taglib.py
@@ -56,6 +56,7 @@ class TagType(GetByValueMixin, ValueAsStrMixin, str, enum.Enum):
     CUMULATIVE_LONG_TAIL_TRUNCATED = "cumulative_long_tail_truncated"
     ZSCORE_OUTLIER = "zscore_outlier"
     KNOWN_ISSUE = "known_issue"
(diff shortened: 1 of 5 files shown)


ghop02 (approved):

TomGoBravo:
[Revisit backfill_vaccination_initiated code (1st dose + 2nd dose !== total doses in the case of J&J).](https://trello.com/c/jeoMFnbb/1243-revisit-backfillvaccinationinitiated-code-1st-dose-2nd-dose-total-doses-in-the-case-of-jj)

PR #1080, opened 2021-04-20
apply region-overrides in data pipeline

TomGoBravo:
This PR addresses https://trello.com/c/f27O08ea/1266-make-it-possible-to-block-in-the-backend-using-the-cms

* Copies `https://raw.githubusercontent.com/covid-projections/covid-projections/develop/src/cms-content/region-overrides/region-overrides.json` to `data/region-overrides.json`
* Adds support for changing region-overrides.json into the config expected by https://github.com/covid-projections/covid-data-model/blob/main/libs/datasets/manual_filter.py
* Applies the manual filter during `data update`
* Adds `MultiRegionDataset.rm_tags_from_timeseries_subset` and uses it to drop tags when a time series is entirely dropped.
* Adds new tag type `KnownIssueNoDate`

## Tested

Ran `data update`. It modified every row in `multiregion-wide-dates.csv` so hard to see exactly what changed but a sample looked correct.


TomGoBravo:
[Make it possible to block in the backend using the CMS](https://trello.com/c/f27O08ea/1266-make-it-possible-to-block-in-the-backend-using-the-cms)

libs/datasets/manual_filter.py
```diff
@@ -96,3 +100,70 @@ def run(dataset: timeseries.MultiRegionDataset, config=CONFIG) -> timeseries.Mul
...
+    elif include_str == "region-and-subregions":
+        if region.is_state():
+            regions_included = [RegionMask(states=[region.state])]
+        elif region.level == AggregationLevel.CBSA:
+            regions_included = [region] + cbsa_to_counties_map[region]
+        else:
+            raise ValueError("region-and-subregions only valid for a state and CBSA")
+    elif include_str == "subregions":
+        if not region.is_state():
+            raise ValueError("subregions only valid for a state")
```
mikelehen, line 142: we could trivially support cbsa's too, right?  But I think you're right that we don't need to, so this seems fine. 😁

libs/datasets/manual_filter.py
```diff
@@ -96,3 +100,70 @@ def run(dataset: timeseries.MultiRegionDataset, config=CONFIG) -> timeseries.Mul
...
+        regions_included = [RegionMask(AggregationLevel.COUNTY, states=[region.state])]
+    else:
+        raise ValueError(f"Invalid include_str: {include_str}")
```
mikelehen, line 145: It'd be nice if the errors reflected the JSON syntax, i.e. "include" not "include_str"

tests/libs/datasets/manual_filter_test.py
```diff
@@ -154,3 +159,11 @@ def test_manual_filter_per_bucket_tag():
...
 
     test_helpers.assert_dataset_like(ds_out, ds_expected)
+
+
+def test_region_overrides_transform():
+    aggregator = statistical_areas.CountyToCBSAAggregator.from_local_public_data()
+    transformed = manual_filter.transform_region_overrides(
+        json.load(open(data.REGION_OVERRIDES_JSON)), aggregator.cbsa_to_counties_region_map
+    )
+    pprint.pprint(transformed)
```
mikelehen, line 169: Is this a placeholder test or intentionally meant to just be a smoke test?

libs/datasets/manual_filter.py
```diff
@@ -96,3 +100,70 @@ def run(dataset: timeseries.MultiRegionDataset, config=CONFIG) -> timeseries.Mul
         dataset = filtered_dataset.append_regions(passed_dataset)
 
     return dataset
+
+
+# Possible values from https://github.com/covid-projections/covid-projections/blob/develop/src/cms-content/region-overrides/region-overrides.json
+_METRIC_TO_FIELDS = {
+    "metrics.caseDensity": common_fields.FIELD_GROUP_TO_LIST_FIELDS[FieldGroup.CASES_DEATHS],
```
mikelehen, line 107: I think we probably want this to just be cases.  Or we could rename "metrics.caseDensity" to something else (see below), but in general I think we block cases way more than deaths and probably want to treat them separate.

libs/datasets/manual_filter.py
```diff
@@ -96,3 +100,70 @@ def run(dataset: timeseries.MultiRegionDataset, config=CONFIG) -> timeseries.Mul
...
+    # "metrics.infectionRate": common_fields.FieldGroup.
+    "metrics.testPositivityRatio": common_fields.FIELD_GROUP_TO_LIST_FIELDS[FieldGroup.TESTS],
+    "metrics.icuCapacityRatio": common_fields.FIELD_GROUP_TO_LIST_FIELDS[
+        FieldGroup.HEALTHCARE_CAPACITY
+    ],
+    "metrics.vaccinationsInitiatedRatio": common_fields.FIELD_GROUP_TO_LIST_FIELDS[
+        FieldGroup.VACCINES
+    ],
```
mikelehen, line 115: It seems like your FieldGroup classifications may be a better way of classifying data to be blocked.  Want to propose a set of stringified names we could use in the JSON and I can sanity-check them and maybe I'll move the frontend over to use the same strings (even if it doesn't implement them completely)?  Or do you have other suggestions for a schema/syntax for specifying the fields to be blocked?

mikelehen (commented):
Nice!  That was simpler and quicker than I expected.  Left some comments but this seems like it's on the right track.  Want to propose some schema changes to add start_date (ideally it'd be optional though) and perhaps rework the "metric" naming to be closer to your field groups and generally clean this up?

If there are other open questions to discuss, let me know!

libs/datasets/manual_filter.py
```diff
@@ -96,3 +100,70 @@ def run(dataset: timeseries.MultiRegionDataset, config=CONFIG) -> timeseries.Mul
...
+    region_overrides: Mapping, cbsa_to_counties_map: Mapping[Region, List[Region]]
+) -> Mapping:
+    filter_configs = []
+    for override in region_overrides["overrides"]:
+        if not override.get("blocked"):
+            continue
+        try:
+            filter_configs.append(_transform_one_override(override, cbsa_to_counties_map))
+        except:
+            raise ValueError(f"Problem with {override}")
```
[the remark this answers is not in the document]
TomGoBravo: Python 3 automatically includes the original exception from a raise within an except. https://docs.python.org/3/tutorial/errors.html#exception-chaining

libs/datasets/manual_filter.py
```diff
@@ -96,3 +100,70 @@ def run(dataset: timeseries.MultiRegionDataset, config=CONFIG) -> timeseries.Mul
...
+        region = Region.from_fips(region_str)
+    else:
+        raise ValueError(f"Invalid region: {region_str}")
+
+    include_str = override["include"]
+    if include_str == "region":
+        regions_included = [region]
+    elif include_str == "region-and-subregions":
+        if region.is_state():
+            regions_included = [RegionMask(states=[region.state])]
```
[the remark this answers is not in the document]
TomGoBravo: `RegionMask` doesn't do anything. It is literally 3 lines: a name, an optional level and optional list of states. I can't say much about its behavior in general because it is highly dependent on the context. I removed the default `level = None` so it must be always explicitly given.

libs/datasets/manual_filter.py
TomGoBravo: Done.

cli/data.py
```diff
@@ -104,6 +108,13 @@ def update(aggregate_to_country: bool, state: Optional[str], fips: Optional[str]
         timeseries_field_datasets, static_field_datasets
     )
     _logger.info("Finished combining datasets")
+
+    aggregator = statistical_areas.CountyToCBSAAggregator.from_local_public_data()
```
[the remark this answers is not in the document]
TomGoBravo: Added comment.

tests/libs/datasets/manual_filter_test.py
TomGoBravo: Just a smoke test. Renamed.

libs/datasets/manual_filter.py
TomGoBravo: I don't have that much vision for how this will be used. How about merge and iterate?

libs/datasets/manual_filter.py
TomGoBravo: Changed.

TomGoBravo:
@mikelehen what do you think about the limitation of MultiRegionDataset that it only supports tag/annotations for timeseries that have at least one real value? When a timeseries is dropped we can't leave a tag/gravestone commemorating it. In general it makes sense to drop tags for a timeseries when dropping the values for it (for example when keeping only a subset of all regions) but the region override logic wants to drop a timeseries and keep an annotation for it going down the pipeline.

libs/datasets/manual_filter.py
mikelehen: Okay.  I was mostly just thinking of using the field group names except maybe split cases/deaths, but we can punt for now.

libs/datasets/manual_filter.py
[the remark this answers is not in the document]
mikelehen: Ah, sweet! Thanks for the pointer.

mikelehen (commented):
@TomGoBravo Yeah, from an API perspective I think _ideally_ we'd be able to attach an annotation to a missing timeseries.  This comes up a lot.  E.g. we are currently blocking all Philadelphia vaccine data because we know it's garbage (though we have a new scraper that'll soon fix this case!).  So ideally rather than it just being absent in the API we'd have annotation of some kind that explains _why_ it's missing.

That said:
1. How many of our API users will actually go digging through the API annotations to figure out why a timeseries is missing?  Probably none.  
2. Even on our website we don't currently have the capability to show a disclaimer when a metric is missing.

So we can probably live with that limitation for now.

To be clear though: What about disclaimers for non-blocked timeseries?  We do surface those on the frontend today (to e.g. highlight known discrepancies like where CT state dashboard shows test positivity including antigen tests but ours is only PCR tests).  We could presumably wire it up so that gets read from region-overrides.json and shows up as an annotation in the API, but per point 1 above, I don't know that any API consumers care.  So as long as we end up having the frontend read this same JSON file and apply disclaimers directly in the frontend, we probably don't strictly need it in the API if it feels too hard / annoying.  Thoughts?

Also, in terms of getting this PR over the finish line, what are you thinking with regards to start_date?

TomGoBravo:
> @TomGoBravo Yeah, from an API perspective I think _ideally_ we'd be able to attach an annotation to a missing timeseries. This comes up a lot. E.g. we are currently blocking all Philadelphia vaccine data because we know it's garbage (though we have a new scraper that'll soon fix this case!). So ideally rather than it just being absent in the API we'd have annotation of some kind that explains _why_ it's missing.
> 
> That said:
> 
> 1. How many of our API users will actually go digging through the API annotations to figure out why a timeseries is missing?  Probably none.
> 2. Even on our website we don't currently have the capability to show a disclaimer when a metric is missing.
> 
> So we can probably live with that limitation for now.

I haven't tested it all the way through but https://github.com/covid-projections/covid-data-model/pull/1083 gets us moving towards not having that limitation.

> To be clear though: What about disclaimers for non-blocked timeseries? We do surface those on the frontend today (to e.g. highlight known discrepancies like where CT state dashboard shows test positivity including antigen tests but ours is only PCR tests). We could presumably wire it up so that gets read from region-overrides.json and shows up as an annotation in the API, but per point 1 above, I don't know that any API consumers care. So as long as we end up having the frontend read this same JSON file and apply disclaimers directly in the frontend, we probably don't strictly need it in the API if it feels too hard / annoying. Thoughts?

Adding disclaimer without blocking is pretty easy, I'll add it to this PR.

> Also, in terms of getting this PR over the finish line, what are you thinking with regards to start_date?

I want to limit the scope of this PR to supporting what is already in the overrides json.

mikelehen:
> I want to limit the scope of this PR to supporting what is already in the overrides json.

I just meant so `run.py data update` doesn't crash about not having a start_date. 😁   Anyway, once this is ready to go can you mark it ready to review and I'll take another pass?

libs/datasets/manual_filter.py
```diff
@@ -96,3 +99,73 @@ def run(dataset: timeseries.MultiRegionDataset, config=CONFIG) -> timeseries.Mul
...
     return dataset
+
+
+# Possible values from
+# https://github.com/covid-projections/covid-projections/blob/develop/src/cms-content/region-overrides/region-overrides.json
+_METRIC_TO_FIELDS = {
+    "metrics.caseDensity": [CommonFields.CASES, CommonFields.NEW_CASES],
+    # "metrics.infectionRate" is not supported because it isn't used in the current
+    # region-overrides.json and doesn't exist as a field in the pipeline where the manual_filter
+    # is applied. Removing cases will likely cause Rt to no longer be produced for a region.
```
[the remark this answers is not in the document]
TomGoBravo: Done

cli/data.py
```diff
@@ -314,7 +316,7 @@ def update_test_combined_data(truncate_dates: bool, state: List[str]):
     # Keep only a small subset of the regions so we have enough to exercise our code in tests.
     test_subset = us_dataset.get_regions_subset(
         [
-            RegionMask(states=[s.strip() for s in state]),
+            RegionMask(level=None, states=[s.strip() for s in state]),
```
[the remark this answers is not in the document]
TomGoBravo: Added comments in RegionMask.

TomGoBravo:
[Make it possible to block in the backend using the CMS](https://trello.com/c/f27O08ea/1266-make-it-possible-to-block-in-the-backend-using-the-cms)

libs/datasets/taglib.py
```diff
@@ -293,6 +294,23 @@ def content(self) -> str:
         return json.dumps(d, separators=(",", ":"))
 
 
+@dataclass(frozen=True)
+class KnownIssueNoDate(TagInTimeseries):
+    disclaimer: str
```
mikelehen, line 299: Should we be internally consistent with the codebase and call this "public_note" (we can consider renaming it in the json file)?

libs/datasets/taglib.py
```diff
@@ -293,6 +294,23 @@ def content(self) -> str:
...
+    TAG_TYPE = TagType.KNOWN_ISSUE_NO_DATE
+
+    @classmethod
+    def make_instance(cls, *, content: str) -> "TagInTimeseries":
+        content_parsed = json.loads(content)
+        return cls(disclaimer=content_parsed["disclaimer"])
+
+    @property
+    def content(self) -> str:
+        d = {"disclaimer": self.disclaimer}
```
mikelehen, line 310: here too?  Actually, does this end up showing up in the API?  Do we already use "disclaimer" for `KnownIssue`?  Should we switch them both to "public_note"?  Not sure...  WDYT?

mikelehen (approved):
LGTM with minor nits. Only somewhat substantial question is whether we should migrate "disclaimer" => "public_note".  The name "disclaimer" was based on the frontend component that consumes it, but I like "public_note"...

libs/datasets/taglib.py
TomGoBravo: `disclaimer` is already used in `KnownIssue`. Agreed that public_note is better. I changed KnownIssueNoDate now and added a TODO for KnownIssue because changing it seems out of scope for this PR.

libs/datasets/manual_filter.py
```diff
@@ -120,8 +136,84 @@ def run(
...
+                filter_.tag, ts_selected_fields.index
+            )
 
         dataset = filtered_dataset.append_regions(passed_dataset)
 
     return dataset
+
+
+# Possible values from
+# https://github.com/covid-projections/covid-projections/blob/develop/src/cms-content/region-overrides/region-overrides.json
```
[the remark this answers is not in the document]
TomGoBravo: Added relative URL since the JSON will be in the same repo.

libs/datasets/timeseries.py
```diff
@@ -784,6 +784,12 @@ def add_tag_to_subset(
         ).reset_index()
         return self.append_tag_df(tag_df)
 
+    def rm_tags_from_timeseries_subset(self, index: pd.MultiIndex) -> "MultiRegionDataset":
```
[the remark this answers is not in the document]
TomGoBravo: Done

data/region-overrides.json
```diff
@@ -0,0 +1,106 @@
...
+      "context": "https://trello.com/c/oHUFzfnb/1282-disabled-large-outlier-in-mobile-county-al-daily-new-cases"
+    },
+    {
+      "include": "region",
+      "metric": "metrics.infectionRate",
+      "blocked": true,
+      "region": "33660",
+      "context": "https://trello.com/c/oHUFzfnb/1282-disabled-large-outlier-in-mobile-county-al-daily-new-cases"
+    }
+  ]
```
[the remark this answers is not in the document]
TomGoBravo: I thought I had. Last modified 3 days ago in develop and main according to https://github.com/covid-projections/covid-projections/tree/develop/src/cms-content/region-overrides  . Is there a different branch to look at?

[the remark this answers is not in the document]
mikelehen: Gah, yikes!  It looks like I forgot to push my branch this morning before shipping the snapshot. 😞 

This is fine as-is I guess. I can update it later.  Else the version in develop will be updated in a couple minutes.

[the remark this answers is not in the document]
TomGoBravo: Opps! I got the new one.

PR #1084, opened 2021-04-21
Add CANScraperCountyProviders and use for vaccine data.

mikelehen:
We are currently only ingesting provider='state' data, but Trevor wrote a Philadelphia-specific scraper (provider='county') to fix a known issue we're getting from the PA state scraped data (provider='state') for Philadelphia.

This PR makes us prefer 'county' data over 'state' where it exists.

Tested via `run.py data update` and `git difftool`.  Only Philadelphia county (and metro) vaccine data and was changed.


libs/datasets/sources/can_scraper_providers.py
```diff
@@ -3,137 +3,149 @@
...
+class CANScraperStateProviders(data_source.CanScraperBase):
+    SOURCE_TYPE = "CANScrapersStateProviders"
+
+    VARIABLES = make_scraper_variables("state")
+
+
+class CANScraperCountyProviders(data_source.CanScraperBase):
+    SOURCE_TYPE = "CANScrapersCountyProviders"
+
+    VARIABLES = make_scraper_variables("county")
```
TomGoBravo, line 151: Once upon a time I had the hope that we'll keep the list of variables here in sync with the variables available in the parquet file to reduce the chance of missed typos and accidentally dropped variables. I'd like to keep moving in that direction (in other words county and state providers have independent VARIABLES) or move towards some kind of common map from parquet file to CommonField (so the big list of all possible ScraperVariable is shared between all CanScraperBase subclasses).

mikelehen: Hrm. I'd rather not use this PR as an opportunity to do a large refactor and I don't think I see the value in duplicating the variables between the county and state providers rather than re-using them via a local helper.  Is there another option you're suggesting?

mikelehen: (in particular, there's no reason to treat state / county dashboards differently in terms of what variables they might have)

TomGoBravo: By that logic I don't think there is a reason for any CanScraperBase subclass to have different variables. Add some TODO to clean this up which we won't bother doing until something breaks :-/

TomGoBravo (approved):

libs/datasets/sources/can_scraper_providers.py
mikelehen: The distinction is that these represent local dashboards that could provide any variables, opposed to e.g. HHSHopsitalStateDataset which we want to only contain HHS hospital data but not e.g. the HHS testing data.

I did realize that renaming can_scraper_state_providers.py to can_scraper_providers.py was probably over-generalized since I guess we consider any data coming from the parquet file as "can_scraper" data, even though most of it isn't scraped per se.  So I've renamed this to can_scraper_local_dashboard_providers.py to be a bit clearer hopefully.

Also added TODO comment per your request.

PR #1085, opened 2021-04-21
Refactor drop_observations

TomGoBravo:
This PR is part of https://trello.com/c/f27O08ea/1266-make-it-possible-to-block-in-the-backend-using-the-cms

Most of these changes are done to simplify drop_observations because https://github.com/covid-projections/covid-data-model/pull/1080 uses different subsets of the code.

* changes manual_filter config to use a pydantic model instead of unstructured JSON
* replaces MultiRegionDataset.remove_regions with MultiRegionDataset.partition_by_region
* adds MultiRegionDataset.replace_timeseries_wide_dates which is used in the new manual filter code and replaces some funky use of dataclasses.replace in ca_vaccination_backfill.py
* adds RegionMask comments
* changes KnownIssue.date to a datetime instead of panda timestamp, for better compatibility with pydantic

## Tested

`/run.py data update` reproduced identical combined data.

### DIFF
--- a/libs/datasets/manual_filter.py
+++ b/libs/datasets/manual_filter.py
@@ -19,21 +19,9 @@ from libs.pipeline import RegionMaskOrRegion
 _logger = structlog.getLogger()
 
 
-# TODO(tom): Make some kind of hierarchy of class to form a schema that can be populated by a JSON.
-CONFIG = {
-    "filters": [
-        {
-            "regions_included": [RegionMask(AggregationLevel.COUNTY, states=["OK"])],
-            "regions_excluded": [Region.from_fips("40109"), Region.from_fips("40143")],
-            "observations_to_drop": {
-                "start_date": "2021-03-15",
-                "drop_fields": [CommonFields.CASES, CommonFields.DEATHS],
-                "internal_note": "https://trello.com/c/HdAKfp49/1139",
-                "public_note": "Something broke with the OK county data.",
-            },
-        },
-    ]
-}
+# The manual filter config is defined with pydantic instead of standard dataclasses to get support
+# for parsing from json, automatic conversion of fields to the expected type and fields that default
+# to None. pydantic reserves the field name 'fields'.
 
 
 class ObservationsToDrop(pydantic.BaseModel):
@@ -64,11 +52,27 @@ class Config(pydantic.BaseModel):
     filters: List[Filter]
 
 
+CONFIG = Config(
+    filters=[
+        Filter(
+            regions_included=[RegionMask(AggregationLevel.COUNTY, states=["OK"])],
+            regions_excluded=[Region.from_fips("40109"), Region.from_fips("40143")],
+            observations_to_drop=ObservationsToDrop(
+                start_date="2021-03-15",
+                drop_fields=[CommonFields.CASES, CommonFields.DEATHS],
+                internal_note="https://trello.com/c/HdAKfp49/1139",
+                public_note="Something broke with the OK county data.",
+            ),
+        )
+    ]
+)
+
+
 def _partition_by_fields(
-    dataset: timeseries.MultiRegionDataset, fields: List[CommonFields]
+    ts_in: pd.DataFrame, fields: List[CommonFields]
 ) -> Tuple[pd.DataFrame, pd.DataFrame]:
-    """Returns a tuple of the timeseries in fields and all others, in wide dates DataFrames."""
-    ts_in = dataset.timeseries_bucketed_wide_dates
+    """Partitions a DataFrame with timeseries wide dates into two: fields and others."""
+    timeseries.check_timeseries_wide_dates_structure(ts_in)
     mask_selected_fields = ts_in.index.get_level_values(PdFields.VARIABLE).isin(fields)
     ts_selected_fields = ts_in.loc[mask_selected_fields]
     ts_not_selected_fields = ts_in.loc[~mask_selected_fields]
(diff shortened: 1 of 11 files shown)


libs/datasets/manual_filter.py
```diff
@@ -10,88 +14,117 @@
...
+    timeseries.check_timeseries_wide_dates_structure(ts_in)
+    start_date = pd.to_datetime(drop_start_date)
+    obsv_selected = ts_in.loc[:, ts_in.columns >= start_date]
     mask_has_real_value_to_drop = obsv_selected.notna().any(1)
-    ts_to_filter = ts_selected_fields.loc[mask_has_real_value_to_drop]
-    ts_no_real_values_to_drop = ts_selected_fields.loc[~mask_has_real_value_to_drop]
+    ts_to_filter = ts_in.loc[mask_has_real_value_to_drop]
+    ts_no_real_values_to_drop = ts_in.loc[~mask_has_real_value_to_drop]
+    ts_filtered = ts_to_filter.loc[:, ts_to_filter.columns < start_date]
+    return ts_filtered, ts_no_real_values_to_drop
```
mikelehen, line 88: Minor but it seems like rather than partitioning the data and returning both we could just filter the data and return the result and `mask_has_real_value_to_drop`, and both this code and the calling code might be simpler.  Feel free to leave it though.

libs/datasets/manual_filter.py
```diff
@@ -10,88 +14,117 @@
...
-        new_tags.add(
-            taglib.KnownIssue(date=drop_start_date, disclaimer=config["public_note"]),
-            location_id=location_id,
-            variable=variable,
-            bucket=bucket,
-        )
+def drop_observations(
+    dataset: timeseries.MultiRegionDataset, filter_: Filter
+) -> timeseries.MultiRegionDataset:
+    """Drops observations according to `config` from every region in `dataset`."""
```
mikelehen, line 94: `config` => `filter`

libs/pipeline.py
```diff
@@ -183,7 +183,11 @@ def get_state_region(self) -> "Region":
 @final
 @dataclass(frozen=True)
 class RegionMask:
+    """Represents attributes which may be used to select a subset of regions."""
+
+    # A level (county, state, ...) OR None to select regions ignoring their level.
```
mikelehen, line 188: Thanks!

tests/libs/datasets/manual_filter_test.py
```diff
@@ -11,34 +12,36 @@
...
+                "regions_included": [RegionMask(AggregationLevel.COUNTY, states=["OK"])],
+                "regions_excluded": [Region.from_fips("40109"), Region.from_fips("40143")],
                 "start_date": "2021-03-15",
-                "field_group": FieldGroup.CASES_DEATHS,
+                "fields_included": common_fields.FIELD_GROUP_TO_LIST_FIELDS[
+                    FieldGroup.CASES_DEATHS
+                ],
                 "internal_note": "https://trello.com/c/HdAKfp49/1139",
                 "public_note": "Something broke with the OK county data.",
+                "action": manual_filter.Action.DROP_OBSERVATIONS,
```
mikelehen, line 40: Should have an ANNOTATE test if we're implementing it...

libs/datasets/manual_filter.py
```diff
@@ -10,88 +14,117 @@
...
+# dreaded "TypeError: non-default argument ...". pydantic reserves the field name 'fields'.
 
 
-def drop_observations(
-    dataset: timeseries.MultiRegionDataset, config
-) -> timeseries.MultiRegionDataset:
-    """Drops observations according to `config` from every region in dataset."""
-    drop_start_date = pd.to_datetime(config["start_date"])
+class Action(enum.Enum):
+    DROP_OBSERVATIONS = enum.auto()
```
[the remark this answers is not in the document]
mikelehen: Oh, actually... it looks like ANNOTATE isn't used or implemented...  Is that an oversight or a TODO?

mikelehen (approved):
LGTM except ANNOTATE seems to be unimplemented.

libs/datasets/manual_filter.py
```diff
@@ -10,88 +14,117 @@
...
+CONFIG = Config(
+    filters=[
+        Filter(
+            regions_included=[RegionMask(AggregationLevel.COUNTY, states=["OK"])],
+            regions_excluded=[Region.from_fips("40109"), Region.from_fips("40143")],
+            fields_included=[CommonFields.CASES, CommonFields.DEATHS],
+            action=Action.DROP_OBSERVATIONS,
+            start_date="2021-03-15",
+            internal_note="https://trello.com/c/HdAKfp49/1139",
+            public_note="Something broke with the OK county data.",
```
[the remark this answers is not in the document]
TomGoBravo: Done in https://github.com/covid-projections/covid-data-model/pull/1086

libs/datasets/manual_filter.py
mikelehen: Okay, feel free to ignore.  I was basically reacting to the fact that we have code that goes out of its way to split the data apart and then the calling code just immediately puts them back together.  So returning a tuple of essentially {filtered result, affected rows} seemed simpler, but I may be missing something and don't care strongly.  🚢  it!

libs/datasets/manual_filter.py
```diff
@@ -10,88 +14,117 @@
...
+
+
+class Filter(pydantic.BaseModel):
+    regions_included: List[RegionMaskOrRegion]
+    regions_excluded: Optional[List[RegionMaskOrRegion]] = None
+    fields_included: List[CommonFields]
+    action: Action
+    start_date: Optional[datetime.date] = None
+    internal_note: str
+    public_note: str
```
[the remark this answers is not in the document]
TomGoBravo: I think we still want the annotation for internal tracking, but the public note could be "". Added comment.

libs/datasets/manual_filter.py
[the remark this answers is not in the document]
TomGoBravo: Changed to boolean "drop_observations" in the filter. Support for drop_observations: False is in https://github.com/covid-projections/covid-data-model/pull/1080

PR #1087, opened 2021-04-22
move vaccine_backfills.derive_vaccine_pct from api to combined data

TomGoBravo:

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/cli/data.py
+++ b/cli/data.py
@@ -164,6 +164,8 @@ def update(aggregate_to_country: bool, state: Optional[str], fips: Optional[str]
         multiregion_dataset = multiregion_dataset.append_regions(country_dataset)
         multiregion_dataset.print_stats("aggregate_to_country")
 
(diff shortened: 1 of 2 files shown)


TomGoBravo:
[Make our vaccine backfill code smarter about how up-to-date the data is.](https://trello.com/c/tNlZtGp6/1279-make-our-vaccine-backfill-code-smarter-about-how-up-to-date-the-data-is)

mikelehen (approved):
I'm fine with this, but I could see the argument that we want `data update` to just pull / merge data from our data sources and then the rest of the pipeline to do interesting calculations (like infer one column from another column) but I'm sure we don't have a strict distinction there so this seems fine.  As you've pointed out before, it's a bit arbitrary what we do before / after the combined dataset update.

ghop02:
Yeah I think that's the similar tension - I think it's originally separated because the api generation calculated the "metrics" of which vaccine ratios were one of them.  However, since we scrape it from sources it's a bit blurrier.

PR #1090, opened 2021-04-22
Keep least stale vaccine percentage

TomGoBravo:
This PR is part of https://trello.com/c/tNlZtGp6/1279-make-our-vaccine-backfill-code-smarter

* Changes `vaccine_backfills.derive_vaccine_pct` to replace existing percentage metrics when the percentage derived from the people count has a real value on a more recent date

## Tested

I hacked in a change to write a wide-dates csv of the dataset before and after this change.
For counties in MD this adds a few days of data to `vaccinations_completed_pct` , for counties in ME this adds about 20 days of data to `vaccinations_initiated_pct`. It looks like the corresponding CBSAs had similar changes.

### DIFF
--- a/libs/datasets/vaccine_backfills.py
+++ b/libs/datasets/vaccine_backfills.py
@@ -10,8 +10,8 @@ MultiRegionDataset = timeseries.MultiRegionDataset
 
 
 def derive_vaccine_pct(ds_in: MultiRegionDataset) -> MultiRegionDataset:
-    """Returns a new dataset containing everything all the input and vaccination percentage
-    metrics derived from the corresponding non-percentage field where not already set."""
+    """Returns a new dataset with vaccination percentage metrics derived from their
+    corresponding non-percentage fields where the percentage metric is missing or less fresh."""
     field_map = {
         CommonFields.VACCINATIONS_INITIATED: CommonFields.VACCINATIONS_INITIATED_PCT,
         CommonFields.VACCINATIONS_COMPLETED: CommonFields.VACCINATIONS_COMPLETED_PCT,
@@ -19,15 +19,15 @@ def derive_vaccine_pct(ds_in: MultiRegionDataset) -> MultiRegionDataset:
 
     ts_in_all = ds_in.timeseries_bucketed_wide_dates
     ts_in_all_variable_index = ts_in_all.index.get_level_values(PdFields.VARIABLE)
-    ts_counts = ts_in_all.loc(axis=0)[ts_in_all_variable_index.isin(field_map.keys())]
-    ts_in_pcts = ts_in_all.loc(axis=0)[ts_in_all_variable_index.isin(field_map.values())]
-    ts_in_without_pcts = ts_in_all.loc(axis=0)[~ts_in_all_variable_index.isin(field_map.values())]
+    ts_in_people = ts_in_all.loc[ts_in_all_variable_index.isin(field_map.keys())]
+    ts_in_pcts = ts_in_all.loc[ts_in_all_variable_index.isin(field_map.values())]
+    ts_in_without_pcts = ts_in_all.loc[~ts_in_all_variable_index.isin(field_map.values())]
 
     # TODO(tom): Preserve provenance and other tags from the original vaccination fields.
     # ds_in_wide_dates / dataset.static.loc[:, CommonFields.POPULATION] doesn't seem to align the
     # location_id correctly so be more explicit with `div`:
     derived_pct_df = (
-        ts_counts.div(
+        ts_in_people.div(
             ds_in.static.loc[:, CommonFields.POPULATION],
             level=CommonFields.LOCATION_ID,
             axis="index",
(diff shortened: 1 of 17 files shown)


tests/libs/datasets/vaccine_backfill_test.py
```diff
@@ -118,3 +118,26 @@ def test_derive_vaccine_pct():
...
+
+def test_derive_vaccine_pct_least_stale():
+    # Test data with cumulative people counts and percentages having slightly different values so
+    # that they can be differentiated.
+    ts_metrics_in = {
+        # These two have the same freshness so the percentage is left untouched.
+        CommonFields.VACCINATIONS_INITIATED: [500, 600, 700],
+        CommonFields.VACCINATIONS_INITIATED_PCT: [51, 61, 71],
+        # The completed percentage is less fresh (NA for most recent date) so will be overwritten
+        # by a time series derived from the cumulative people count.
```
mikelehen, line 131: Thanks for the comments. Makes it nice to read the test. 😁

libs/datasets/vaccine_backfills.py
```diff
@@ -35,14 +36,30 @@ def derive_vaccine_pct(ds_in: MultiRegionDataset) -> MultiRegionDataset:
...
-    # TODO(tom): Compare derived_pct_df and ts_in_pcts to pick least stale.
-    # Drop derived percentages where ts_in_pcts already has a timeseries with the same index.
-    derived_pct_without_in_pcts = derived_pct_df.loc[~derived_pct_df.index.isin(ts_in_pcts.index)]
-    # Double check that there is no overlap between the percentages that will be added and the
-    # existing data.
-    assert ts_in_all.index.intersection(derived_pct_without_in_pcts.index).empty
+    def append_most_recent_date_index_level(df: pd.DataFrame) -> pd.DataFrame:
+        """Appends most recent date with real (not NA) value as a new index level."""
+        most_recent_date = df.apply(pd.Series.last_valid_index, axis=1)
+        return df.assign(most_recent_date=most_recent_date).set_index(
```
mikelehen, line 42: Just when I think I understand enough python syntax to get by, I see `most_recent_date=most_recent_date` and have no idea how that works. 😬

mikelehen (approved):
Looks good, thanks for tackling.  

https://covid-projections-git-mikelehen-test-1775-covidactnow.vercel.app/internal/compare?left=1774&locations=0&metric=6&right=1775&sort=3 indicates we got Maine county vaccine data back. 🎉  The other diffs look comparable to what's in snapshot 1776 out of `main` so I think we're good.

Sorry for randomizing your day with this.  I forget that what feels like a simple "if statement" in my head ends up more complicated when done via pandas on an entire dataset...  But at least this problem shouldn't bite us again.

libs/datasets/vaccine_backfills.py
TomGoBravo: This is calling a function with the key of a keyword parameter equal to a local var, actually pretty common:

```
$ find -name '*py' | xargs grep -P -r -e '\((\w+)=\1,'
./tests/libs/series_utils_test.py:    smoothed = series_utils.smooth_with_rolling_average(series=series, window=2)
./tests/libs/series_utils_test.py:    smoothed = series_utils.smooth_with_rolling_average(series=series, window=5)
./cli/data.py:                dataset = dataset.get_subset(state=state, fips=fips)
./pyseir/cli.py:    for state in _states_region_list(state=state, default=ALL_STATES):
./pyseir/models/suppression_policies.py:    return interp1d(x=x, y=y, fill_value="extrapolate")
./libs/github_utils.py:    artifact_url, run_number = _get_artifact_zip_url(run_number=run_number, latest=latest)
./libs/datasets/data_version.py:            return func(version=version, **kwargs)
...
```
But I find the usage with `DataFrame.assign` a bit goofy because the parameter name is turned into a column label of the dataframe. I think it'd be better having a real dict, ie `df.assign({'most_recent_date': most_recent_date)...`

TomGoBravo:
[Make our vaccine backfill code smarter about how up-to-date the data is.](https://trello.com/c/tNlZtGp6/1279-make-our-vaccine-backfill-code-smarter-about-how-up-to-date-the-data-is)

libs/datasets/vaccine_backfills.py
mikelehen: Ohh....  I see.  Yeah, I knew Python had a version of named arguments but I didn't fully realize that it could essentially just be treated as an arbitrary dict under the covers...  and so with `.assign()`, the lefthand side can be an arbitrary column name as opposed to an actual function argument, so it didn't click that that's the mechanism being leveraged here... Makes sense now.

Yeah. FWIW, it seems like `.assign()` does accept an actual dict as well, so leaning towards the the `.assign({'most_recent_date': most_recent_date})` syntax as a convention in our code might be reasonable. 😁   Anyway, thanks for the pointer.

TomGoBravo: I don't think `.assign({'most_recent_date': most_recent_date})` works. It raises `TypeError: assign() takes 1 positional argument but 2 were given` (the first positional arg is the implicit `self` and the second is the dict). `.assign(**{'most_recent_date': most_recent_date})` works but depends on a different slightly esoteric python calling syntax.

mikelehen: Gah, sorry. You're of course right.  I just glanced at the [docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html) and the first example had `{'temp_c': [17.0, 25.0]}` in it, but looking closer, that was actually a call to `pd.DataFrame()` to set up an example dataframe. Sorry! 😬

PR #1096, opened 2021-04-29
backfill_vaccination_initiated incorrectly added empty time series

TomGoBravo:
This PR fixes a bug in `backfill_vaccination_initiated` that I noticed while working on https://github.com/covid-projections/covid-data-model/pull/1094

`computed_initiated` was being kept even when it was all NA for a location+bucket. This happens when one or both of administered and completed were all NA. The bug cause combined datasets to include a row with a derived tag but no real values. The fix is changing dropna axis param to `0, or ‘index’ : Drop rows which contain missing values.` (from DataFrame.dropna docs).

## Tested

```
./run.py data update
```
shows 2533 dropped empty `vaccinations_initiated` for various buckets that all have derived tag `{}`.

### DIFF
--- a/libs/datasets/vaccine_backfills.py
+++ b/libs/datasets/vaccine_backfills.py
@@ -92,7 +92,7 @@ def backfill_vaccination_initiated(dataset: MultiRegionDataset) -> MultiRegionDa
 
     # Compute and keep only time series with at least one real value
     computed_initiated = administered - completed
-    computed_initiated = computed_initiated.dropna(axis=1, how="all")
+    computed_initiated = computed_initiated.dropna(axis=0, how="all")
     # Keep the computed initiated only where there is not already an existing time series.
     computed_initiated = computed_initiated.loc[
         ~computed_initiated.index.isin(existing_initiated.index)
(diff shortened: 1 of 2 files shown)


libs/datasets/vaccine_backfills.py
```diff
@@ -92,7 +92,7 @@ def backfill_vaccination_initiated(dataset: MultiRegionDataset) -> MultiRegionDa
 
     # Compute and keep only time series with at least one real value
     computed_initiated = administered - completed
-    computed_initiated = computed_initiated.dropna(axis=1, how="all")
+    computed_initiated = computed_initiated.dropna(axis=0, how="all")
```
mikelehen, line 95: Out of curiosity, any reason to prefer the opaque `0` / `1` constants over the more descriptive `index` / `columns`?

mikelehen (approved):

libs/datasets/vaccine_backfills.py
TomGoBravo: The value for the axis parameter often doesn't make sense to me (does 'index', 'all' mean the values in every index offset must be na to drop a column OR it iterates over the index checking all the columns?) so I started using something more obscure in hopes that I'd be careful. Clearly failed here.

PR #1097, opened 2021-04-29
MultiRegionDataset.get_timeseries_bucketed_wide_dates and friends

TomGoBravo:
This is some more cleanup associated with https://github.com/covid-projections/covid-data-model/pull/1094

* Adds tag parameter `Derived.f` to save the function name, useful now that we're adding that tag to the same field from multiple places.
* Replaces some direct access to `MultiRegionDataset.timeseries_bucketed_wide_dates` with new function `get_timeseries_bucketed_wide_dates` that returns one field and similarly for `timeseries_not_bucketed_wide_dates`.
* Replaces some calls to Index.get_level_values with Index.unique where a copy of the value per index entry isn't needed. See https://stackoverflow.com/a/50364903

## Tested

```
./run.py data update
```
shows the same diff as https://github.com/covid-projections/covid-data-model/pull/1096 because main hasn't been update plus counties in IL have backfill_vaccination_initiated added to derived tag of vaccinations_initiated. As expected, don't see any other changes.

### DIFF
--- a/libs/datasets/taglib.py
+++ b/libs/datasets/taglib.py
@@ -315,16 +315,16 @@ class KnownIssueNoDate(TagInTimeseries):
 class Derived(TagInTimeseries):
     TAG_TYPE = TagType.DERIVED
     # Name of the function which added this derived tag.
-    f: str
+    function_name: str
 
     @classmethod
     def make_instance(cls, *, content: str) -> "TagInTimeseries":
         content_parsed = json.loads(content)
-        return cls(f=content_parsed.get("f", ""))
+        return cls(function_name=content_parsed.get("f", ""))
 
     @property
     def content(self) -> str:
-        d = {"f": self.f}
+        d = {"f": self.function_name}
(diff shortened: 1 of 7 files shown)


libs/datasets/taglib.py
```diff
@@ -314,15 +314,18 @@ def content(self) -> str:
 @dataclass(frozen=True)
 class Derived(TagInTimeseries):
     TAG_TYPE = TagType.DERIVED
+    # Name of the function which added this derived tag.
+    f: str
```
ghop02, line 318: It would be nice to make this name a bit more longer - I'm guessing its to save space in the csv. but since you're manually serializing/deserializing probably can make this a bit longer

ghop02 (approved):

libs/datasets/taglib.py
TomGoBravo: Done in 2b58e223

PR #1098, opened 2021-05-06
drop_regions_without_population fix for country data

TomGoBravo:
This PR is part of https://trello.com/c/sJDKzKIU/1302-make-usa-aggregation-use-incoming-usa-data-if-it-exists-rather-than-aggregate-the-states

* re-write drop_regions_without_population to not depend on MultiRegionDataset.static. Country data was silently dropped because the country did not appear in 'static'.
* Extend print_stats to breakdown by level so I can see where the country data is being dropped
* print_stats at a few more stages of data update

## Tested

`./run.py data update` reproduced identical output to main.

### DIFF
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -650,7 +650,6 @@ class MultiRegionDataset:
             self.timeseries_bucketed_long.index.get_level_values(CommonFields.LOCATION_ID)
             .map(dataset_utils.get_geo_data()[CommonFields.AGGREGATE_LEVEL])
             .value_counts()
-            .rename(None)
         )
         stats_in_text = "\n".join(
             [
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -1,3 +1,4 @@
+import textwrap
 from typing import (
     Any,
     Collection,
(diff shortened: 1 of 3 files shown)


libs/datasets/timeseries.py
```diff
@@ -645,7 +646,21 @@ def print_stats(self, name: str):
             .groupby(common_fields.COMMON_FIELD_TO_GROUP, sort=False)
             .sum()
         )
-        print(f"Dataset {name}:\n{count}")
+        by_level = (
+            self.timeseries_bucketed_long.index.get_level_values(CommonFields.LOCATION_ID)
+            .map(dataset_utils.get_geo_data()[CommonFields.AGGREGATE_LEVEL])
+            .value_counts()
+            .rename(None)
```
mikelehen, line 653: What does this `rename(None)` do?  Can't tell intuitively or from reading the `rename()` docs.  Probably deserves a comment if it's actually doing something.

mikelehen (approved):

libs/datasets/timeseries.py
TomGoBravo: It clears the series name, which was confusing when the series was output using repr. But I changed to `to_string()` which doesn't output the name so removed the rename.

TomGoBravo:
[Make USA aggregation use incoming USA data if it exists rather than aggregate the states.](https://trello.com/c/sJDKzKIU/1302-make-usa-aggregation-use-incoming-usa-data-if-it-exists-rather-than-aggregate-the-states)

PR #1100, opened 2021-05-10
Use unaggregated country data

TomGoBravo:
This PR addresses https://trello.com/c/sJDKzKIU/1302-make-usa-aggregation-use-incoming-usa-data-if-it-exists-rather-than-aggregate-the-states

* Factored out aggregation to the US country region so it is easy to add a unit test.
* Adds US population count in data source `fips_population.py` so that us country data isn't dropped by `drop_regions_without_population`.
* Adds property `MultiRegionDataset.variables` giving easy access to all variables in the dataset index
* Adds `MultiRegionDataset.drop_na_columns` which drops variables that don't have any real values

### Tested

`./run.py data update` and `git difftool --no-prompt -t vimdiff main -- data/multiregion-wide-dates.csv` shows only cases, deaths, new_cases and new_deaths change. They now include the NYT source tag and the values changed a tiny bit.


TomGoBravo:
[Make USA aggregation use incoming USA data if it exists rather than aggregate the states.](https://trello.com/c/sJDKzKIU/1302-make-usa-aggregation-use-incoming-usa-data-if-it-exists-rather-than-aggregate-the-states)

libs/datasets/sources/fips_population.py
```diff
@@ -76,60 +76,12 @@ def standardize_data(cls, data: pd.DataFrame) -> pd.DataFrame:
         states_aggregated[cls.Fields.FIPS] = states_aggregated[cls.Fields.STATE].map(ABBREV_US_FIPS)
         states_aggregated[cls.Fields.COUNTY] = None
 
-        common_fields_data = pd.concat([data, states_aggregated])
-        return common_fields_data
-
+        us_row = {
+            CommonFields.FIPS: "0",
+            CommonFields.COUNTRY: "USA",
+            CommonFields.POPULATION: states_aggregated[CommonFields.POPULATION].sum(),
```
ghop02, line 82: I think this makes sense - should we double check that our number matches other public US population numbers?

cli/data.py
```diff
@@ -179,18 +193,43 @@ def update(
...
+def _log_unexpected_aggregated_variables_to_drop(variables_to_drop: pd.Index):
+    unexpected_drops = variables_to_drop.difference(US_AGGREGATED_EXPECTED_VARIABLES_TO_DROP)
+    if not unexpected_drops.empty:
+        log = structlog.get_logger()
+        log.warn(
+            US_AGGREGATED_VARIABLE_DROP_MESSAGE, variable=unexpected_drops.to_list(),
+        )
+
+
+def run_aggregate_to_country(dataset_in: timeseries.MultiRegionDataset):
```
ghop02, line 212: Feels like the `run_aggregate_to_country` should be in `libs/datasets/usa_aggregation.py` or isn't there a `custom_aggregations` module? This file doesn't have too much of the guts in it and probably best to keep it that way

ghop02 (commented):
LGTM! A couple comments - mostly just small organizational ones around moving some of the more in the weeds code in `cli/data` to `libs`

ghop02 (approved):

libs/datasets/sources/fips_population.py
TomGoBravo: I took a look at https://www.census.gov/popclock/ which says it is 50 states and DC. Looking at 
```
geo_df = pd.read_csv('data/geo-data.csv')
static_df = pd.read_csv('data/multiregion-static.csv').merge(geo_df, on='location_id')
static_df.loc[(static_df['aggregate_level'] == 'state') & static_df['state'].isin(set(us_state_abbrev.STATES_AND_DC.values()))]['population'].sum()
```
we say 328.2M. Picking the date on https://www.census.gov/popclock/ I need to go back to about June 1 2019 to get that.
OTOH this PR doesn't modify the population. `multiregion-static` isn't changed.
@mikelehen Please make a new trello card if you think it is worth looking into this further.

PR #1104, opened 2021-05-12
calculate manual_filter diff using MultiRegionDatasetDelta

TomGoBravo:
This PR addresses https://trello.com/c/1p0uTkYH/1300-pipeline-generate-csv-of-blocked-metrics-and-make-it-easy-for-data-dashboard-to-read-it

* Adds `MultiRegionDataset.static_long`, the static DataFrame restructured into a Series
* Adds `MultiRegionDatasetDiff` to calculate the diff between two `MultiRegionDataset`
* Uses `MultiRegionDatasetDiff` to record timeseries removed by the manual filter.
* Adds some more asserts and comments related to `MultiRegionDataset.static` that I noticed while working on this change.

### DIFF
--- a/cli/data.py
+++ b/cli/data.py
@@ -120,7 +120,7 @@ def update(
     )
     before_manual_filter = multiregion_dataset
     multiregion_dataset = manual_filter.run(multiregion_dataset, region_overrides_config)
-    delta = timeseries.MultiRegionDatasetDelta.make(
+    delta = timeseries.MultiRegionDatasetDiff.make(
         old=before_manual_filter, new=multiregion_dataset
     )
     delta.timeseries_removed.write_to_wide_dates_csv(
--- a/cli/data.py
+++ b/cli/data.py
@@ -118,7 +118,15 @@ def update(
     region_overrides_config = manual_filter.transform_region_overrides(
         json.load(open(REGION_OVERRIDES_JSON)), aggregator.cbsa_to_counties_region_map
     )
+    before_manual_filter = multiregion_dataset
     multiregion_dataset = manual_filter.run(multiregion_dataset, region_overrides_config)
+    delta = timeseries.MultiRegionDatasetDelta.make(
+        old=before_manual_filter, new=multiregion_dataset
+    )
+    delta.timeseries_removed.write_to_wide_dates_csv(
+        pathlib.Path("data/manual_filter_removed-wide-dates.csv"),
+        pathlib.Path("data/manual_filter_removed-static.csv"),
+    )
 
     multiregion_dataset.print_stats("combined")
     multiregion_dataset = outlier_detection.drop_tail_positivity_outliers(multiregion_dataset)
--- a/libs/datasets/timeseries.py
+++ b/libs/datasets/timeseries.py
@@ -1607,21 +1607,26 @@ def make_source_url_tags(ds_in: MultiRegionDataset) -> MultiRegionDataset:
 # eq=False because instances are large and we want to compare by id instead of value
 @final
 @dataclasses.dataclass(frozen=True, eq=False)
-class MultiRegionDatasetDelta:
+class MultiRegionDatasetDiff:
     """Represents a delta/diff between two MultiRegionDataset objects."""
 
     old: MultiRegionDataset
     new: MultiRegionDataset
 
     @staticmethod
-    def make(*, old, new) -> "MultiRegionDatasetDelta":
-        return MultiRegionDatasetDelta(old=old, new=new)
+    def make(*, old, new) -> "MultiRegionDatasetDiff":
+        return MultiRegionDatasetDiff(old=old, new=new)
 
     @property
     def timeseries_removed(self) -> MultiRegionDataset:
-        """A dataset containing time series, tags and static values in old but not new."""
+        """A dataset containing time series, tags and static values in old but not new.
+
+        A time series is considered removed if it has at least one real (not-NA) value in
+        `old` and no real values (all NA) in `new`. Changes in the set of dates with a real value
+        and changes in the values themselves are ignored.
+        """
         # removed is currently calculated when accessed but it may make sense to move this to
-        # `make` depending on future uses of MultiRegionDatasetDelta.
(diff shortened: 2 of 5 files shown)


cli/data.py
```diff
@@ -118,7 +118,15 @@ def update(
...
     )
+    before_manual_filter = multiregion_dataset
     multiregion_dataset = manual_filter.run(multiregion_dataset, region_overrides_config)
+    delta = timeseries.MultiRegionDatasetDelta.make(
+        old=before_manual_filter, new=multiregion_dataset
+    )
+    delta.timeseries_removed.write_to_wide_dates_csv(
+        pathlib.Path("data/manual_filter_removed-wide-dates.csv"),
+        pathlib.Path("data/manual_filter_removed-static.csv"),
+    )
```
mikelehen, line 129: Wow, this is elegant!

libs/datasets/timeseries.py
```diff
@@ -770,6 +786,11 @@ def from_timeseries_df(timeseries_df: pd.DataFrame) -> "MultiRegionDataset":
         return MultiRegionDataset(timeseries=timeseries_df)
 
     def add_fips_static_df(self, latest_df: pd.DataFrame) -> "MultiRegionDataset":
+        # This function is only called on empty datasets and is very unlikely to get any new
+        # callers. This is a useful constraint when simplifying add_static_values.
+        # TODO(tom): Uncomment after fixing test_multi_region_to_from_timeseries_and_latest_values
+        # assert self.timeseries_bucketed.empty
```
mikelehen, line 792: What's the plan for this TODO?  Upcoming PR?  Maybe sometime?  Other?

libs/datasets/timeseries.py
```diff
@@ -778,6 +799,8 @@ def add_static_values(self, attributes_df: pd.DataFrame) -> "MultiRegionDataset"
         scalars_without_geodata = attributes_df.loc[
             :, attributes_df.columns.difference(GEO_DATA_COLUMNS)
         ]
+        # TODO(tom): When this assert always passes remove all the junk in _merge_attributes.
```
mikelehen, line 802: This TODO is a bit confusing.  Should it live in `_merge_attributes` next to the code we want to remove?

libs/datasets/timeseries.py
```diff
@@ -1433,7 +1464,7 @@ def combined_datasets(
     if static_series:
         output_static_df = pd.concat(
             static_series, axis=1, sort=True, verify_integrity=True
-        ).rename_axis(index=CommonFields.LOCATION_ID)
+        ).rename_axis(index=CommonFields.LOCATION_ID, columns=PdFields.VARIABLE)
```
mikelehen, line 1467: nit: Should we have the same Pandas 1.2 TODO here?

libs/datasets/timeseries.py
```diff
@@ -1571,3 +1602,42 @@ def make_source_url_tags(ds_in: MultiRegionDataset) -> MultiRegionDataset:
     )
     source_url[TagField.TYPE] = TagType.SOURCE_URL
     return ds_in.append_tag_df(source_url)
+
+
+T_SERIES_OR_DATAFRAME = TypeVar("T_SERIES_OR_DATAFRAME", pd.Series, pd.DataFrame)
```
mikelehen, line 1607: nit: I thought `T_SERIES` was a weird abbreviation of TIMESERIES at first glance.  I don't think we use `T_` as a type prefix anywhere else in the codebase.  And in general I think the convention for type variables is to use CamelCase (e.g. `AnyStr`, `Dict`, `Union`, etc.).  And we actually have an identical `SeriesOrDataframe` type variable already over here: https://github.com/covid-projections/covid-data-model/blob/769738a9da44ad9fd98518bf869042bd222cad73/libs/parallel_utils.py#L28 (though I'm the one that added it so who knows if I did it optimally 😁 ).

libs/datasets/timeseries.py
```diff
@@ -1571,3 +1602,42 @@ def make_source_url_tags(ds_in: MultiRegionDataset) -> MultiRegionDataset:
...
+    old: MultiRegionDataset
+    new: MultiRegionDataset
+
+    @staticmethod
+    def make(*, old, new) -> "MultiRegionDatasetDelta":
+        return MultiRegionDatasetDelta(old=old, new=new)
+
+    @property
+    def timeseries_removed(self) -> MultiRegionDataset:
+        """A dataset containing time series, tags and static values in old but not new."""
```
mikelehen, line 1624: What exactly happens with timeseries that are partially filtered via a `start_date`.  At first glance I thought this wouldn't include them, but I guess maybe we add a manual filtered tag/annotation to them and so that gets them included in this result? Is that right?  Might be worth a mention in the comment since it's a bit non-intuitive.

mikelehen (approved):
LGTM with nits.

TomGoBravo:

libs/datasets/timeseries.py
TomGoBravo: I suspect it'll still be needed here because this DF is built from single columns. I changed the TODO to remind myself to search for all calls to rename_axis.

libs/datasets/timeseries.py
TomGoBravo: I'm 80% sure this is not an intended use of pandas but `from pandas.core.generic import FrameOrSeries` does what we want and is heavily used within pandas. _typing.py has a nice comment explaining TypeVar vs Union. See also https://stackoverflow.com/a/58775376

libs/datasets/timeseries.py
TomGoBravo: You are on the mark. Changes in the set of dates with a real value and changes in the values themselves are ignored but the extra tag will be added in the timeseries_removed dataset. I'd like to incorporate changes within a timeseries more cleanly, not sure how yet.

libs/datasets/timeseries.py
TomGoBravo: I don't really have a plan. Upcoming maybe sometime when I don't feel like I'm scrambling to do something more urgent.

PR #1107, opened 2021-05-13
Browse dashboard files

TomGoBravo:
This PR is for https://trello.com/c/1p0uTkYH/1300-pipeline-generate-csv-of-blocked-metrics-and-make-it-easy-for-data-dashboard-to-read-it

* Adds dashboard feature to pick one of two different datasets. New classes `RepoWrapper` and `DatasetFile` help.
* Replaces buttons to access pivot table presets with a dropdown. This makes it easier to handle a dataset change because the state of the dropdown is preserved, unlike the last button pressed.
* Factors out manual filter removed csv file paths. Having multiple constants isn't pretty but better than duplicating string literals.
* In timeseries_stats, add OBSERVATION_COUNT and remove HAS_URL because I think it is more interesting. Add PerTimeseries.dataset because the MultiRegionDataset is often needed too.


dash_app/dashboard.py
```diff
@@ -55,7 +62,62 @@ def _generate_next_value_(name, start, count, last_values):  # pylint: disable=n
...
+@dataclasses.dataclass(frozen=True)
+class RepoWrapper:
+    _repo: git.Repo
+    head_commit_str: str
+    _working_copy_dataset: timeseries.MultiRegionDataset
+
+    @staticmethod
+    def make() -> "RepoWrapper":
+        # Add odbt=git.db.GitDB if there are problems reading files from git. See
+        # https://github.com/gitpython-developers/GitPython/issues/642#issuecomment-618117030
```
mikelehen, line 99: Not sure how likely it is that people will hit this, but if it's at all likely, then it might be nice to catch the problem if at all possible and either automatically apply the work around or log an explicit message about the workaround, so somebody doesn't have to go spelunking to find this comment.   But if it's not likely, then meh. 🤷

dash_app/dashboard.py
```diff
@@ -55,7 +62,62 @@ def _generate_next_value_(name, start, count, last_values):  # pylint: disable=n
...
+    # Define a custom __new__ so DashboardFile instances use 'filename' as their value and have a
+    # description.
+    def __new__(cls, filename, description):
+        # Initialize super class (str) with filename.
+        obj = super().__new__(cls, filename)
+        # _value_ is a special name of enum. Set it here so enum code doesn't attempt to call
+        # str(filename, description).
+        obj._value_ = filename
+        obj.description = description
+        obj.filename = filename
```
mikelehen, line 83: nit: FWIW- It was confusing to me why we have a `filename` here but then instead of using it, code goes and looks up the corresponding `dataset_utils.MANUAL_FILTER_REMOVED_WIDE_DATES_CSV_PATH` constants...  I guess this `filename` is really just a unique ID used in the html.  It might make sense to rename it if we aren't planning to actually use it as a filename.

dash_app/dashboard.py
```diff
@@ -276,42 +351,32 @@ def update_dataset_page_content(pathname):
...
-            preset = more_itertools.first(TimeSeriesPivotTablePreset)
-        else:
-            preset = TimeSeriesPivotTablePreset.get_by_btn_id(
-                ctx.triggered[0]["prop_id"].split(".")[0]
-            )
-
+    def time_series_pivot_table_preset_btn_clicked(
+        pivot_table_presets_dropdown_value, datasets_dropdown_value
+    ):
+        """Make a new pivot table according to ..."""
```
mikelehen, line 377: `...`?

libs/datasets/dataset_utils.py
```diff
@@ -55,14 +55,16 @@ def _get_public_data_path():
 
 DATA_DIRECTORY = REPO_ROOT / "data"
 
-# TODO(tom): Clean up how the test-combined csv paths are passed around when removing
-#  DatasetPointer.
+# TODO(tom): Clean up how the wide-dates and static csv paths are passed around when removing
+#  DatasetPointer. See also enum `DashboardFile` in dashboard.py.
```
mikelehen, line 59: Oh.  I guess maybe this TODO covers my comment above.

mikelehen (approved):
LGTM!

TomGoBravo:

PR #1108, opened 2021-05-14
add all timeseries touched by manual filter

TomGoBravo:
This PR is part of https://trello.com/c/1p0uTkYH/1300-pipeline-generate-csv-of-blocked-metrics-and-make-it-easy-for-data-dashboard-to-read-it

* Replaces use of `MultiRegionDatasetDiff` to create the view of manual filter / blocked data with custom function `manual_filter.touched_subset`. This new output includes, for time series blocked or annotated by the manual filter, the unmodified input time series and tags added by the manual filter

### Tested

Ran `./run.py data update` and `git difftool --no-prompt -t vimdiff main -- data/manual_filter_removed-wide-dates.csv`. There I saw a bunch of counties in CA, CT state test data, counties in IL vaccine data, NY state and county tests. Maui hospital numbers, Mobile AL, Maricopa AZ were already in the file and unchanged.


TomGoBravo:

mikelehen (approved):
Nice, this looks good.

I think it might be better if this only picked up `blocked=True` timeseries.  As-is it picks up all of our disclaimers as well, which is a bit noisy given that we have disclaimers for every county in CA, CT, IL, etc...  But since we're not actually blocking the data, we don't really need them included here.  We can always look at the data on the website or in combined datasets, etc.  But this isn't a blocker...  and maybe can be punted indefinitely.

TomGoBravo:
> Nice, this looks good.
> 
> I think it might be better if this only picked up `blocked=True` timeseries. As-is it picks up all of our disclaimers as well, which is a bit noisy given that we have disclaimers for every county in CA, CT, IL, etc... But since we're not actually blocking the data, we don't really need them included here. We can always look at the data on the website or in combined datasets, etc. But this isn't a blocker... and maybe can be punted indefinitely.

`touched_subset` doesn't have easy access to the blocked boolean right now. I could add the bool to the tag but think it might be more useful to only add time series where the number of observations decreased. I'll do this in a follow up PR.

mikelehen:
Cool.  Another option would be to just add a different tag for when we block data vs just add a disclaimer.

PR #1111, opened 2021-05-17
Pickle MultiRegionDataset

TomGoBravo:
Add simple support for serialization and deserialization of `MultiRegionDataset` using python pickle. Files created like this are much smaller than any CSV (less than 1/2 a wide-dates file). Compared to parquet and hd5 this makes restoring the entire object trivial but the files are not easy to read outside of Python.

Adds the raw combined data to the data dashboard.

Adds a command to copy a pickle file to CSV. For example: `./run.py data pickle-to-csv data/multiregion.pkl.gz data/multiregion.pkl-wide-dates.csv`

To dump a time series from the new combined raw file try:

```
from libs.datasets.timeseries import MultiRegionDataset
import pathlib

ds = MultiRegionDataset.from_compressed_pickle(pathlib.Path('data/combined-raw.pkl.gz'))
ds.timeseries_bucketed_wide_dates.loc[('iso1:us', 'cases', 'all'), :]
```


mikelehen (commented):
This might be a good one for @ghop02 to weigh in on.  I'm a bit loathe to use a non-standard format that requires our own internal tools to convert to/from standard formats.  But maybe it's fine in practice.

I do think this will probably break @BrettBoval 's workflow for Surgo.  They would need to enlist in our repo, set up their dev environment, and run the command-line tool to generate a csv.  Parquet would be much friendlier in this regard.  But then again, we don't generally expect people to be reading files directly from our repos, so again, maybe not a major issue in practice.

TomGoBravo:
> This might be a good one for @ghop02 to weigh in on. I'm a bit loathe to use a non-standard format that requires our own internal tools to convert to/from standard formats. But maybe it's fine in practice.
> 
> I do think this will probably break @BrettBoval 's workflow for Surgo. They would need to enlist in our repo, set up their dev environment, and run the command-line tool to generate a csv. Parquet would be much friendlier in this regard. But then again, we don't generally expect people to be reading files directly from our repos, so again, maybe not a major issue in practice.

This PR does add use of non-standard formats but only for new files, existing files remain in their current format. I want this change so I can more easily and quickly run subsets of the pipeline locally.

I spent a while trying to serialize a MultiRegionDataset in one parquet or hd5 file but haven't had good luck yet. It looks like a parquet file has one table and columns must be all one type so getting all tags, static and time series data in one table is annoying. hd5 files can contain multiple tables but are huge.

ghop02:
> > This might be a good one for @ghop02 to weigh in on. I'm a bit loathe to use a non-standard format that requires our own internal tools to convert to/from standard formats. But maybe it's fine in practice.
> > I do think this will probably break @BrettBoval 's workflow for Surgo. They would need to enlist in our repo, set up their dev environment, and run the command-line tool to generate a csv. Parquet would be much friendlier in this regard. But then again, we don't generally expect people to be reading files directly from our repos, so again, maybe not a major issue in practice.
> 
> This PR does add use of non-standard formats but only for new files, existing files remain in their current format. I want this change so I can more easily and quickly run subsets of the pipeline locally.

Can you elaborate on what you mean by easily and quickly running subsets of the pipeline locally?  Curious what you have in mind!

> 
> I spent a while trying to serialize a MultiRegionDataset in one parquet or hd5 file but haven't had good luck yet. It looks like a parquet file has one table and columns must be all one type so getting all tags, static and time series data in one table is annoying. hd5 files can contain multiple tables but are huge.

There is a lot of convenience to the pickle.  If we are adverse to pickling objects (i'm generally pretty opposed to pickling, but also in this case could see it working out fine practically) we could save the csv into a single folder and compress that folder (then that compressed folder would essentially be a pickled representation).  

I feel like pickling objects is mostly problematic when actively developing on the pickled objects. If i make a change to the structure of multiregiondataset, loading/saving could be problematic.  Which might not be an issue anymore.  I think ned summarizes the [dangers of pickling here](https://nedbatchelder.com/blog/202006/pickles_nine_flaws.html).  

Anyways, I would strongly prefer not to pickle the main multi region dataset file, but if there's some other use case in mind I think i could be swayed (though still think zipping the output of 3 csvs feels a bit better than pickling)

TomGoBravo:
> > > This might be a good one for @ghop02 to weigh in on. I'm a bit loathe to use a non-standard format that requires our own internal tools to convert to/from standard formats. But maybe it's fine in practice.
> > > I do think this will probably break @BrettBoval 's workflow for Surgo. They would need to enlist in our repo, set up their dev environment, and run the command-line tool to generate a csv. Parquet would be much friendlier in this regard. But then again, we don't generally expect people to be reading files directly from our repos, so again, maybe not a major issue in practice.
> > 
> > 
> > This PR does add use of non-standard formats but only for new files, existing files remain in their current format. I want this change so I can more easily and quickly run subsets of the pipeline locally.
> 
> Can you elaborate on what you mean by easily and quickly running subsets of the pipeline locally? Curious what you have in mind!

What I've already been doing is saving the dataset just before the thing I want to run again, then manually deleting all the code before it and changing the save to a load. Easy and fast.

> > I spent a while trying to serialize a MultiRegionDataset in one parquet or hd5 file but haven't had good luck yet. It looks like a parquet file has one table and columns must be all one type so getting all tags, static and time series data in one table is annoying. hd5 files can contain multiple tables but are huge.
> 
> There is a lot of convenience to the pickle. If we are adverse to pickling objects (i'm generally pretty opposed to pickling, but also in this case could see it working out fine practically) we could save the csv into a single folder and compress that folder (then that compressed folder would essentially be a pickled representation).
> 
> I feel like pickling objects is mostly problematic when actively developing on the pickled objects. If i make a change to the structure of multiregiondataset, loading/saving could be problematic. Which might not be an issue anymore. I think ned summarizes the [dangers of pickling here](https://nedbatchelder.com/blog/202006/pickles_nine_flaws.html).
> 
> Anyways, I would strongly prefer not to pickle the main multi region dataset file, but if there's some other use case in mind I think i could be swayed (though still think zipping the output of 3 csvs feels a bit better than pickling)

That makes sense for data that are other systems may want to read. We haven't prioritized writing that code and I want something that works for me that doesn't take long to write.

mikelehen (approved):
Yeah, I guess I'm not opposed to this if it's just for adding some internal snapshots to the pipeline so we can use our internal tools to inspect the data, but I do like the ability to just open files in a jupyter notebook / whatever (like Brett was doing), which we lose if we're relying on pickling / unpickling internal types.

So I'd prefer we find another solution (such as zipping a folder of csv/parquet/whatever) as Chris suggested for our long-term needs...  at which point we might as well use that for our internal needs too?

PR #1113, opened 2021-05-18
add drop_observations after today UTC

TomGoBravo:
This PR addresses https://trello.com/c/UKqdcj0z/1345-pipeline-block-data-for-future-dates

* Adds function `timeseries.drop_observations(..., after=)` and tag `drop_future_observation` to record where it did something
* Use this function to drop data that is more recent than today, UTC
* Clean up print_stats calls so "combined" is right after the combined data is produced and adding "manual filter" stats

### Tested

Running `./run.py data update` with covid-data-public manually tweaked to have a date in the future.

### DIFF
--- a/libs/datasets/taglib.py
+++ b/libs/datasets/taglib.py
@@ -59,6 +59,7 @@ class TagType(GetByValueMixin, ValueAsStrMixin, str, enum.Enum):
     KNOWN_ISSUE = "known_issue"
     KNOWN_ISSUE_NO_DATE = "known_issue_no_date"
     DERIVED = "derived"
+    DROP_FUTURE_OBSERVATION = "drop_future_observation"
 
     PROVENANCE = PdFields.PROVENANCE
     SOURCE_URL = "source_url"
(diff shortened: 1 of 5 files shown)


libs/datasets/taglib.py
```diff
@@ -328,6 +329,22 @@ def content(self) -> str:
...
 
+@dataclass(frozen=True)
+class DropFutureObservation(TagInTimeseries):
+    TAG_TYPE = TagType.DROP_FUTURE_OBSERVATION
+    after: datetime.date
+
+    @classmethod
+    def make_instance(cls, *, content: str) -> "TagInTimeseries":
+        content_parsed = json.loads(content)
+        return cls(after=datetime.date.fromisoformat(content_parsed["after"]),)
```
mikelehen, line 340: Is that trailing `,` intentional / necessary?

mikelehen (approved):
LGTM, thanks!

libs/datasets/taglib.py
TomGoBravo: Nope. Upgrading black fixes this but also makes lots of other changes so won't do it in this PR.

PR #1115, opened 2021-05-18
touched_subset only returns time series with dropped observations

TomGoBravo:
This PR is last remaining part of https://trello.com/c/1p0uTkYH/1300-pipeline-generate-csv-of-blocked-metrics-and-make-it-easy-for-data-dashboard-to-read-it

Follow up on https://github.com/covid-projections/covid-data-model/pull/1108#issuecomment-842694398 to only put timeseries that actually had observations dropped in data/manual_filter_removed-wide-dates.csv

This also fixes a test breakage introduced by 08fe29e1834.

### Tested

Ran `./run.py data update` with main c76029ee merged in and covid-data-public in sync. Changes only seen in data/manual_filter_removed-wide-dates.csv. They are removal of CA county cases, CT state tests, IL county vaccine, NY state and county tests. This matches what is `blocked: false` in data/region-overrides.json


TomGoBravo:

libs/datasets/manual_filter.py
```diff
@@ -153,23 +153,42 @@ def touched_subset(
...
-    # Get all ds_in.tag associated with any time series in touch_index
-    tag_in = ds_in.tag.loc[ds_in.tag.index.droplevel(taglib.TagField.TYPE).isin(touch_index)]
-    # Get only the ds_out.tag where the type is _EXPECTED_TYPES. These are the tags added by
-    # `run` and not in ds_in.tag.
-    tag_out = ds_out.tag.loc[tag_mask]
+    assert wide_dates_in_df.index.names == ts_has_tag_index.names
+    assert ts_has_tag_index.isin(wide_dates_in_df.index).all()
+    # For time series identified by ts_has_tag_index find the number of observations in ds_in and
+    # ds_out.
+    ts_in_observations = wide_dates_in_df.reindex(ts_has_tag_index).notna().sum(axis="columns")
```
mikelehen, line 168: nit: I haven't fully internalized these type prefixes, but is this actually a timeseries?  Aren't we summing across the dates and this is not time-related anymore?  Perhaps just `in_observations` and `out_observations` or something?

libs/datasets/manual_filter.py
```diff
@@ -153,23 +153,42 @@ def touched_subset(
...
+    ts_out_observations = (
+        ds_out.timeseries_bucketed_wide_dates.reindex(ts_has_tag_index).notna().sum(axis="columns")
+    )
+    # Some time series may have been dropped from ds_in to ds_out (if all observations were
+    # dropped) but no time series were added.
+    assert ts_out_observations.index.isin(ts_in_observations.index).all()
```
mikelehen, line 174: Since we explicitly `reindex()`ed both of these based on `ts_has_tag_index` it seems like these indexes should always be exactly the same (and therefore the assert probably is unnecessary), no?

mikelehen (approved):
LGTM with nits.

libs/datasets/manual_filter.py
TomGoBravo: I'm not very consistent with the prefixes and suffixes :-/
Renamed to `out_observations_by_ts` and `in_observations_by_ts`.

libs/datasets/manual_filter.py
TomGoBravo: Good point. Removed and updated comment.

PR #1117, opened 2021-05-19
combine by distribution

TomGoBravo:
This PR addresses https://trello.com/c/F76nvt6x/1375-combine-data-sources-per-distribution-instead-of-per-field

This depends on https://github.com/covid-projections/covid-data-public/pull/222

### Tested

`./run.py data update` then `git difftool --no-prompt -t vimdiff main -- data/multiregion-wide-dates.csv`
shows many new timeseries.

Counties in many states (AL, CT, FL, GA, IA, ID, IL, IN, MD, MI, MN, MO, MS, MT, NC, ND, NJ, NM, NY, OR, VT, WA, WV, WY) add vaccinations_completed with buckets age:18_plus,age:65_plus.

In MA some counties add vaccinations_completed all from CDC and vaccinations_initiated all estimate_initiated_from_state_ratio. 3 counties add vaccinations_completed all blocked without a public note.

DC and WI state level add lots of new non-all distributions for cases, deaths, vaccinations.

NE state, TN state, PA state and one county (iso1:us#iso2:us-pa#fips:42101) add a bunch of non-all vaccine distributions.

TN counties add cases age distributions.


TomGoBravo:
[Combine data sources per distribution instead of per field](https://trello.com/c/F76nvt6x/1375-combine-data-sources-per-distribution-instead-of-per-field)

mikelehen:
> though they seem to be blocked further down the pipeline

Hrm... we should be blocking 3 MA counties, but not sure why others would be getting blocked!

Also, looks like there are some unit test failures FYI.

TomGoBravo:
> > though they seem to be blocked further down the pipeline
> 
> Hrm... we should be blocking 3 MA counties, but not sure why others would be getting blocked!

Data was correct, description wrong. Updated it.

> Also, looks like there are some unit test failures FYI.

Fixed now that PdFields is update.

libs/datasets/timeseries.py
```diff
@@ -634,21 +649,33 @@ def timeseries_bucketed_long(self) -> pd.Series:
...
+        )
+
     @cached_property
     def wide_var_not_null(self) -> pd.DataFrame:
         """True iff there is at least one real value in any bucket with given location and
         variable"""
         wide_var_not_null = (
-            self.timeseries_bucketed_long.notnull()
-            .groupby([CommonFields.LOCATION_ID, PdFields.VARIABLE], sort=False)
+            self.timeseries_distribution_long.notna()
```
mikelehen, line 667: Are we intentionally switching from `notnull()` to `notna()`. Is it safe to assume that we never have `None` or `NaT` and so this is equivalent?

libs/datasets/timeseries.py
```diff
@@ -1559,12 +1587,14 @@ def _with_common_index(
...
 ) -> Mapping[MultiRegionDataset, pd.DataFrame]:
     dataset_wide_first = more_itertools.first(datasets_wide.values())
-    assert dataset_wide_first.index.names == [CommonFields.LOCATION_ID]
+    assert dataset_wide_first.index.names == [CommonFields.LOCATION_ID, PdFields.DISTRIBUTION]
     assert dataset_wide_first.columns.names == [PdFields.VARIABLE]
-    common_index = pd.Index(
-        np.unique(np.concatenate([ds_w.index for ds_w in datasets_wide.values()])),
-        name=dataset_wide_first.index.name,
-    )
+    common_index = None
```
mikelehen, line 1592: nit: Colud set this to `dateset_wide_first.index` and then remove the `is None` case below?

mikelehen (approved):
Nice! Thanks for taking care of this.

libs/datasets/timeseries.py
TomGoBravo: `notnull` and `notna` are aliases. https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.notna.html and thanks to https://github.com/covid-projections/covid-data-model/blob/main/libs/datasets/timeseries.py#L993 we can be sure all values are floats (real or nan).

libs/datasets/timeseries.py
TomGoBravo: If I changed it I'd also need to skip the first when iterating through `values()` so decided to stick with the branch instead of adding more abstractions.

mikelehen: Out of curiosity, wouldn't `union()` gracefully ignore duplicates?  Definitely no problem with the existing code though.

TomGoBravo: I'd expect so but it seems clearer to write what I want to happen.

PR #1131, opened 2021-05-26
crudely replace icu metrics with None and np.nan

TomGoBravo:
This PR addresses https://trello.com/c/r7LsqyyQ/1328-remove-icu-headroom-and-typical-usage-rate-from-api-and-purge-covid-care-map

PR #1132, opened 2021-05-27
Revert "crudely replace icu metrics with None and np.nan"

TomGoBravo:
Reverts covid-projections/covid-data-model#1131


TomGoBravo:
[Remove icu headroom and typical usage rate from API and purge covid care map.](https://trello.com/c/r7LsqyyQ/1328-remove-icu-headroom-and-typical-usage-rate-from-api-and-purge-covid-care-map)

PR #1136, opened 2021-06-08
Allow region to be a comma-separated list in region-overrides.json.

mikelehen:
Tested via `./run.py data update` and verifying `multiregion-wide-dates.csv` and `manual_filter_removed-wide-dates.csv` did not change.


libs/datasets/manual_filter.py
```diff
@@ -240,14 +240,46 @@ def _transform_one_override(
...
+    if not isinstance(include_str, str):
+        raise ValueError(f"Invalid 'include': {include_str}")
+
+    # Split on commas.
+    region_strs = [single_region_str.strip() for single_region_str in region_str.split(",")]
+    # Convert each region string to 1+ regions and aggregate.
+    regions_included = []
+    for region_str in region_strs:
+        for region in _transform_region_str(region_str, include_str, cbsa_to_counties_map):
+            regions_included.append(region)
```
mikelehen, line 256: FWIW- I tried to do this all with a list comprehension but to get the regions list properly flattened it seemed easier to just do it procedurally like this... 🤷

TomGoBravo: I think `regions_included.extend(transform_region_str(region_str, include_str, cbsa_to_counties_map))` eliminates one loop without much complication.

libs/datasets/manual_filter.py
```diff
@@ -240,14 +240,46 @@ def _transform_one_override(
...
     region_str = override["region"]
+    include_str = override["include"]
+
+    if not isinstance(region_str, str):
+        raise ValueError(f"Invalid 'region': {region_str}")
+    if not isinstance(include_str, str):
+        raise ValueError(f"Invalid 'include': {include_str}")
+
+    # Split on commas.
+    region_strs = [single_region_str.strip() for single_region_str in region_str.split(",")]
```
TomGoBravo, line 251: How about doing this parsing of the region_str in `_transform_region_str`?

TomGoBravo:
https://github.com/covid-projections/covid-data-model/blob/main/tests/libs/datasets/manual_filter_test.py#L235 is an example of making the KNOWN_ISSUE_NO_DATE tag in expected data when blocked is true. You could alternatively change blocked to false and add the tag like `TimeseriesLiteral([6, 8], annotation=[tag])`, which avoids the complication of adding a tag to a timeseries with no real values.

mikelehen (commented):
Thanks! Made suggested changes.  PTAL @TomGoBravo

mikelehen:
[Metric blocking: Support single override targeting multiple FIPS.](https://trello.com/c/QZYvrX0d/1401-metric-blocking-support-single-override-targeting-multiple-fips)

tests/libs/datasets/manual_filter_test.py
```diff
@@ -369,6 +369,58 @@ def test_region_overrides_transform_and_filter_start_end_dates(
...
+    )
+
+    tag_expected = test_helpers.make_tag(
+        taglib.TagType.KNOWN_ISSUE_NO_DATE,
+        public_note=region_overrides["overrides"][0]["disclaimer"],
+    )
+    ds_expected = test_helpers.build_dataset(
+        {
+            region_wy: {
+                CommonFields.VACCINATIONS_INITIATED: TimeseriesLiteral(
```
TomGoBravo, line 408: `CommonFields.CASES` may be short enough to avoid the line wrapping.

libs/datasets/manual_filter.py
```diff
@@ -279,6 +264,41 @@ def _transform_one_override(
     )
 
 
+def _transform_region_str(
+    raw_region_str: str, include_str: str, cbsa_to_counties_map: Mapping[Region, List[Region]]
+) -> List[Region]:
```
TomGoBravo, line 269: `-> List[RegionMaskOrRegion]`

TomGoBravo (approved):

libs/datasets/manual_filter.py
TomGoBravo: Hmm... `list(chain.from_iterable(map(_transform_region_str, region_str.split(','))))` is pretty deeply nested but not horrible.

mikelehen: Ah, nice.  I opted to leave it as-is, but thanks for the pointer.

PR #1137, opened 2021-06-09
Convert label_api_snapshot.yml to workflow_dispatch.

mikelehen:
This way if we ever need to revert an API snapshot, we can do it via the github UI...

Also hardcoded the label to "latest" since we don't need that flexibility.

### DIFF
--- a/tools/label-api.sh
+++ b/tools/label-api.sh
@@ -13,7 +13,6 @@ prepare () {
     exit_with_usage
   else
     SNAPSHOT_ID=$1
-    LABEL="latest"
   fi
 
   if ! [[ $SNAPSHOT_ID =~ ^[0-9]+$ ]] ; then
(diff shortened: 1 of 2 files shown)


mikelehen:
[Add workflow dispatch for labeling API snapshots (in case of need to revert to a prior snapshot).](https://trello.com/c/txEhKDhk/1398-add-workflow-dispatch-for-labeling-api-snapshots-in-case-of-need-to-revert-to-a-prior-snapshot)

tools/label-api.sh
```diff
@@ -40,8 +39,8 @@ exit_with_usage () {
 execute () {
   curl -H "Authorization: token $GITHUB_TOKEN" \
       --request POST \
-      --data "{\"event_type\": \"label-api-snapshot\", \"client_payload\": { \"label\": \"${LABEL}\", \"snapshot_id\": \"${SNAPSHOT_ID}\" } }" \
-      https://api.github.com/repos/covid-projections/covid-data-model/dispatches
+      --data "{\"inputs\": { \"snapshot_id\": \"${SNAPSHOT_ID}\" } }" \
```
ghop02, line 42: I think you have to specify the ref as well? Maybe that's changed, but i think i remember that was the case (and i think is the pattern elsewhere)

ghop02 (approved):
One question about payload syntax, but other than that looks good!

tools/label-api.sh
mikelehen: I'm not sure!  I was hoping it was optional and removed it.  I think I'll try it and find out. 😁

ghop02: such is life with github actions!

mikelehen: You were right! 😆 

```
{
  "message": "Invalid request.\n\n\"ref\" wasn't supplied.",
  "documentation_url": "https://docs.github.com/rest/reference/actions#create-a-workflow-dispatch-event"
}
```

Fixed here: https://github.com/covid-projections/covid-data-model/commit/345aa948a67bee005b4e805e0fd3f14e22f43299

PR #1138, opened 2021-06-15
Round metric values.

mikelehen:
This fixes the Arkansas issue we looked at this morning (as well as a number of other cases where states/counties were on a metric threshold).

I verified that the generated summary file on the frontend (where we already were doing this kind of rounding) was unaffected except:
1. As desired, risk levels were fixed when a rounded metric value ended up on a metric threshold.
2. JavaScript and Python round slightly differently when the value ends in a `5` (e.g. `0.115` rounds to `0.12` in JavaScript but `0.11` in Python).

Also worth noting that the resulting .timeseries.json API endpoints are ~35% smaller.

### DIFF
--- a/libs/metrics/top_level_metrics.py
+++ b/libs/metrics/top_level_metrics.py
@@ -48,6 +48,25 @@ class MetricsFields(common_fields.ValueAsStrMixin, str, enum.Enum):
     VACCINATIONS_COMPLETED_RATIO = "vaccinationsCompletedRatio"
 
 
+# These precisions should be inline with
+# https://github.com/covid-projections/covid-projections/blob/c076f39f54dcf6ca3f20fbb67839c37bb8b0f5bf/src/common/metric.tsx#L90
+# but note that percentages (ICU_*, TEST_POSITIVITY, VACCINATIONS_*) need an
+# additional 2 digits of precisions since they will be converted from ratio
+# (0.xyz) to percentage (XY.Z%).
+
+METRIC_ROUNDING_PRECISION = {
+    MetricsFields.CASE_DENSITY_RATIO: 1,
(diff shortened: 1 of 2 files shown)


mikelehen:
Note the failing tests are unrelated to my change.  One of the tests has been crashing (only in CI) for the last few days.  Last time this happened, Tom disabled parallel tests (https://github.com/covid-projections/covid-data-model/blob/164d80feee77ca2e771ca1d78d27b1677c7d1e55/Makefile#L7) but the problem seems to have come back. 😬

mikelehen:
[Location page color may not match summary file?](https://trello.com/c/lN8IWsqM/1409-location-page-color-may-not-match-summary-file)

mikelehen:
[More Rigorous Sig Figs in the API](https://trello.com/c/jUJfYZkC/792-more-rigorous-sig-figs-in-the-api)

ghop02:
> Note the failing tests are unrelated to my change. One of the tests has been crashing (only in CI) for the last few days. Last time this happened, Tom disabled parallel tests (
> 
> https://github.com/covid-projections/covid-data-model/blob/164d80feee77ca2e771ca1d78d27b1677c7d1e55/Makefile#L7
> 
> ) but the problem seems to have come back. 😬

Ah sad.

Does skipping those two tests that are failing let the tests pass?  If they do - we could skip them in pytest and make a ticket for fixing those..

But not too worried about it now

ghop02 (approved):
This is great! Thank you!

mikelehen:
I'll try disabling the test and see...  

I was also wondering about removing pytest-xdist to see if that solves the crash (or makes it more visible so we know what's going on)...

PR #1139, opened 2021-06-15
Disable crashing test.

mikelehen:
`test_state_providers_smoke_test` started crashing even without parallelization.  I suspect it's running out of memory.  Disabling it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,8 @@ setup-dev: requirements.txt requirements_test.txt
 	pip install --upgrade -r requirements.txt -r requirements_test.txt
 	pre-commit install
 
-# As a work-around for a test failure in tests/libs/datasets/data_source_test.py::test_state_providers_smoke_test
-# run tests on a single process. The failure seems to happen on the server (which runs unittest) but not locally.
 unittest:
-	pytest -n 1 tests/
+	pytest -n 2 tests/
 
 unittest-not-slow:
 	pytest -k 'not slow' -n 2 --durations=5 tests/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,10 +4,8 @@ setup-dev: requirements.txt requirements_test.txt
 	pip install --upgrade -r requirements.txt -r requirements_test.txt
 	pre-commit install
 
-# As a work-around for a test failure in tests/libs/datasets/data_source_test.py::test_state_providers_smoke_test
-# run tests on a single process. The failure seems to happen on the server (which runs unittest) but not locally.
 unittest:
-	pytest -n 1 tests/
+	pytest -n 2 tests/
```
mikelehen, line 8: Note that putting this back to `-n 2` made a pretty trivial difference (6:23 => 6:04 time for CI workflow to run)...  So if the parallelization causes us any headaches in the future, we should probably just nuke it... especially since we're shifting to maintenance mode. 🤷

TomGoBravo (approved):
I think automatic tests are an important monitoring tool that give us the confidence to reduce how closely humans monitor this system. This test is likely to fail when unexpected data changes are likely to break the pipeline. If we're lucky any breakage will be obvious without this test and easy to track down, but I wouldn't count on it. I'm fine with this change until we get a chance to try throwing memory at the problem.

mikelehen:
Definitely a fan of finding a better fix. This was just to unblock CI!

TomGoBravo:
[tests/libs/datasets/data_source_test.py::test_state_providers_smoke_test crashing (running out of memory?)](https://trello.com/c/FFAUKu3k/1429-tests-libs-datasets-datasourcetestpyteststateproviderssmoketest-crashing-running-out-of-memory)

PR #1150, opened 2021-07-15
Disable outlier detection for days following July 4th.

mikelehen:


ghop02 (approved):
LGTM! Not too hung up on it, but a small test would be nice if it's not a bunch of work

PR #1152, opened 2021-07-15
add Daily new cases risk level to csv

ghop02:
Addresses https://trello.com/c/rJUSKX9z

Adds case density risk level to the timeseries json and csv files.

One result of this is that this makes all of the timeseries files bigger, but dont' think it's a huge issue


mikelehen (approved):
LGTM. Thanks!

PR #1154, opened 2021-07-18
Add --refresh-datasets flag to allow resuming from pickle instead of loading from covid-data-public.

mikelehen:


cli/data.py
```diff
@@ -94,28 +94,43 @@ def update_forecasts(filename):
     help="Print summary stats at several places in the pipeline. Producing these takes extra time.",
     default=True,
 )
+@click.option(
+    "--refresh-datasets/--no-refresh-datasets",
```
mikelehen, line 98: Open to other names if you have thoughts.

ghop02 (approved):
do you think it would be helpful as a parameter in github actions? I'm not sure how much time it saves but could be useful when updating region blocks

mikelehen:
Oh, adding it to github actions isn't a bad idea.  I might try to tackle that in a subsequent PR.

PR #1156, opened 2021-07-21
Scale down first day reporting after zeros

ghop02:


smcclure17:
LGTM, although I may not be the most knowledgeable on the issue

PR #1160, opened 2021-07-22
Plumb new CDC county vaccine data in.

mikelehen:


libs/datasets/sources/cdc_new_vaccine_counties_dataset.py
```diff
@@ -0,0 +1,49 @@
...
+            measurement="cumulative",
+            unit="doses",
+            provider="cdc2",
+            common_field=CommonFields.VACCINES_ALLOCATED,
+        ),
+        ccd_helpers.ScraperVariable(
+            variable_name="total_vaccine_distributed",
+            measurement="cumulative",
+            unit="doses",
+            provider="cdc2",
```
ghop02, line 25: Will we be able to remove cdc2 at some point?

ghop02 (approved):

libs/datasets/sources/cdc_new_vaccine_counties_dataset.py
ghop02: or rather rename cdc2 -> cdc

smcclure17 (approved):

libs/datasets/sources/cdc_new_vaccine_counties_dataset.py
mikelehen: cc/ @smcclure17 Not 100% sure what this looks like on the scraper side, but I think we probably could if we want to...  In theory we could make the scraper publish to both `cdc` and `cdc2` for a little while, then migrate this, and then delete the `cdc2` data.

One of the tricky things I'm not sure about is the other vaccine variables (distributed, allocated) that aren't available through `cdc2` right now...

smcclure17: I think we could. The vaccine distributed data comes from the CDC state-level scraper (which is still active), not the old county-vaccine scraper. Also, the new CDC county source has all the same variables as the old CDC county source, but I didn't add them as I didn't think we used/wanted them.

PR #1161, opened 2021-07-29
Add cdc transmission levels

ghop02:
This adds the latest cdc transmission level to the json adn csv output.

Note that this does not add to the timeseries data.  it's not a ton more work, but figured this was the minimum steps to get it working.

### DIFF
--- a/libs/metrics/cdc_transmission_levels.py
+++ b/libs/metrics/cdc_transmission_levels.py
@@ -70,7 +70,7 @@ def top_level_transmission_level(
     elif TransmissionLevel.UNKNOWN in level_list:
         return TransmissionLevel.UNKNOWN
 
-    return RiskLevel.LOW
+    return TransmissionLevel.LOW
 
 
 def calculate_transmission_level_from_metrics(
--- a/libs/metrics/cdc_transmission_levels.py
+++ b/libs/metrics/cdc_transmission_levels.py
@@ -87,5 +87,5 @@ def calculate_transmission_level_from_metrics(
     case_density_level = case_density_transmission_level(metrics.caseDensity)
     test_positivity_level = test_positivity_transmission_level(metrics.testPositivityRatio)
 
-    overall_level = top_level_transmission_level(case_density_level, test_positivity_level)
+    overall_level = overall_transmission_level(case_density_level, test_positivity_level)
     return overall_level
--- a/libs/metrics/cdc_transmission_levels.py
+++ b/libs/metrics/cdc_transmission_levels.py
@@ -1,0 +1,112 @@
+from typing import List, Optional
+import math
+import numpy as np
+import pandas as pd
+
+from covidactnow.datapublic.common_fields import CommonFields
+from libs.metrics.top_level_metrics import MetricsFields
+from libs.metrics import top_level_metrics
+from api import can_api_v2_definition
+
+TransmissionLevel = can_api_v2_definition.CDCTransmissionLevel
+
+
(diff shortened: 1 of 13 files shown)


ghop02:
Test build with the current cdc transmission levels is [this run](https://github.com/covid-projections/covid-data-model/actions/runs/1077325810)

libs/metrics/cdc_transmission_levels.py
```diff
@@ -0,0 +1,115 @@
...
+from libs.metrics import top_level_metrics
+from api import can_api_v2_definition
+
+TransmissionLevel = can_api_v2_definition.CDCTransmissionLevel
+
+
+def calc_transmission_level(value: Optional[float], thresholds: List[float]) -> TransmissionLevel:
+    """Check the value against thresholds to determine the transmission level for the metric.
+
+    Each threshold is the upper limit for the risk level.
```
mikelehen, line 17: nit: risk => transmission everywhere.

libs/metrics/cdc_transmission_levels.py
```diff
@@ -0,0 +1,115 @@
...
+    thresholds = [10 / 7.0, 50 / 7.0, 100 / 7.0]
+    return calc_transmission_level(value, thresholds)
+
+
+def test_positivity_transmission_level(value: float) -> TransmissionLevel:
+    thresholds = [0.05, 0.08, 0.10]
+    return calc_transmission_level(value, thresholds)
+
+
+def top_level_transmission_level(
```
mikelehen, line 57: top_level => overall for consistency with our overall risk level?

libs/metrics/cdc_transmission_levels.py
```diff
@@ -0,0 +1,115 @@
...
+
+def calculate_transmission_level_timeseries(
+    metrics_df: pd.DataFrame, metric_max_lookback=top_level_metrics.MAX_METRIC_LOOKBACK_DAYS
+):
+    metrics_df = metrics_df.copy()
+    # Shift infection rate to align with infection rate delay
+    infection_rate = metrics_df[MetricsFields.INFECTION_RATE].shift(
+        top_level_metrics.RT_TRUNCATION_DAYS
+    )
+    metrics_df[MetricsFields.INFECTION_RATE] = infection_rate
```
mikelehen, line 105: Uh, do we need this infection_rate stuff?

libs/pipelines/api_v2_pipeline.py
```diff
@@ -140,11 +141,14 @@ def build_timeseries_for_region(
             metrics_results
         )
         risk_levels = top_level_metric_risk_levels.calculate_risk_level_from_metrics(metrics_latest)
+        transmission_level = cdc_transmission_levels.calculate_transmission_level_from_metrics(
```
mikelehen, line 144: cdc_transmission_level just to be 100% clear?

mikelehen (approved):
Looks good!  I left some comments, but they're mostly nits.

Spot checking some counties, it seems like this is working right.  The main place we differ is they don't have UNKNOWN so in places that haven't reported test positivity recently, we have them as UNKNOWN but CDC often has them as LOW.  I think that's okay.  Nice work!

libs/metrics/cdc_transmission_levels.py
ghop02: No - I can actually delete these functions because we're not doing timeseries transmissions levels for now (would be easy enough to add in in the future if we want)

PR #1162, opened 2021-08-01
reduced vaccine estimation backfill threshold by one day

smcclure17:
Reduces the number of days until the 1+ dose estimation kicks in so that estimation will start immediately once the data would go stale. This should prevent counties that are switching to estimation from going gray/stale for one day in between. 

### Please Check
- [ ] Are tests passing?


mikelehen (approved):
Thank you!

PR #1165, opened 2021-08-10
Update definition to lower priority of unknown + update docs

ghop02:
Moves down the priority of an unknown value. If one of the values is low, it will prefer that over unknown values.

Also improves product update docs to give some more specifics

### DIFF
--- a/api/docs/docs/updates.md
+++ b/api/docs/docs/updates.md
@@ -15,6 +15,22 @@ You can now access the current CDC Community transmissions levels in the Covid A
 The CDC community transmission levels are similar to the Covid Act Now risk levels, but have slightly different thresholds.  
 See [definitions of CDC community transmission levels](https://covid.cdc.gov/covid-data-tracker/#cases_community) for more details.
 
+You can find CDC community transmission levels in the field ``cdcTransmissionLevel`` in all API responses.
+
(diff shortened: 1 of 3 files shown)


api/docs/docs/updates.md
```diff
@@ -15,6 +15,22 @@ You can now access the current CDC Community transmissions levels in the Covid A
...
+| ------------ | ------------------------------- |
+| 0 | Low |
+| 1 | Moderate |
+| 2 | Substantial |
+| 3 | High |
+| 4 | Unknown |
+
+Note that the CDC does not include an "unknown" transmission level.
+We chose to include in cases where either test positivity or case data
+is stale for more than 15 days.
```
mikelehen, line 32: This doesn't reflect the change you just made, right?  I think it should be something like:

> Note that we have introduced an "Unknown" level for when both case data and test positivity data are missing for at least 15 days.  The CDC does not have an "Unknown" level and instead will designate a location as "Low" when case and test positivity data are missing.

mikelehen (approved):

smcclure17 (approved):
lgtm!

api/docs/docs/updates.md
ghop02: 👍

PR #1176, opened 2021-09-14
Remove CovidTrackingSource.

mikelehen:
Covid Tracking stopped publishing data last year so it's useless, but we still used it for columns we don't care about (cumulative hospitalized, etc.) and as a backup for HHS-provided hospital data (which we don't need because HHS covers all states).


BrettBoval (approved):
LGTM. I'm assuming these removals might break the deprecated ICU Headroom code, but Michael says that he has a separate PR to remove that code too.

PR #1178, opened 2021-09-15
Regenerate test combined datasets data (and fix tests).

mikelehen:
This is the result of running `./run.py data update-test-combined-data` to update our test combined datasets file (basically just a subset of the full one to make it easier for tests to load) now that we have staffed_beds in it.  But getting new data required me to update some tests.

### DIFF
--- a/tests/libs/build_api_v2_test.py
+++ b/tests/libs/build_api_v2_test.py
@@ -62,6 +62,11 @@ def test_build_summary_for_fips(
         type=FieldSourceType.OTHER,
         url="https://healthdata.gov/Hospital/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/anag-cw7u",
     )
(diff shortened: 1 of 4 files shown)


tests/libs/build_api_v2_test.py
```diff
@@ -111,13 +116,13 @@ def test_build_summary_for_fips(
             newDeaths=FieldAnnotations(
                 anomalies=[
                     {
-                        "date": datetime.date(2020, 4, 14),
-                        "original_observation": 345.0,
+                        "date": datetime.date(2021, 3, 13),
+                        "original_observation": 110.0,
```
mikelehen, line 120: Not 100% sure why these outliers changed in this way (maybe NYT corrected some of their underlying data or something).

smcclure17 (approved):

PR #1179, opened 2021-09-15
remove ND from CDC exclusion list

smcclure17:
This removes ND from the exclusion list such that the pipeline will now use CDC data. 

It looks like the ND scraper was previously collecting `% of 12+ population vaccinated` as `% of total population vaccinated` causing our data to be inflated. 

For example Cass County has a population of 185,181,  and is [reporting 104,220 people with at least one dose](https://www.health.nd.gov/covid19vaccine/dashboard): `104,220 / 185,181 = 0.56`, which fairly closely lines up with the [CDC's reported value](https://data.cdc.gov/Vaccinations/COVID-19-Vaccinations-in-the-United-States-County/8xkx-amqh/data) of 54%. Previously we were reporting over 60% (we we're multiplying the `12 and older at least one dose` percentage by the total population). 

I think it makes sense to:
- delete the incorrect historical data
- switch to the CDC source such that we surface the historical data (and because that data appears to be of good/fair quality)

---
screenshot of Cass County dashboard:
image

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -177,10 +177,6 @@ CDCVaccinesStatesAndNationDataset = datasource_regions(
 )
 
 CDC_COUNTY_EXCLUSIONS = [
-    # Glacier County, MT - reports 99.9% of 12+ population vaccinated [7/2021]
-    Region.from_fips("30035"),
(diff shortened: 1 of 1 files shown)


mikelehen (approved):
Sounds good. Thanks for digging into it!

PR #1180, opened 2021-09-16
Add STAFFED_BEDS to combined datasets.

mikelehen:
We currently expose licensed_beds from CovidCareMap in our API (as hospitalBeds.capacity) even though we know this data isn't very high-quality. We'd like to switch to using staffed_beds from the HHS hospital data (and get rid of CovidCareMap), but we currently don't include it in our combined dataset files.

This PR fixes that.

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -261,6 +261,11 @@ ALL_TIMESERIES_FEATURE_DEFINITION: FeatureDataSourceMap = {
         HHSHospitalCountyDataset,
         HHSHospitalStateDataset,
     ],
+    CommonFields.STAFFED_BEDS: [
+        CANScraperStateProviders,
+        HHSHospitalCountyDataset,
+        HHSHospitalStateDataset,
+    ],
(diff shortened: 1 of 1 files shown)


libs/datasets/combined_datasets.py
```diff
@@ -261,6 +261,11 @@ def datasource_regions(
         HHSHospitalCountyDataset,
         HHSHospitalStateDataset,
     ],
+    CommonFields.STAFFED_BEDS: [
+        CANScraperStateProviders,
```
mikelehen, line 265: TBH I'm not 100% sure if we want to use CANScraperStateProviders, but I just made this match ICU_BEDS above. Chances are HHS covers all states anyway, so it'll never get used.

smcclure17: I think we removed most (or all) of our non-vaccine related state level scrapers, so I'm not 100% sure that there would be state-provided data to fall back on (I could check the database to verify that), but I do think it's good to leave in

smcclure17 (approved):

PR #1181, opened 2021-09-20
remove IL from CDC vaccine exclusion list

smcclure17:


mikelehen (approved):
🙏

mikelehen:
I think we have a stale disclaimer that can/should definitely go now:
> Region: IL
> subregions
> metrics.vaccinationsInitiatedRatio
> https://trello.com/c/buXoYmKy/1140-il-county-first-doses-underreporting
> Note that Illinois does not report enough information to correctly distinguish between administration of one dose vaccines and two dose vaccines, and therefore our "1+ dose" data above is approximated using the fully-vaccinated data and a state-wide average ratio.

smcclure17:
ah, yep, I'll remove that from the CMS

PR #1183, opened 2021-09-22
Relabel unused columns to be unique.

mikelehen:
Addresses the concern a couple API users have expressed about "unused" API columns being duplicated.

### DIFF
--- a/libs/pipelines/api_v2_pipeline.py
+++ b/libs/pipelines/api_v2_pipeline.py
@@ -267,13 +267,6 @@ def _model_to_dict(data: dict):
         if isinstance(value, pydantic.BaseModel):
             value = _model_to_dict(value.__dict__)
 
-            # When converting API model objects to dicts, we always add an
-            # "unused" entry to the dict. This allows us to use "unused" as a
-            # column in our CSV output to have a blank column. This is handy
-            # since when we remove an API column we replace it with "unused" to
-            # preserve column ordering).
-            results["unused"] = None
-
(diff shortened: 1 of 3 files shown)


libs/pipelines/api_v2_pipeline.py
```diff
@@ -267,13 +267,6 @@ def _model_to_dict(data: dict):
         if isinstance(value, pydantic.BaseModel):
             value = _model_to_dict(value.__dict__)
 
-            # When converting API model objects to dicts, we always add an
-            # "unused" entry to the dict. This allows us to use "unused" as a
-            # column in our CSV output to have a blank column. This is handy
-            # since when we remove an API column we replace it with "unused" to
-            # preserve column ordering).
-            results["unused"] = None
```
mikelehen, line 275: So it turns out this code was just unnecessary.  I thought there was code down the pipeline that would barf if "unused" wasn't in these dicts, but I was mistaken.  All the tests pass and I built some API endpoints and verified the CSV files correctly contain the column with empty entries for all rows.

smcclure17:
lgtm! thanks for hopping on this 🙏 

It might be nice to have an API snapshot/test build with this to verify it, but I trust you've tested it thoroughly so no real need at all.

🚢 🚢

smcclure17 (approved):

mikelehen:
Yeah, I am pretty sure I manually tested this well enough but I'll kick off a snapshot run to sanity-check!

PR #1186, opened 2021-09-28
remove NE cases from NYT exclusion list

smcclure17:
NE has restarted reporting county-level case data. This removes the NE filter on the NYT cases dataset, overriding the State Scraper source. 

USAFacts did not appear to have the case data, so I've left the filter in place for that dataset

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -150,22 +150,15 @@ JOPLIN_COUNTIES = [
     Region.from_fips("29145"),
 ]
 
-NE_COUNTIES = RegionMask(AggregationLevel.COUNTY, states=["NE"])
-
 # NY Times has cases and deaths for all boroughs aggregated into 36061 / New York County.
 # Remove all the NYC data so that USAFacts (which reports each borough separately) is used.
 # Remove counties in MO that overlap with Kansas City and Joplin because we don't handle the
 # reporting done by city, as documented at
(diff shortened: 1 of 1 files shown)


data/multiregion.json
```diff
@@ -7,9 +7,9 @@
     "is_dirty": false
   },
   "model_git_info": {
-    "sha": "3c234319406d624ec66830fe4a656bc5253cb49d",
-    "branch": "main",
```
mikelehen, line 11: Do you mind reverting the `data/` changes and then just kicking off a new dataset update from main once this is merged?

Probably doesn't really matter but as a matter of principle it's nice to always be building our data / snapshots from `main` bits.

mikelehen (changes requested):
Do you mind removing `NE_COUNTIES` and `CANScraperUSAFactsProviderWithoutNe`?

Even though USA Facts doesn't have NE data back (yet), it won't hurt anything to stop excluding it, since NYT is higher-priority and has all NE counties.  So I'd rather go ahead and clean this all up now else it'll be here forever and we'll forget why it existed (especially since there don't seem to be any existing comments about the NE exclusions 😬 ).

data/multiregion.json
smcclure17: yep! will do, also I don't think we would want this data as it includes the backlogged data

mikelehen (approved):
LGTM!

PR #1193, opened 2021-11-03
add and prioritize new source for CDC testing data

smcclure17:


mikelehen (approved):
Looks good!

libs/datasets/data_source.py
```diff
@@ -151,7 +151,7 @@ def make_dataset(cls) -> timeseries.MultiRegionDataset:
             source_type=cls.SOURCE_TYPE,
         )
         data = rows.unstack(PdFields.VARIABLE)
-        data = cls._check_and_removed_unexpected_data(data)
+        data = cls._check_and_removed_unexpected_data(data).sort_index(ascending=True)
```
smcclure17, line 154: I was getting an error saying that the datetime index of the dataset was out of order (non-increasing). I'm not sure why this seems to only affect this new dataset, but explicitly sorting the index here solves the issue.

mikelehen: Hrm, yeah. That's weird, but seems like a fine fix. 🤷

PR #1194, opened 2021-11-15
remove NE county region mask from NYT and USA Facts sources

smcclure17:
The NE case scraper seems to be back after being down mid and late last week, but since NYT now has true county-level case data I think it makes sense to use that. 

This leaves the region mask and the filtered USA Facts source in the code, as it seems this might happen again in the future. 

Let me know if you'd like to handle this in any other way, or if you have any suggestions/changes in mind!

Also, not sure why the lint/formatting is failing, but I can look into it

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -150,9 +150,12 @@ JOPLIN_COUNTIES = [
     Region.from_fips("29145"),
 ]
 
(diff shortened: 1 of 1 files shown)


mikelehen (approved):
Looks great.  The lint failure seems to be the broken black github action that we ran into with BBE.  We probably should migrate to the official black action like I did over there.  I can do that later today if you don't beat me to it.

PR #1195, opened 2021-11-15
Switch to official psf/black@stable github action

smcclure17:
mirrors the changes made in https://github.com/rewiringamerica/build-back-electric/pull/174/files


mikelehen (approved):
Awesome, thanks!

PR #1197, opened 2021-11-17
Block CDC 1st dose data for NE counties.

mikelehen:
See https://covid-projections-git-bump-snapshot-2560-1471-covidactnow.vercel.app/internal/compare/?left=2558&locations=NE&metric=6&right=2560&sort=3 for a snapshot built with these changes.


smcclure17 (approved):
looks great! less disruptive than I anticipated

sorry for the delay 😅 thanks for this!

PR #1198, opened 2021-11-17
Reduce parallelization to prevent model build from crashing.

mikelehen:


smcclure17 (approved):

libs/parallel_utils.py
```diff
@@ -33,7 +33,10 @@ def parallel_map(func: Callable[[T], R], iterable: Iterable[T]) -> Iterable[R]:
     if USE_MULTIPROCESSING:
         # Setting maxtasksperchild to one ensures that we minimize memory usage over time by creating
         # a new child for every task. Addresses OOMs we saw on highly parallel build machine.
-        with multiprocessing.Pool(maxtasksperchild=1) as pool:
+        # But that might not be enough. Also make sure we don't spawn more than 64 processes (the
+        # build machine is 96-core)
+        processes = min(os.cpu_count(), 64)
```
smcclure17, line 38: just for curiosity, is there a specific reason for choosing 64 or is it arbitrary?

mikelehen: Pretty arbitrary.  I figure 33% headroom felt reasonable. 🤷

mikelehen: FWIW- I wouldn't be surprised if this parallelization isn't all that productive (especially with maxtasksperchild=1 which I think means we'll be spawning / killing processes a ton) and we could optimize this much better if we did some investigation.  But not worth worrying about for now.  Just want the build to be reliable. 😁

smcclure17: sounds good! that does make sense

PR #1200, opened 2021-11-29
remove demographic data from CMS blocks

smcclure17:
I kicked off an action to update the datasets with these changes here: https://github.com/covid-projections/covid-data-model/actions/runs/1517344858 and will verify with a snapshot before merging

### DIFF
--- a/libs/datasets/manual_filter.py
+++ b/libs/datasets/manual_filter.py
@@ -83,7 +83,13 @@ def _partition_by_fields(
 ) -> Tuple[pd.DataFrame, pd.DataFrame]:
     """Partitions a DataFrame with timeseries wide dates into two: fields and others."""
     timeseries.check_timeseries_wide_dates_structure(ts_in)
-    mask_selected_fields = ts_in.index.get_level_values(PdFields.VARIABLE).isin(fields)
+
+    # NOTE(sean): By default we only block non-demographic data. The demographic
+    # data does not reach the website, but is used by partners so we want this to flow through,
+    # regardless of block status.
+    mask_selected_fields = ts_in.index.get_level_values(PdFields.VARIABLE).isin(
+        fields
+    ) & ts_in.index.get_level_values(PdFields.DEMOGRAPHIC_BUCKET).isin(["all"])
     ts_selected_fields = ts_in.loc[mask_selected_fields]
     ts_not_selected_fields = ts_in.loc[~mask_selected_fields]
     return ts_selected_fields, ts_not_selected_fields
--- a/libs/datasets/manual_filter.py
+++ b/libs/datasets/manual_filter.py
@@ -85,7 +85,7 @@ def _partition_by_fields(
     timeseries.check_timeseries_wide_dates_structure(ts_in)
 
     # NOTE(sean): By default we only block non-demographic data. The demographic
-    # data does not reach the website, but is used by partners so we want this to flow through,
+    # data does not reach the website, but is used by partners so we want this to flow through
     # regardless of block status.
     mask_selected_fields = ts_in.index.get_level_values(PdFields.VARIABLE).isin(
         fields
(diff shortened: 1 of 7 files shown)


libs/datasets/manual_filter.py
```diff
@@ -83,7 +83,13 @@ def _partition_by_fields(
...
     """Partitions a DataFrame with timeseries wide dates into two: fields and others."""
     timeseries.check_timeseries_wide_dates_structure(ts_in)
-    mask_selected_fields = ts_in.index.get_level_values(PdFields.VARIABLE).isin(fields)
+
+    # NOTE(sean): By default we only block non-demographic data. The demographic
+    # data does not reach the website, but is used by partners so we want this to flow through
+    # regardless of block status.
+    mask_selected_fields = ts_in.index.get_level_values(PdFields.VARIABLE).isin(
+        fields
+    ) & ts_in.index.get_level_values(PdFields.DEMOGRAPHIC_BUCKET).isin(["all"])
```
smcclure17, line 92: there doesn't seem to be a pandas function for comparing index values to one value, (like [`Series.eq()`](https://pandas.pydata.org/docs/reference/api/pandas.Series.eq.html)). This could probably be done with boolean indexing, but I think I prefer this as it keeps the syntax consistent with what's above. 

If we have strong feelings though, I can look for a way to compare the index to one value directly

tests/libs/datasets/manual_filter_test.py
```diff
@@ -165,13 +165,16 @@ def test_manual_filter_per_bucket_tag():
     region = Region.from_fips("40031")
```
smcclure17, line 165: had to update a couple of tests b/c the behavior of the filters changed

tests/libs/datasets/manual_filter_test.py
```diff
@@ -165,13 +165,16 @@ def test_manual_filter_per_bucket_tag():
...
     }
-    # kids have real values that will be removed so will have a tag. guys do not and are not
-    # expected to get a tag.
+
+    # NOTE(sean): as of 11-29-2021:
+    # demographic buckets have data but will not be removed, and should have no tag.
+    # non-demographic bucket will have data removed and should have a tag
     ds_in = test_helpers.build_default_region_dataset(
-        {CommonFields.CASES: {kids: [1, 2, 3], guys: [4]}, **other_data},
+        {CommonFields.CASES: {kids: [1, 2, 3], guys: [4], all_: [5, 6, 7]}, **other_data},
```
mikelehen, line 177: I assume this is just me not understanding this codebase that well, but why does "all_" end up with an underscore here?  Is this a test-only thing or something?

tests/libs/datasets/manual_filter_test.py
```diff
@@ -274,12 +277,14 @@ def test_region_overrides_transform_and_filter_blocked_false():
         taglib.TagType.KNOWN_ISSUE_NO_DATE,
         public_note=region_overrides["overrides"][0]["disclaimer"],
     )
+    # NOTE(sean): as of 11-29:
```
mikelehen, line 280: nit: Can you add -2021 ?

mikelehen (approved):
Nice!  I was worried this would be trickier.

Once we've validated the results, can you revert the dataset update before merging (and then you can kick off an official one from main after merging)?

tests/libs/datasets/manual_filter_test.py
smcclure17: `all()` is a built-in function so I added the `_` just to not mix up the name of the variable with the function, so there's no specific reason (we could rename this var to `all_demographics` instead or something)

mikelehen: Ohhh.  Gotcha'. That makes perfect sense.

smcclure17:
🙏 

kicked off https://github.com/covid-projections/covid-data-model/actions/runs/1517699872 to make a snapshot

smcclure17:
Spot checking it appears to be working as expected: 

the `all` demographics bucket is blocked, but none of the others are. 
image

the vaccination data is blocked, but no other data is: 
image
(Nov 18th was the last time ICU/hospital data was reported for this county)

Still will verify with https://github.com/covid-projections/covid-data-model/actions/runs/1517699872 when it finishes

PR #1202, opened 2021-11-30
Make data update pull right before pushing in case other changes were committed.

mikelehen:
Should avoid some of the recent failures.

### DIFF
--- a/tools/push-data-update.sh
+++ b/tools/push-data-update.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Do a fresh pull in case any changes have been pushed to the branch.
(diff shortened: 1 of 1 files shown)


tools/push-data-update.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Do a fresh pull in case any changes have been pushed to the branch.
+git pull "https://${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:${GITHUB_REF}
```
mikelehen, line 4: Note that this is a copy/paste of the below `push` command but changed to a `pull` instead, so I think it should be correct. 😁

smcclure17 (approved):
🙏

PR #1215, opened 2022-01-05
Don't remove trailing zeros on an all-zero series

smcclure17:
If the timeseries is all zeros then `last_nonzero_index` is `None`, causing the comparison on line 164 to fail. 

This change means that we will never truncate the trailing zeros for an all-zero series (even if the series is shorter than `stall_length`)

this addresses the sentry error: https://sentry.io/organizations/covidactnow/issues/2907283165/?environment=Environment.PRODUCTION&project=5203044&referrer=alert_email


mikelehen (approved):
Thanks!

PR #1216, opened 2022-01-06
Cap vaccination rates at 95%

smcclure17:
This only caps the vaccination metric/percentage and leaves the actual value untouched. I can update this to cap the cumulative values, too, if that's preferred


mikelehen (approved):
Looks good! Thanks for updating the tests too.

PR #1218, opened 2022-01-12
Migrate NYT and HHS sources from covid-data-public to can-scrapers

smcclure17:
[snapshot 299](https://covid-projections-git-bump-snapshot-2801-1692-covidactnow.vercel.app/internal/compare/?left=2799&locations=0&metric=5&right=2801&sort=3) was generated off this branch and appears to have not caused any unwanted changes 

I think the `test_pyseir_end_to_end_idaho` test is failing because the Idaho case data is blocked, which means there's no data to calculate Rt with (it passes if I substitute a different FIPS like Los Angeles, 06037)


mikelehen (approved):
Looks great! Thank you.

Before we merge this we may want to think about our build scheduling to try to make sure the parquet file is updated by the time our builds run.  Basically, if NYT updates at 10:10 PM PT, when will that make it into a parquet file?  And then try to make sure we have a build scheduled to pick it up (with a bit of cushion but not too much 😁 )...

I'm guessing we're going to end up with our first snapshot with NYT's updated data being a bit delayed compared to what it was in the past.  We should probably give Tom a heads-up when that happens so he can keep an eye out for API feedback.

Thanks!

mikelehen:
Hrm, yeah. I'm surprised the NYT scraper takes an hour to insert its data.  Might be worth looking into that at some point.  One thing I've wondered is if Postgres is the best option for us.  Given we're not doing any low-latency, optimized queries or anything, would BigQuery fit our needs better?  Not at all sure but it might be worth looking into.

In any case, your plan sounds good to me!

smcclure17:
Yeah, I agree that there are probably better options for this--worth looking into. 


142b5da22fa689a000251d6742fe6b1b0b87d181 adds a scheduler to run the pipeline daily and sets the default behavior to trigger the API build

cbea414f8b9fb5a31be11c6d1643391f4e60f242 persists the parquet file and sets the pipeline to read the file from local storage. This is mainly to give us visibility into what Parquet file was used for each snapshot. I went this route mostly because it didn't require making changes in `can-scrapers`. 

I'm more than willing to update this to pull from the most recent timestamped file in GCS and log the name of the file (per our slack conversation) if you'd prefer that.

libs/datasets/sources/can_scraper_helpers.py
```diff
@@ -21,7 +22,11 @@
 
 # Airflow jobs output a single parquet file with all of the data - this is where
 # it is currently stored.
-PARQUET_PATH = "data/can-scrape/can_scrape_api_covid_us.parquet"
+GCS_PARQUET_PATH = (
+    "https://storage.googleapis.com/can-scrape-outputs/final/can_scrape_api_covid_us.parquet"
+)
+
+LOCAL_PARQUET_PATH = dataset_utils.DATA_DIRECTORY / "can_scrape_api_covid_us.parquet"
```
mikelehen, line 29: I think we'll need to configure *.parquet files to be stored in LFS in .gitattributes.

Maybe go ahead and check the current parquet file in with this PR to make sure that it ends up in LFS?

mikelehen (approved):
This looks good but let's make sure the parquet file ends up in LFS.  We should probably disable the github actions in covid-data-public at the same time we merge this so it's not kicking off ill-timed runs that compete for resources with the new scheduled runs.

I'm a tiny bit tempted to schedule more than one build (e.g. 8 UTC and 10 or 11 UTC) so that there's a backup build in case we screw up the timing for the first one, but up to you if you feel confident about this. 😁

libs/datasets/sources/can_scraper_helpers.py
smcclure17: good call,

added this and a second pipeline run at 10 UTC. The `git push` with LFS is taking a long time :/

PR #1219, opened 2022-01-12
covid-data-public migration

smcclure17:
- Migrates `covid-data-public/data/misc` into this repo
- Removes references to `covid-data-public`, and now saves datasets in the current repo instead. 

Next PR(s) will:
- Migrate `covid-data-public/covidactnow/datapublic` (the Common Fields and any other necessary scripts)
- Update the Github Actions to get rid of covid-data-public


libs/datasets/data_version.py
```diff
@@ -1,104 +0,0 @@
-import click
```
smcclure17, line 1: This code works with the covid-data-public repo, but it is never used anywhere, so I've removed it here

libs/notebook_helpers.py
```diff
@@ -1,32 +1,13 @@
...
 import pandas as pd
 
-from libs.datasets import dataset_utils
 from libs.datasets.sources import nytimes_dataset
 from libs.datasets.sources import can_scraper_local_dashboard_providers
 
 _logger = logging.getLogger(__name__)
 
 
-def set_covid_data_public():
```
smcclure17, line 13: This further depreciates some of the notebooks, but the notebooks that use this method don't appear to be functional anyway. if we'd like, I can try to update the notebooks to work with this update, but I'm not sure that's worth the effort

libs/datasets/can_raw_location_page_urls.py
```diff
@@ -0,0 +1,67 @@
+import structlog
```
mikelehen, line 1: WDYT about creating a top-level `scripts` directory like covid-data-public had?  That way it's clearer that this is a standalone script that's not part of the main data pipeline code?  Putting it under `libs/datasets` seems a bit confusing to me.  Open to other ideas though.

libs/notebook_helpers.py
mikelehen: Nah, definitely don't bother.  I've never used the notebooks. I'm not sure who did TBH.  I think we could just delete them if they get in our way. 🤷

mikelehen (approved):
Nice, looks good!

libs/datasets/can_raw_location_page_urls.py
smcclure17: yeah that sgtm 🙏 , done in e5357366e03317c325c11282651a25b5905dcbd5

libs/notebook_helpers.py
smcclure17: yeah sounds good--it seems like they were used for exploratory purposes--I've never looked at them either haha

PR #1223, opened 2022-01-18
trigger API build on scheduled Update combined datasets actions

smcclure17:
The scheduled runs are not triggering the API build because `trigger_api_build` is only defined using `workflow_dispatch`. This should default to `true` if no `trigger_api_build` input exists (according to https://github.community/t/how-can-you-use-expressions-as-the-workflow-dispatch-input-default/141454/2)

Not sure of a great way to test this, as I'm not sure if scheduled runs will be triggered on non-main branches. I'll take a quick look into this


mikelehen (approved):
Ah, good catch. Thanks!

PR #1224, opened 2022-01-18
Combine historical and as-origninally-posted CDC test positivity datasets 

smcclure17:
Essentially creates a dataset for both sources then merges them. 

This uses the historical data wherever possible, and appends/combines any additional days from the as-posted data. Typically, the historical data lags behind by about 3 days, so this will use the as-originally-posted data for those 3 most recent days. This uses the as-originally-posted dataset URL and name for the tags.

As an aside, I moved the `persist_parquet()` call to run once when `./run.py data update` is run. Where I previously had  it, it would execute every time a can-scraper data source was loaded, which is unnecessary

### DIFF
--- a/libs/datasets/data_source.py
+++ b/libs/datasets/data_source.py
@@ -137,8 +137,6 @@ class CanScraperBase(DataSource, abc.ABC, metaclass=_CanScraperBaseMeta):
     @staticmethod
     @lru_cache(None)
     def _get_covid_county_dataset() -> ccd_helpers.CanScraperLoader:
-        # fetch and persist Parquet file from GCS, then load it
-        ccd_helpers.CanScraperLoader.persist_parquet()
         return ccd_helpers.CanScraperLoader.load_from_local()
 
     @classmethod
--- a/libs/datasets/sources/cdc_testing_dataset.py
+++ b/libs/datasets/sources/cdc_testing_dataset.py
@@ -4,6 +4,7 @@ from functools import lru_cache
 from libs.datasets import AggregationLevel
 from libs.datasets import data_source
 import pandas as pd
+import abc
 from covidactnow.datapublic.common_fields import CommonFields
 from libs.datasets.sources import can_scraper_helpers as ccd_helpers
 from libs.datasets.timeseries import MultiRegionDataset
@@ -40,7 +41,16 @@ def remove_trailing_zeros(data: pd.DataFrame) -> pd.DataFrame:
     return data
 
 
-class CDCHistoricalTestingDataset(data_source.CanScraperBase):
+class CDCTestingBase(data_source.CanScraperBase, abc.ABC):
+    """Default CanScraperBase class with make_dataset() overridden to handle CDC test data."""
+
+    @classmethod
+    @lru_cache(None)
+    def make_dataset(cls) -> MultiRegionDataset:
+        return modify_dataset(super().make_dataset())
+
+
+class CDCHistoricalTestingDataset(CDCTestingBase):
     """Data source connecting to the official CDC historical test positivity dataset."""
 
     SOURCE_TYPE = "CDCHistoricalTesting"
@@ -55,16 +65,8 @@ class CDCHistoricalTestingDataset(data_source.CanScraperBase):
         ),
     ]
 
-    @classmethod
-    @lru_cache(None)
-    def make_dataset(cls) -> MultiRegionDataset:
-        ds = super().make_dataset()
-        return dataclasses.replace(
-            ds, timeseries=remove_trailing_zeros(ds.timeseries), timeseries_bucketed=None
-        )
 
-
-class CDCOriginallyPostedTestingDataset(CDCHistoricalTestingDataset):
+class CDCOriginallyPostedTestingDataset(CDCTestingBase):
     """Data source connecting to the official as-originally-posted CDC test positivity dataset."""
 
     SOURCE_TYPE = "CDCOriginallyPostedTesting"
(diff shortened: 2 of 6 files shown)


libs/datasets/sources/cdc_testing_dataset.py
```diff
@@ -58,7 +58,60 @@ class CDCHistoricalTestingDataset(data_source.CanScraperBase):
...
+
+        historical_ts = historical_ds.timeseries
+        as_posted_ts = as_posted_ds.timeseries
+
+        merged_ts = historical_ts.combine_first(as_posted_ts)
+
+        merged_ds = dataclasses.replace(
+            as_posted_ds, timeseries=merged_ts, timeseries_bucketed=None,
+        )
+        return modify_dataset(merged_ds)
```
mikelehen, line 114: Can we call `modify_dataset()` directly in `CDCOriginallyPostedTestingDataset` and `CDCHistoricalTestingDataset` so that they behave as "valid" data sources? (e.g. with test positivity as a ratio instead of percent).  Then if we needed to we could safely switch to `CDCHistoricalTestingDataset` as our source of choice in the future without having to modify this code.

libs/datasets/sources/cdc_testing_dataset.py
```diff
@@ -58,7 +58,60 @@ class CDCHistoricalTestingDataset(data_source.CanScraperBase):
...
     @lru_cache(None)
     def make_dataset(cls) -> MultiRegionDataset:
-        return modify_dataset(super().make_dataset())
+        ds = super().make_dataset()
+        return dataclasses.replace(
+            ds, timeseries=remove_trailing_zeros(ds.timeseries), timeseries_bucketed=None
+        )
+
+
+class CDCOriginallyPostedTestingDataset(CDCHistoricalTestingDataset):
```
mikelehen, line 67: Not a big deal and maybe this is common in python, but it feels like a bit of a hack to use inheritance like this as a means of sharing code even though there isn't a clear parent class / subclass relationship here.  I think it would be cleaner to create a separate `CDCTestingDatasetBase` base class or something.

libs/datasets/data_source.py
```diff
@@ -137,8 +137,6 @@ class CanScraperBase(DataSource, abc.ABC, metaclass=_CanScraperBaseMeta):
     @staticmethod
     @lru_cache(None)
     def _get_covid_county_dataset() -> ccd_helpers.CanScraperLoader:
-        # fetch and persist Parquet file from GCS, then load it
-        ccd_helpers.CanScraperLoader.persist_parquet()
```
mikelehen, line 141: Nice catch!

tests/libs/datasets/sources/cdc_testing_dataset_test.py
```diff
@@ -23,7 +23,7 @@ def test_cdc_testing_modify_dataset():
     ds_out = cdc_testing_dataset.modify_dataset(ds_in)
 
     dc_data_after_filter = {
-        CommonFields.TEST_POSITIVITY_7D: [0.5, 0.6, None, None, None, None],
+        CommonFields.TEST_POSITIVITY_7D: [0.5, 0.6, 0, 0, 0, 0],
```
mikelehen, line 26: This test looks kind of silly now, since it used to be testing that trailing zeros are removed (among other things I guess) but now it's just got a bunch of trailing zeros for no reason, I guess.

Maybe we can put this test back if we make my other suggested changes?  Else perhaps we should either modify this test to remove the trailing zeros or add a comment explaining that we used to test removal of trailing zeros but no longer are doing that. 🤷

libs/datasets/sources/cdc_testing_dataset.py
```diff
@@ -87,6 +140,4 @@ def modify_dataset(ds: MultiRegionDataset) -> MultiRegionDataset:
 
     ts_copy = ts_copy.append(dc_results, verify_integrity=True).sort_index()
 
-    return dataclasses.replace(
-        ds, timeseries=remove_trailing_zeros(ts_copy), timeseries_bucketed=None
-    )
+    return dataclasses.replace(ds, timeseries=ts_copy, timeseries_bucketed=None)
```
mikelehen, line 143: Hrm...  I guess you moved the remove_trailing_zeros() call to each of the individual scraper datasources.  I guess I can theoretically see that being useful (if the historical dataset had trailing zeros that we wanted to replace with data from the as-original one).  But I'm curious if you actually ran into a problem in practice?

Per my other comment, if we call modify_dataset() in each individual source, maybe we can undo this change?

mikelehen (approved):
Awesome, thanks! I think this looks good, but I have a couple potential tweaks. Let me know what you think.

libs/datasets/sources/cdc_testing_dataset.py
smcclure17: Smart, I agree it's best to have each piece be as complete as possible, updated to follow this in 2f849b8c4b9051e8cef29ee55fcb650dfa179a1c

libs/datasets/sources/cdc_testing_dataset.py
smcclure17: yep, good call! That was my thought process (removing any trailing zeros so that they would be replaced by newer data) but I didn't notice that actually happening. With the above changes this was reverted

libs/datasets/sources/cdc_testing_dataset.py
smcclure17: yeah I agree, done 🙏

mikelehen (approved):
Changes look good. Thank you!

PR #1228, opened 2022-01-26
migrate covidactnow/datapublic code from covid-data-public 

smcclure17:
most of the changes here are updating files to pull from the current repo instead of covid-data-public. 

This does not bring over [`census_data_helpers.py`](https://github.com/covid-projections/covid-data-public/blob/main/covidactnow/datapublic/census_data_helpers.py) because it is never used anywhere. 

dataset update [here](https://github.com/covid-projections/covid-data-model/actions/runs/1752109292) and API build [here](https://github.com/covid-projections/covid-data-model/actions/runs/1752425603) seem to test this successfully. The resulting snapshot is [2848](https://actnowcoalition.slack.com/archives/C0182MKHWCR/p1643227969002500)


requirements.txt
```diff
@@ -44,10 +44,6 @@ more_itertools==8.3.0
 typing-extensions==3.7.4.3
 xlrd==1.2.0  # For panda's read_excel method
 compress_pickle==2.0.1
--e ../covid-data-public
```
smcclure17, line 47: remove covid-data-public as a requirement

api/can_api_v2_definition.py
```diff
@@ -7,7 +7,7 @@
 from libs.datasets import timeseries
 import pydantic
 import datetime
-from covidactnow.datapublic.common_fields import GetByValueMixin
+from datapublic.common_fields import GetByValueMixin
```
mikelehen, line 10: As we evolve this codebase we should probably try to unify the code.  Having a "datapublic" package really doesn't make any sense anymore.  Not for this PR though!

mikelehen (approved):
Nice, looks good!

api/can_api_v2_definition.py
smcclure17: Yeah I agree it does feel a bit disjointed

PR #1229, opened 2022-01-26
update snapshot generation to remove covid-data-public

smcclure17:
Remove covid-data-public references from the update datasets and deploy api Github actions, and from the shell scripts powering these actions. 

Snapshot [2868](https://github.com/covid-projections/covid-data-model/actions/runs/1778967136) was built off of this branch. This failed b/c of a missing quotation mark that has since been added in 92c0f371fd61fef5136678ef6b8babaffdeaea29, but otherwise appears to be succeeding.


run.sh
```diff
@@ -69,14 +57,6 @@ prepare () {
   SOURCE_DATA_DIR="./data"
 }
 
-execute_raw_data_qa() {
```
smcclure17, line 72: `execute_raw_data_qa` appears to be depreciated, and it was relying on `DATA_SOURCES_DIR` (the covid-data-public directory) so I've opted to just remove it from this script. I didn't remove the code itself in `check_raw_case_death_data.py`

mikelehen: 👍  I think we'd be fine to remove the source code too if we ever feel like it.  It was an early attempt at better QA of our data but didn't really pan out.

mikelehen (approved):
Looks great, thanks!

run.sh
smcclure17: cool, I've gone ahead and removed it in 8c3fa830bd62d9a0b78b52fce8e2c42d29f5d963 (along with updating .github/workflows/CI_test_deploy_api.yml which I missed previously)

PR #1237, opened 2022-01-28
Block CDC datasource for NC counties

smcclure17:
Block CDC data so that state-dashboard data can flow through for North Carolina counties.

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -159,10 +159,6 @@ JOPLIN_COUNTIES = [
 NE_COUNTIES = RegionMask(AggregationLevel.COUNTY, states=["NE"])
 
 
-# 1/27/2020: North Carolina CDC data is inaccurate due to issues tracking
-# booster doses. So we use the state provided data instead.
-NC_COUNTIES = RegionMask(AggregationLevel.COUNTY, states=["NC"])
-
 # NY Times has cases and deaths for all boroughs aggregated into 36061 / New York County.
 # Remove all the NYC data so that USAFacts (which reports each borough separately) is used.
 # Remove counties in MO that overlap with Kansas City and Joplin because we don't handle the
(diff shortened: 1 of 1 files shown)


libs/datasets/combined_datasets.py
```diff
@@ -158,6 +158,11 @@ def datasource_regions(
 # so this region mask might not always be in use.
 NE_COUNTIES = RegionMask(AggregationLevel.COUNTY, states=["NE"])
 
+
+# 1/27/2020: North Carolina CDC data is inaccurate due to issues tracking
```
mikelehen, line 162: 2020? 😛

libs/datasets/combined_datasets.py
```diff
@@ -216,15 +221,16 @@ def datasource_regions(
     ]
 )
 CDCNewVaccinesCompletedBoosterCountiesWithoutExceptions = datasource_regions(
-    CDCNewVaccinesCountiesDataset, exclude=[CDC_STATE_EXCLUSIONS, *CDC_COUNTY_EXCLUSIONS]
+    CDCNewVaccinesCountiesDataset,
+    exclude=[CDC_STATE_EXCLUSIONS, *CDC_COUNTY_EXCLUSIONS, NC_COUNTIES],
```
mikelehen, line 225: Is there a reason not to just add NC_COUNTIES directly to CDC_COUNTY_EXCLUSIONS?

mikelehen (approved):
Looks good, but with a typo and a question.

libs/datasets/combined_datasets.py
smcclure17: oh no... 🤦 😭

libs/datasets/combined_datasets.py
smcclure17: right, that would make much more sense

done in 49aa642bbf27ef04df1df3e75f4903193cea8f70

PR #1240, opened 2022-02-03
Update CDC Vaccine Exclusions

smcclure17:
Updating CDC Vaccine blocks as the result of the review here: https://www.dropbox.com/scl/fi/rtbyfq9pfvnj2w1fki8fg/02_02_2022_-CDC-Vaccine-data-review.paper?dl=0&rlkey=pq42hko1hxcx6l9ga86bb6f2t

### DIFF
--- a/libs/datasets/combined_datasets.py
+++ b/libs/datasets/combined_datasets.py
@@ -196,6 +196,8 @@ CDC_COUNTY_EXCLUSIONS = [
     Region.from_fips("02060"),
     # Culebra Municipio, PR - reports 99.9% of 12+ population vaccinated [7/2021]
     Region.from_fips("72049"),
+    # Chattahoochee County, GA - reports well over 100% 1+ dose vaccinated [2/2022]
+    Region.from_fips("13053"),
 ]
 
 # Excluded for a variety of reasons (lower overall coverage, data irregularities, etc.)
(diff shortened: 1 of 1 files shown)


libs/datasets/combined_datasets.py
```diff
@@ -196,29 +196,24 @@ def datasource_regions(
...
         "PA",
-        # SD - Missing a lot of counties and 1st dose data.
-        "SD",
         # VA - Very low coverage.
         "VA",
         "VT",
-        "WV",
         # CDC reporting 0.1% vaccinated
         "HI",
-        # TODO(sean) 10/23/21: Block TX CDC data until we have a chance to properly QA it
```
smcclure17, line 220: a couple of months late here sorry :/

libs/datasets/combined_datasets.py
```diff
@@ -196,29 +196,24 @@ def datasource_regions(
     Region.from_fips("02060"),
     # Culebra Municipio, PR - reports 99.9% of 12+ population vaccinated [7/2021]
     Region.from_fips("72049"),
+    # Chattahoochee County, GA - reports well over 100% 1+ dose vaccinated [2/2022]
+    Region.from_fips("13053"),
```
smcclure17, line 200: If we'd prefer to not mix sources within a state/region we can remove these county exclusions and potentially look to block these locations instead

mikelehen: I'm a bit on the fence.  Let's go ahead with keeping them for now, I guess.  I get more nervous when there are a lot of regions in a state or they're the biggest counties.  But small one-off counties 🤷

libs/datasets/combined_datasets.py
mikelehen: 🤣

mikelehen (approved):
Sorry for the delay!

PR #1244, opened 2022-02-14
Update Pandas, Numpy, Matplotlib, and Fastparquet versions

smcclure17:


libs/metrics/test_positivity.py
```diff
@@ -119,7 +119,7 @@ def _make_output_dataset(
     # Drop all-NA timeseries now, as done in from_timeseries_wide_dates_df. This makes sure
     # `locations` is used to build provenance information for only timeseries in the returned
     # MultiRegionDataset.
-    wide_date_df = wide_date_df.dropna("rows", "all")
```
smcclure17, line 122: use keyword arguments to fix: 

```
FutureWarning: In a future version of pandas all arguments of DataFrame.dropna will be keyword-only
```

I don't actually know what "rows" is here--`axis` in `dropna()` only has the options `index` or `column`. I assume this is the same as `index` because that's what's used to drop rows (and it is the default) so I've replaced it.

libs/datasets/timeseries.py
```diff
@@ -183,7 +183,7 @@ def demographic_distributions_by_field(
                 # variables with different buckets for a given distribution, including none's
                 # will mix buckets.  To properly fix, consider passing latest in a long format
                 # rather than a wide variables format.
-                if value is None:
+                if value is None or np.isnan(value):
```
smcclure17, line 186: `np.nan` values are not (any longer?) equal to `None`, so empty items were not being caught by this statement. I've left the `None` logic in place in case there are actual `None` values, but we could probably just use  `if np.isnan(value):` by itself.

libs/datasets/region_aggregation.py
```diff
@@ -237,7 +237,7 @@ def _apply_scaling_factor(
     )
     # Check that scale_factors has location index and CommonFields in columns.
     assert scale_factors.index.names == [CommonFields.LOCATION_ID]
-    assert scale_factors.columns.difference(CommonFields).empty
+    assert scale_factors.columns.difference(CommonFields.list()).empty
```
smcclure17, line 240: selecting columns with the CommonFields enum class raised:
```
raise TypeError("Input must be Index or array-like")
```

so I've created a class method to create a list of all the values

libs/datasets/timeseries.py
```diff
@@ -216,7 +216,7 @@ def __post_init__(self):
 
     @property
     def date_indexed(self) -> pd.DataFrame:
-        return self.data.set_index(CommonFields.DATE)
+        return self.data.set_index(CommonFields.DATE).asfreq("D")
```
smcclure17, line 219: in some cases (as was the case in `test_calculate_test_positivity`) the index was a date index, but did not have a frequency, causing comparisons against otherwise-equal dataframes/series to fail. 

This explicitly adds the "day" frequency to the date index

tests/libs/datasets/timeseries_test.py
```diff
@@ -92,15 +92,15 @@ def test_multi_region_to_from_timeseries_and_latest_values(tmp_path: pathlib.Pat
         )
     )
     region_97111 = multiregion.get_one_region(Region.from_fips("97111"))
-    assert region_97111.date_indexed.at["2020-04-02", "m1"] == 2
+    assert region_97111.date_indexed.at[pd.to_datetime("2020-04-02"), "m1"] == 2
```
smcclure17, line 95: Not entirely sure why, but accessing dates with strings no longer worked. This was happening before the change above to add `.freq("D)` to `date_indexed`, so I don't believe it was caused by that

mikelehen (approved):

mikelehen:
Don't understand these issues at a deep level, but superficially, they look good / make sense.  Thanks for the explanatory comments!

PR #1248, opened 2022-02-15
Add optional run number to download execute-model results from

smcclure17:
Allow API snapshots to be built on an existing `execute-model` run from a previous snapshot. 

Snapshot built on a previous model [here](https://github.com/covid-projections/covid-data-model/runs/5218915753?check_suite_focus=true).

Snapshot building off this branch without an existing model [here](https://github.com/covid-projections/covid-data-model/actions/runs/1854295934)

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -78,8 +78,8 @@ execute_model() {
   # Move state output to the expected location.
   mkdir -p ${API_OUTPUT_DIR}/
 
-  Capture all the PDFs pyseir creates in output/pyseir since they are
-  extremely helpful for debugging / QA'ing the model results.
+  # Capture all the PDFs pyseir creates in output/pyseir since they are
+  # extremely helpful for debugging / QA'ing the model results.
   echo ">>> Generating pyseir.zip from PDFs in output/pyseir."
   pushd output
   zip -r "${API_OUTPUT_DIR}/pyseir.zip" pyseir/* -i '*.pdf'
--- a/run.sh
+++ b/run.sh
@@ -15,7 +15,7 @@ prepare () {
     echo "Example: $0 ./api-results/"
     echo "Example: $0 ./api-results/ execute_model"
     echo "Example: $0 ./api-results/ execute_api"
-    echo "Example: $0 ./api-results/ execute_api 2920"
+    echo "Example: $0 ./api-results/ execute_model 2920"
     exit 1
   else
     API_OUTPUT_DIR="$(abs_path $1)"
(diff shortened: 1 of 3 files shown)


run.sh
```diff
@@ -9,12 +9,13 @@ set -o errexit
...
   # Parse args if specified.
-  if [ $# -lt 1 ] || [ $# -gt 2 ]; then
-    echo "Usage: $0 [output-directory] (optional - specific function)"
+  if [ $# -lt 1 ] || [ $# -gt 3 ]; then
+    echo "Usage: $0 [output-directory] (optional - specific function) (optional - pyseir model snapshot number)"
     echo
     echo "Example: $0 ./api-results/"
     echo "Example: $0 ./api-results/ execute_model"
     echo "Example: $0 ./api-results/ execute_api"
+    echo "Example: $0 ./api-results/ execute_api 2920"
```
mikelehen, line 18: nit: I think this should be execute_model

run.sh
```diff
@@ -61,19 +68,29 @@ execute_model() {
...
-
-  # Capture all the PDFs pyseir creates in output/pyseir since they are
-  # extremely helpful for debugging / QA'ing the model results.
-  echo ">>> Generating pyseir.zip from PDFs in output/pyseir."
-  pushd output
-  zip -r "${API_OUTPUT_DIR}/pyseir.zip" pyseir/* -i '*.pdf'
-  popd
+  if [ ! -z "$PYSEIR_ARTIFACT_SNAPSHOT" ]; then
+    echo ">>> Downloading state and county models from existing snapshot ${PYSEIR_ARTIFACT_SNAPSHOT}."
+    ./run.py utils download-model-artifact "${PYSEIR_ARTIFACT_SNAPSHOT}" --output-dir=${API_OUTPUT_DIR%/*}
```
mikelehen, line 73: I guess `${API_OUTPUT_DIR%/*}` is the parent directory of API_OUTPUT_DIR but that wasn't obvious to me, so I'd probably create a variable for that:

```
API_OUTPUT_PARENT="${API_OUTPUT_DIR%/*}"
```

or similar.  That said, I wonder why we are putting it there...  Would it make sense to just use `/tmp/` or something since we're going to move the files afterwards anyway?

mikelehen: That said, if you don't feel like changing it and needing to retest it, feel free to just keep this all as-is.  Maybe add a `TODO(sean): Might be simpler to just download the model results to /tmp/` comment to revisit next time we're messing with this code.

mikelehen (approved):
Nice, looks good!

PR #1249, opened 2022-02-16
unblock all of idaho

clatko:
No current issue is being addressed.

Comparing data from NYT and USAFacts, the Idaho data stagnates around January 07, but is in sync again by January 10 (sampling of Twin Falls).

Revising the disclaimer to the IDHW verbage and unblocking the data.

### Please Check
- [x] Are the [JSON schemas](https://github.com/covid-projections/covid-data-model/tree/main/api/schemas) updated?
- [x] Are tests passing?

### DIFF
--- a/data/region-overrides.json
+++ b/data/region-overrides.json
@@ -537,11 +537,11 @@
     {
       "include": "region-and-subregions",
       "metric": "metrics.caseDensity",
-      "blocked": true,
+      "blocked": false,
       "start_date": "",
       "end_date": "",
       "region": "ID",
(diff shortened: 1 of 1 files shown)


smcclure17 (approved):
Hi Chris, thanks for taking the time to make this change! 

Just for clarification, we’ve had this blocked because we didn’t trust that the data being reported was an accurate representation of the number of actual Covid cases in ID. At this point, with cases decreasing across the country, their delay in handling cases will probably begin to show inflated case numbers at some point in the future. 

That being said, we do now feel comfortable removing this block and just keeping an eye on it going forward, so I'll go ahead and merge this. 

Thanks again! 
Sean

PR #1250, opened 2022-02-16
Add repo dispatch for triggering via HTML request from can-scrapers

smcclure17:
Adds a `repository_dispatch` option to allow the action to be triggered by a `post` request in can-scrapers. This would replace the cron scheduler currently used (although I've kept the schedule active for now). 

I've been following the steps outlined [here](https://dev.to/rikurouvila/how-to-trigger-a-github-action-with-an-htt-request-545), and have a [PR in the works to make the trigger request in can-scrapers](https://github.com/covid-projections/can-scrapers/pull/413).

### DIFF
--- a/.github/workflows/update_repo_datasets.yml
+++ b/.github/workflows/update_repo_datasets.yml
@@ -2,8 +2,8 @@ name: Update combined datasets
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


mikelehen (approved):
👍

PR #1259, opened 2022-03-21
Add HSA locations to geo-data.csv and Region class

smcclure17:
Adds HSA `location_ids` to `data/geo-data.csv` and helper functions to the `Region` class. 

- HSA location ids are formatted as: `iso1:us#hsa:<HSA NUMBER>`, with padded zeros where necessary (e.g `iso1:us#hsa:001`. 
- HSA codes range from 1-905
- The HSA definitions used by the CDC are not contained within states (meaning counties from different states can belong to the same HSA)


smcclure17:
I'm actually not convinced this will be necessary, so I don't think we need to look to merge this until it is needed

mikelehen (approved):
This looks good to me once we need it.

FWIW I'm a bit surprised that cdc_hsa_mapping.csv isn't (apparently?) getting put in git LFS...  I think any CSV's under `data/` should automatically go in LFS. 
```
╰─$ git lfs track
Listing tracked patterns
    data/*.csv (.gitattributes)
    data/*.gz (.gitattributes)
    data/*.parquet (.gitattributes)
    tests/data/test-combined-*.csv (.gitattributes)
```

But maybe it *is* in LFS and I'm getting confused by the github UI.  In any case, it's only 3k and will never change, so I don't really care.

smcclure17:
Huh, yeah good point. Maybe because it's under the `misc/` subdirectory? Checking `

PR #1260, opened 2022-03-22
Add HSAs and HSA populations to API

smcclure17:
Snapshot `3116` was generated with this PR, and can be previewed here: 
https://data.covidactnow.org/snapshot/3116/v2/county/36001.timeseries.json
https://data.covidactnow.org/snapshot/3116/v2/state/MA.timeseries.json
https://data.covidactnow.org/snapshot/3116/v2/cbsa/11100.timeseries.json

https://data.covidactnow.org/snapshot/3116/v2/counties.timeseries.csv
https://data.covidactnow.org/snapshot/3116/v2/states.csv
https://data.covidactnow.org/snapshot/3116/v2/cbsas.csv


libs/datasets/dataset_utils.py
```diff
@@ -18,7 +18,7 @@
     CommonFields.COUNTY,
 ]
 
-NON_NUMERIC_COLUMNS = GEO_DATA_COLUMNS + [CommonFields.CAN_LOCATION_PAGE_URL]
+NON_NUMERIC_COLUMNS = GEO_DATA_COLUMNS + [CommonFields.CAN_LOCATION_PAGE_URL, CommonFields.HSA]
```
smcclure17, line 21: So I think the HSA column was being converted to numerics in the [`_merge_attributes()` method](https://github.com/covid-projections/covid-data-model/blob/main/libs/datasets/timeseries.py#L391-L392) when being loaded into a `MultiRegionDataset`, so specifying this column as non-numeric should hopefully solve the issue

smcclure17: Nvm, that didn't work :/ ([see this snapshot](https://data.covidactnow.org/snapshot/3107/v2/county/36061.timeseries.json))

smcclure17: 977caf12cc058f3f994cc007bdca73a3102c7c8d fixes this in a kind of hacky way

libs/datasets/sources/hsa_population.py
```diff
@@ -0,0 +1,65 @@
...
+from libs.pipeline import Region
+from libs.datasets.dataset_utils import AggregationLevel
+from libs.datasets.statistical_areas import CountyToHSAAggregator
+from libs.datasets.sources.fips_population import FIPSPopulation
+
+
+def get_location_level(location_id):
+    location = Region.from_location_id(location_id)
+    # TODO(sean): Weed out locations whose levels cant be determined.
+    # In this case it's fine because we only want to select counties.
```
mikelehen, line 15: Just for my own understanding, do we have any idea what locations are triggering this?  Or that's what this comment is saying we need to figure out?  I'm fine with leaving this as a TODO. Just checking.

mikelehen (approved):
This looks great!

libs/datasets/sources/hsa_population.py
smcclure17: I took a look and the two FIPS throwing this are `iso2:us#iso2:us-dc#fips:11999` and `iso2:us#iso2:us-fl#fips:12999`, neither of which are valid locations so I'm not sure what they represent. 

Somewhat unrelated (it wasn't breaking anything here), I realized that I didn't add implementation for  `Region.level` for HSA locations, so I've added that in a01d4a78795e22c05cc217cee9f57da8587a2b7b, though this has no effect on the rest of this PR.

PR #1263, opened 2022-03-23
Add HSA aggregations

smcclure17:
Add fields to track HSA level data for county hospital fields. 

I've tested by running `./run.py data update --state CT`, verifying that the expected columns exist, and validating that the number of staffed beds for an HSA (004) match the expected value (the sum of the county data).


libs/datasets/statistical_areas.py
```diff
@@ -103,4 +105,47 @@ def from_local_data() -> "CountyToHSAAggregator":
...
+        ds_in_ts = ds_in_ts.set_index([CommonFields.LOCATION_ID, CommonFields.DATE]).sort_index()
+
+        # Combine aggregated fields with untouched fields.
+        out_ts = ds_in_ts.join(ds_ts_other)
+
+        # Drop all the columns used in the calculations and combine with non-county data.
+        out_ts = out_ts.loc[:, counties.timeseries.columns]
+        out_ts = out_ts.combine_first(dataset_in.timeseries)
+
+        return dataclasses.replace(counties, timeseries=out_ts, timeseries_bucketed=None)
```
smcclure17, line 151: We might want to update the tags here to clarify that this data has been calculated/aggregated

datapublic/common_fields.py
```diff
@@ -152,6 +152,18 @@ def list(cls):
...
     )
 
+    # HSA level Healthcare Fields
+    STAFFED_BEDS_HSA = "staffed_beds_hsa", FieldGroup.HEALTHCARE_CAPACITY
+    HOSPITAL_BEDS_IN_USE_ANY_HSA = "hospital_beds_in_use_any_hsa", FieldGroup.HEALTHCARE_CAPACITY
+    CURRENT_HOSPITALIZED_HSA = "current_hospitalized_hsa", FieldGroup.HEALTHCARE_CAPACITY
+    ICU_BEDS_HSA = "icu_beds_hsa", FieldGroup.HEALTHCARE_CAPACITY
+    CURRENT_ICU_HSA = "current_icu_hsa", FieldGroup.HEALTHCARE_CAPACITY
+    CURRENT_ICU_TOTAL_HSA = "current_icu_total_hsa", FieldGroup.HEALTHCARE_CAPACITY
+    WEEKLY_NEW_HOSPITAL_ADMISSIONS_COVID_HSA = (
```
smcclure17, line 162: I don't think we need to add `NEW_HOSPITAL_ADMISSIONS_COVID` here because we won't have any county-level data for that field, but it also wouldn't hurt to add it.

mikelehen (approved):
Nice.  This looks good to me!

libs/datasets/statistical_areas.py
```diff
@@ -103,4 +124,54 @@ def from_local_data() -> "CountyToHSAAggregator":
...
+            hsa.location_id: [county.location_id for county in counties]
+            for hsa, counties in self.hsa_to_counties_region_map.items()
+        }
+        hsa_ts[CommonFields.LOCATION_ID] = hsa_ts.index.get_level_values(
+            CommonFields.LOCATION_ID
+        ).map(location_id_map)
+
+        # Create an row for each county in each HSA using HSA data.
+        # NOTE: Every county in an HSA will have data attributed to it, regardless of whether or
+        # not we have actually collected data for that county.
```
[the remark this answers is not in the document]
smcclure17: Yeah, this was kinda unclear. I've updated this in 486fcb2ab2754f9b98c08f6494d9cab4cb92aebd to hopefully be more clear. What I was trying to convey was that even if a county did not have any county-level data they can still have HSA level data

PR #1266, opened 2022-03-27
Add option to github action to skip refreshing datasets.

mikelehen:
`./run.py data update` accepts a `--no-refresh-datasets` option that tells it to skip fetching / combining datasources and instead re-use the checked-in `combined-raw.pkl.gz` data.  This PR exposes that option to the github action.  This can be used to save time when you just need to block data and re-build the snapshot.

I also tweaked the github actions to work with branches that have `/` in the name.

### DIFF
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -15,7 +15,7 @@ on:
         description: 'Sentry environment of build. Should be "production" on main build and "staging" for other branches.'
         required: true
       pyseir_snapshot:
-        description: 'Optionally download an existing pyseir model artifact from a previous snapshot number instead of generating a new one.'     
+        description: 'Optionally download an existing pyseir model artifact from a previous snapshot number instead of generating a new one.'
         required: false
         default: ""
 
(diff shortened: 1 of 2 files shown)


smcclure17 (approved):
Ah, both of these are super helpful changes, thanks! 🎉 🎉

.github/workflows/deploy_api.yml
```diff
@@ -48,15 +48,15 @@ env:
...
+  PYSEIR_ARTIFACT_SNAPSHOT: ${{ github.event.inputs.pyseir_snapshot }}
 
 jobs:
   build-and-publish-snapshot:
     runs-on: self-hosted
     steps:
     - name: Parse covid data model branch name and set env variable
       run: |
-        echo "COVID_DATA_MODEL_REF=${GITHUB_REF##*/}" >> $GITHUB_ENV
+        echo "COVID_DATA_MODEL_REF=${GITHUB_REF_NAME}" >> $GITHUB_ENV
```
mikelehen, line 59: GITHUB_REF is something like "refs/heads/<branch_name>" and so we were stripping off everything up to the (last) `/` character to get the branch name, but that made branches like `mikelehen/foo` not work correctly.  But it looks like we can just use `GITHUB_REF_NAME` which will be something like `mikelehen/foo`.  

https://docs.github.com/en/actions/learn-github-actions/environment-variables

mikelehen:
FYI- I merged this but then reverted it since I wasn't able to test it end-to-end before merging and I don't want to risk tomorrow's deploy.  I'll merge again tomorrow. https://github.com/covid-projections/covid-data-model/pull/1269

PR #1272, opened 2022-03-28
Weekly hospital admissions

igorkofman:
This PR addresses https://trello.com/c/mQAfsaWF/399-add-cdc-community-level-related-metrics-to-api

Adds `weeklyCovidAdmissions` actual to hospitalBeds for all endpoints and hsaHospitalBeds for counties only.

Preview snapshot:
https://data.covidactnow.org/snapshot/3142/v2/county/25017.timeseries.json
https://data.covidactnow.org/snapshot/3142/v2/state/MA.timeseries.json
https://data.covidactnow.org/snapshot/3142/v2/cbsa/31080.timeseries.json
https://data.covidactnow.org/snapshot/3142/v2/county/MA.timeseries.csv
https://data.covidactnow.org/snapshot/3142/v2/state/MA.timeseries.csv


libs/datasets/weekly_hospitalizations.py
```diff
@@ -0,0 +1,55 @@
...
+
+    state_ds = dataset_in.get_subset(aggregation_level=level_to_replace)
+    state_ts = state_ds.timeseries_bucketed.copy()  # copy to avoid SettingWithCopy warning.
+
+    # If there's no data for field_in then just return the original dataset.
+    if not state_ts[field_in].first_valid_index():
+        return dataset_in
+
+    # Make sure we do not have any pre-existing weekly data before overwriting the field_out column.
+    assert not state_ts[field_out].any()
```
smcclure17, line 39: We could instead make this non-breaking and just use existing data if it exists instead of calculating it, but we shouldn't ever (at the moment) have any raw weekly data at the state level, so if this gets thrown then that probably means something is wrong.

libs/datasets/weekly_hospitalizations.py
```diff
@@ -0,0 +1,55 @@
...
+        level_to_replace: Aggregation level to update data for.
+        field_in: Column to sum into weekly data, should have units of new daily occurrences (e.g. new daily hospital admissions).
+        field_out: Column to sum data into. Result will have units of new weekly occurrences (e.g. new weekly hospital admissions).  
+    """
+
+    state_ds = dataset_in.get_subset(aggregation_level=level_to_replace)
+    state_ts = state_ds.timeseries_bucketed.copy()  # copy to avoid SettingWithCopy warning.
+
+    # If there's no data for field_in then just return the original dataset.
+    if not state_ts[field_in].first_valid_index():
```
mikelehen, line 35: nit: you use `.any()` below.  Could we use that here instead of `first_valid_index()` ?

libs/datasets/weekly_hospitalizations.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Args:
+        dataset_in: The dataset to update.
+        level_to_replace: Aggregation level to update data for.
+        field_in: Column to sum into weekly data, should have units of new daily occurrences (e.g. new daily hospital admissions).
+        field_out: Column to sum data into. Result will have units of new weekly occurrences (e.g. new weekly hospital admissions).  
+    """
+
+    state_ds = dataset_in.get_subset(aggregation_level=level_to_replace)
+    state_ts = state_ds.timeseries_bucketed.copy()  # copy to avoid SettingWithCopy warning.
```
mikelehen, line 32: The `state_` prefix doesn't exactly make sense since `level_to_replace` could be a different level.  Perhaps `subset_ts` or `replace_ts` or something ?

mikelehen (approved):
Couple nits, but looks great!

libs/datasets/weekly_hospitalizations.py
smcclure17: ah, yep! `any` does feel more relevant here

PR #1275, opened 2022-03-28
Add community levels to API

mikelehen:
Wires up `communityLevels.cdcCommunityLevel` and `communityLevels.canCommunityLevel`.

All of the changes (except my last commit to ffill() the metrics before calculating canCommunityLevel) are in [this snapshot](https://data.covidactnow.org/snapshot/3162/v2/) which is visible at https://covid-projections-git-mikelehen-wire-up-comm-a4e145-covidactnow.vercel.app/?s=31461972.

The ffill() change was validated locally with `./run.py api generate-api-v2 --state ID output output/api`. An example county with the change in effect can be found [here](https://gist.githubusercontent.com/mikelehen/03f0586a66b59bb6b1067820e109504c/raw/7859ca035848d9f5d26d64a5953486813a0055fc/16083.timeseries.json) where you'll see `communityLevelTimeseries.canCommunityLevel` has no gaps since we're forward-filling the nans in the metrics.


libs/metrics/community_levels.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if (
+        is_invalid_value(weekly_cases_per_100k)
+        or is_invalid_value(beds_with_covid_ratio)
+        or is_invalid_value(weekly_admissions_per_100k)
+    ):
+        # TODO(michael): For now, return None if we are missing any of the
+        # underlying metrics. Once we see how bad this is we can compare against
+        # CDC and decide how to handle.
+        return None
```
mikelehen, line 33: I'll evaluate this based on the progress being built before merging this.

mikelehen:
[Add CDC Community Level-related metrics to API.](https://trello.com/c/mQAfsaWF/399-add-cdc-community-level-related-metrics-to-api)

api/can_api_v2_definition.py
```diff
@@ -541,6 +552,69 @@ class CdcTransmissionLevelTimeseriesRow(base_model.APIBaseModel):
...
+        - 0: Low
+        - 1: Medium
+        - 2: High
+
+    See https://www.cdc.gov/coronavirus/2019-ncov/science/community-levels.html
+    for details about how the Community Level is calculated and should be
+    interpretted.
+
+    Note that we provide two versions of the Community Level. One is called
+    canCommunityLevel which is calculated on a daily basis using CAN's data
```
smcclure17, line 599: Just to note for counties, this is calculated daily but I think it will only have a non `null` value once a week in the timeseries, (on the day that the hospital data is reported see screenshot). This makes sense/is expected I think, but it might be good to clarify that there won't be a `canCommunityLevel` value for every day in the API. 

image

libs/metrics/community_levels.py
```diff
@@ -0,0 +1,99 @@
...
+
+from api import can_api_v2_definition
+
+CommunityLevel = can_api_v2_definition.CommunityLevel
+CommunityLevels = can_api_v2_definition.CommunityLevels
+
+
+def is_invalid_value(value: Optional[float]) -> bool:
+    # TODO(michael): Do we really need all these checks?
+    return value is None or math.isinf(value) or np.isnan(value)
```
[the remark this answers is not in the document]
smcclure17: Might be good to investigate, but I'm 100% okay with just leaving it as is (rather be safe than sorry 😅)

api/can_api_v2_definition.py
```diff
@@ -518,7 +529,7 @@ def empty(cls) -> "RiskLevels":
 # Additional class used for bulk timeseries where we are not including all risk levels
 # right now, only the overall risk level.
 class RiskLevelsRow(base_model.APIBaseModel):
-    overall: RiskLevel = pydantic.Field(..., description="Overall risk level for region.")
+    overall: RiskLevel = pydantic.Field(..., description="Overall risk level for region .")
```
smcclure17, line 532: super nit: is this extra space intentional?

smcclure17 (approved):
Sweet! This looks great from what I can see 🎉 

I'll be interested to see how many locations are missing data/unable to calculate the canCommunityLevel (hopefully not many)

smcclure17:
Let me know if I'm missing something, but I don't seem to see any data for these columns under the csv timeseries: https://data.covidactnow.org/snapshot/3150/v2/states.timeseries.csv

mikelehen:
Hrm.  You don't seem to be missing something.  Good catch!  I'll take a look.  It also looks like the full API build failed to generate some endpoints properly, e.g. county data for DC and MP.

smcclure17:
Oh huh, I wonder if that has to do with HSAs--I don't think MP counties are mapped to HSA (I don't know about DC) so I wonder if that might be causing a problem

PR #1276, opened 2022-03-28
Calculate and wire up bedsWithCovidPatientsRatio into API

smcclure17:
This PR addresses: https://trello.com/c/mQAfsaWF/399-add-cdc-community-level-related-metrics-to-api

Snapshot `3136` was built with this branch and can be previewed:
https://data.covidactnow.org/snapshot/3136/v2/county/36001.timeseries.json
https://data.covidactnow.org/snapshot/3136/v2/state/NY.timeseries.json
https://data.covidactnow.org/snapshot/3136/v2/county/NY.timeseries.csv
https://data.covidactnow.org/snapshot/3136/v2/state/NY.timeseries.csv

`bedsWithCovidPatientsRatio` is the corresponding metric to the CDC's: 

image


api/can_api_v2_definition.py
```diff
@@ -392,6 +395,11 @@ class Metrics(base_model.APIBaseModel):
         description="Ratio of staffed intensive care unit (ICU) beds that are currently in use.",
     )
 
+    bedsWithCovidPatientsRatio: Optional[float] = pydantic.Field(
+        ...,
+        description="Ratio of staffed hospital beds that are currently in use by COVID patients.",
```
mikelehen, line 400: Can we add a note that this is calculated at the HSA level for counties?

libs/build_api_v2.py
```diff
@@ -189,6 +189,7 @@ def build_region_summary(
     "vaccinationsCompletedRatio": CommonFields.VACCINATIONS_COMPLETED_PCT,
     "vaccinationsAdditionalDoseRatio": CommonFields.VACCINATIONS_ADDITIONAL_DOSE_PCT,
     "icuCapacityRatio": CommonFields.CURRENT_ICU,
+    "bedsWithCovidPatientsRatio": CommonFields.CURRENT_HOSPITALIZED,
```
mikelehen, line 192: Do you understand exactly what this `METRICS_NAME_TO_COMMON_FIELD` is used for?  It's a bit confusing since there isn't a 1:1 mapping between metrics and common fields.  If you have useful context, would it be possible to add a comment to it?

mikelehen (approved):
Looks good!

libs/build_api_v2.py
smcclure17: I don't really have a good sense, but I think it might be used to link metrics back to a data source in order to build the annotations.

api/can_api_v2_definition.py
smcclure17: Ah, good idea

smcclure17: aec817a283741576042293f6d812f257cf2c1dbd adds the copy "For counties, this is calculated using HSA-level data for the corresponding area."

PR #1277, opened 2022-03-28
Fix Black formatting

smcclure17:
Black failing due to: https://github.com/psf/black/issues/2964

Updating to the most recent version fixes this, but would reformat a lot of (~60) files.


ichard26:
I'll note you don't have to bump Black to 22.3.0, if you're fine with pinning Click to something below 8.1.0 Black won't blow up.

PR #1278, opened 2022-03-29
Add HSA names to API

smcclure17:
Adds names of HSAs to county endpoints, (or "Unknown HSA Name" when an HSA name is not available.) HSA names lined up/were available for all HSAs except those in Puerto Rico.

Tested by generating API files locally with:
```
./run.py data update --state CT
python ./pyseir/cli.py build-all --states=CT
./run.py api generate-api-v2 --state CT output -o output/api
```

County endpoint:
image

State endpoint:
image

I can generate a full snapshot for this, but I feel pretty confident that this should work as expected


api/can_api_v2_definition.py
```diff
@@ -532,6 +532,10 @@ class RegionSummary(base_model.APIBaseModel):
         ...,
         description="3 digit Health Service Area identification code. For CBSA, state, and country regions hsa is omitted. For more on HSAs see: https://github.com/covid-projections/covid-data-model/blob/main/data/misc/README.md",
     )
+    hsaName: Optional[str] = pydantic.Field(
+        ...,
+        description="Name of Health Service Area. For CBSA, state, and country regions hsa is omitted. For more on HSAs see: https://github.com/covid-projections/covid-data-model/blob/main/data/misc/README.md",
```
mikelehen, line 537: "hsa is omitted" => "hsaName is omitted" ?

api/can_api_v2_definition.py
```diff
@@ -642,6 +646,11 @@ class RegionTimeseriesRowWithHeader(base_model.APIBaseModel):
         description="3 digit Health Service Area identification code. For CBSA, state, and country regions hsa is omitted. For more on HSAs see: https://github.com/covid-projections/covid-data-model/blob/main/data/misc/README.md",
     )
 
+    hsaName: Optional[str] = pydantic.Field(
+        ...,
+        description="Name of Health Service Area. For CBSA, state, and country regions hsa is omitted. For more on HSAs see: https://github.com/covid-projections/covid-data-model/blob/main/data/misc/README.md",
```
mikelehen, line 651: ditto

data/misc/mapping_temp.csv
```diff
@@ -0,0 +1,3221 @@
+FIPS,HSA,hsa_name
+1001,171,"Montgomery (Montgomery), AL - Autauga, AL"
```
mikelehen, line 2: Is this `mapping_temp.csv` file meant to be checked in?

libs/datasets/sources/hsa_population.py
```diff
@@ -40,6 +42,12 @@ def make_dataset(cls) -> timeseries.MultiRegionDataset:
         populations = FIPSPopulation.make_dataset().static
         county_index = populations.index.map(get_location_level) == AggregationLevel.COUNTY
         counties = populations[county_index].copy()  # copy to remove SettingWithCopy error
+        hsa_names: pd.DataFrame = (
+            pd.read_csv(HSA_LIST_PATH, dtype={CommonFields.HSA: str})
+            .loc[:, [CommonFields.HSA, CommonFields.HSA_NAME]]
+            .drop_duplicates()
+            .assign(hsa_name=lambda row: row[CommonFields.HSA_NAME].fillna("Unknown HSA Name"))
```
mikelehen, line 49: Can you make this just "Unknown"?

I'm imagining in the frontend we're going to have copy something like:
> Over the past 7 days, the {hsaName} Health Service Area had xyz staffed beds and xyz covid patients, resulting in xy% of staffed beds filled with covid patients.

mikelehen (approved):
Nice!  Looks good.

data/misc/mapping_temp.csv
smcclure17: nope it is not :( thanks

api/can_api_v2_definition.py
smcclure17: ah, thanks for the catch--updated in 9aeffdab7a5a35dcab0e502b1f7f5cd9ccc57eb2

PR #1279, opened 2022-03-31
Calculate and add weeklyCovidAdmissionsPer100k to API

smcclure17:
This PR addresses https://trello.com/c/mQAfsaWF/399-add-cdc-community-level-related-metrics-to-api

Tested by generating API files locally with:
```
./run.py data update --state CT
python ./pyseir/cli.py build-all --states=CT
./run.py api generate-api-v2 --state CT output -o output/api
```
county timeseries:
image
image


tests/libs/datasets/timeseries_test.py
```diff
@@ -2103,7 +2103,7 @@ def test_pickle_test_dataset_size(tmp_path: pathlib.Path):
     test_dataset = test_helpers.load_test_dataset()
     test_dataset.get_timeseries_not_bucketed_wide_dates(CommonFields.CASES)
     test_dataset.to_compressed_pickle(pkl_path)
-    assert pkl_path.stat().st_size < 800_000
+    assert pkl_path.stat().st_size < 900_000
```
smcclure17, line 2106: Not sure what the significance of this is. 🤷 My assumption is that it's to make sure the pickle doesn't get misconfigured and become 2x/3x the expected size. With this PR the size increases to `808369`

smcclure17:
Ah, just noticed two tests started failing again. Let me fix these before any reviews

libs/metrics/top_level_metrics.py
```diff
@@ -46,6 +46,7 @@ class MetricsFields(common_fields.ValueAsStrMixin, str, enum.Enum):
     INFECTION_RATE_CI90 = "infectionRateCI90"
     ICU_CAPACITY_RATIO = "icuCapacityRatio"
     BEDS_WITH_COVID_PATIENTS_RATIO = "bedsWithCovidPatientsRatio"
+    WEEKLY_COVID_ADMISSIONS_DENSITY_RATIO = "weeklyCovidAdmissionsPer100k"
```
mikelehen, line 49: nit: For clarity I'd be a fan of sticking to "per 100k" terminology instead of "density ratio".  So I would just do `WEEKLY_COVID_ADMISSIONS_PER_100K`

I know Igor did WEEKLY_CASE_DENSITY_RATIO instead of WEEKLY_CASES_PER_100K but since that's directly mirroring the existing CASE_DENSITY_RATIO I think it's okay.

libs/metrics/top_level_metrics.py
```diff
@@ -106,6 +108,12 @@ def calculate_metrics_for_timeseries(
 
     icu_capacity_ratio = icu_capacity.calculate_icu_capacity(data)
     beds_with_covid_patients_ratio = calculate_covid_patient_ratio(data, timeseries.region)
+    weekly_admissions_density = calculate_weekly_admissions_density(
```
mikelehen, line 111: ditto.  weekly_admissions_per_100k

mikelehen (approved):

libs/metrics/top_level_metrics.py
smcclure17: Yeah, that sounds good to me! I'll update

PR #1281, opened 2022-04-01
Fix DC HSA population.

mikelehen:


smcclure17 (approved):
Cool, looks good! thanks

libs/datasets/custom_aggregations.py
```diff
@@ -72,6 +72,11 @@ def replace_dc_county_with_state_data(
         static_excluding_numbers.reset_index()
     )
 
+    # The aggregation will have replaced hsa_population with nan.  We need to fix it.
+    dc_county_dataset.static.loc[
+        :, CommonFields.HSA_POPULATION
+    ] = dataset_with_dc_county.static.loc[:, CommonFields.HSA_POPULATION]
```
mikelehen, line 78: This ended up more straightforward than changing `static_excluding_numbers` since `add_static_values()` threw an assert that hsa_population was already in the dataset. So I just replace it after the fact.

I think it might be kind of hacky to mutate `dc_county_dataset.static` directly (instead of using `dataclasses.replace()` or whatever), but since it's only used inside this function, I think it's fine?

smcclure17: Yeah, I think in this instance as long as it updates the dataframe I'm not really concerned with it at all 🙏

PR #1282, opened 2022-04-02
Rework black formatting check in github action.

mikelehen:
Also stops checking out `covid-data-public`.


smcclure17 (approved):
Thanks for jumping on this! looks great

PR #1283, opened 2022-04-04
Update docs regarding weekly / daily.

mikelehen:
Tweaks docs to reflect that canCommunityLevel updates daily.  Though these docs may need more tweaking if we switch data sources to the county timeseries data used by CDC Covid Data Tracker or something.


smcclure17 (approved):
lgtm!

PR #1285, opened 2022-04-06
Use CDC hospital metrics for counties / metros.

mikelehen:
Snapshot built with these changes: https://data.covidactnow.org/snapshot/3180/v2/
Preview site with map wired up to that snapshot (but none of the other can-82 changes): https://covid-projections-git-mikelehen-wire-up-comm-a4e145-covidactnow.vercel.app/


libs/datasets/timeseries.py
```diff
@@ -741,6 +741,7 @@ def print_stats(self, name: str):
                 textwrap.indent(by_level.to_string(), "  "),
             ]
         )
+        print(datetime.datetime.now())
```
mikelehen, line 744: I'm slowly trying to understand why our actions are so slow and adding more logging to help understand where the time goes...

libs/metrics/top_level_metrics.py
```diff
@@ -292,6 +292,13 @@ def calculate_contact_tracers(
 
 
 def calculate_covid_patient_ratio(data: pd.DataFrame, region: Region):
+    # If we have precomputed data, use it.
+    if (
+        CommonFields.BEDS_WITH_COVID_PATIENTS_RATIO_HSA in data.columns
```
mikelehen, line 297: The `CommonFields.BEDS_WITH_COVID_PATIENTS_RATIO_HSA in data.columns` is here so I don't have to update tests and the `test-combined-wide-dates.csv` file... which is maybe hacky?  Not sure.

smcclure17: This is fine with me--it also means we might create community levels for locations that the CDC doesn't cover?

smcclure17 (approved):
looks good to me! thanks for pushing these through

libs/metrics/top_level_metrics.py
mikelehen: Yeah.  I think in that case the column will exist (it basically only doesn't exist in tests) but won't have any data, and we will indeed calculate the community level ourselves, which I think is fine / good?

PR #1286, opened 2022-04-06
Restore demographic vaccine data

smcclure17:
Demographic data was getting clobbered/removed from the `MultiregionDatasets` due to my use of `timeseries` instead of `timeseries_bucketed`. This updates the code where necessary to restore demographic data

post-fix: https://gist.githubusercontent.com/smcclure17/a50c3c6395bb37552771104b11279c8f/raw/a1ad4a60a4fd6af40d1807f26b7ec0d2d23f3a12/ca-06037.json

pre-fix: https://gist.githubusercontent.com/smcclure17/2365b842dbbb510d620453298957ce52/raw/4edb7fe774f88e8f80bfd90a8ba9a9407efb9418/ca-06037.json

### DIFF
--- a/libs/datasets/custom_patches.py
+++ b/libs/datasets/custom_patches.py
@@ -35,7 +35,7 @@ def patch_maryland_missing_case_data(
     # Need to convert timestamp to strings to match the index dtype.
     dates_to_replace = [str(date.date()) for date in dates_to_replace]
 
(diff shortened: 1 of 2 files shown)


mikelehen (approved):
Looks good.  I wonder if there's an easy way to add a test that would have caught this, but we can get this merged first.

PR #1288, opened 2022-04-11
Add `load_demographics` option to `MultiregionDataset.from_wide_dates_csv` and related methods

smcclure17:
Allow multiregion CSVs to be chunk/stream loaded, only keeping non-demographic data in order to reduce memory usage and make tests pass again.  

see https://www.dropbox.com/scl/fi/xpj6dpx7eyrpcc0t9c6it/2022-04-11_-covid-data-model-broken-unit-tests.paper?dl=0&rlkey=uap5z38sv4cz37uzopiykg8b0


libs/datasets/timeseries.py
```diff
@@ -960,8 +960,19 @@ def from_csv(path_or_buf: Union[pathlib.Path, TextIO]) -> "MultiRegionDataset":
         return dataset
 
     @staticmethod
```
mikelehen, line 962: This is tangential to this PR but I'm noticing that we have `@functools.lru_cache(maxsize=None)` on several functions that call this function but we don't have it here, even though this is the function that dose the real work and so this is the innermost place we could apply the caching and have it provide the most benefit.  As-is, each function that calls this will probably load the full dataset once before it's cached for subsequent uses.

So I'm wondering if we could / should add `@functools.lru_cache(maxsize=None)` here and remove it from all of the callers.

Feel free to punt / ignore since it's not directly related to this PR but it may speed up CI / tests a bit.

mikelehen (approved):
Nice!  This is a pretty slick solution.  Looks good to me with one drive-by tangential comment.

libs/datasets/timeseries.py
smcclure17: Ooh, that's a good flag. I think I'll ship this as-is, but will create a follow-up PR to try that out 🙏

PR #1290, opened 2022-04-12
Allow one hospital metric to be missing and still calculate community level.

mikelehen:


libs/metrics/community_levels.py
```diff
@@ -22,16 +22,24 @@ def calculate_community_level(
...
+    # TODO(michael): The CDC footnotes say:
+    #     If the number of cases in 7 days for a jurisdiction is missing, the
+    #     7-day case rate is assigned to the “low” category. If both 7-day
+    #     admissions and 7-day percentage inpatient beds indicators are N/A, the
+    #     community burden category is assigned N/A.
+    #
+    # For now I'm allowing 1 hospital metric to be missing, but not allowing
+    # cases to be missing since that is rare and usually indicates we're
+    # blocking data or something. In that case, I'd rather have no community
+    # level calculated. But we can revisit if it ends up being a problem.
```
mikelehen, line 34: If you think we should implement the CDC logic directly (treat missing cases as low), I could be convinced. 🤷

smcclure17 (approved):
looks great! thanks as always

PR #1292, opened 2022-04-13
Add disclaimer for potential hospitalization data undercounts

smcclure17:


api/docs/docs/data-definitions.md
```diff
@@ -160,6 +160,7 @@ Fields:
 * CSV column names: ``metrics.weeklyCovidAdmissionsPer100k``
 * JSON file fields: ``metrics.weeklyCovidAdmissionsPer100k``, ``metricsTimeseries.*.weeklyCovidAdmissionsPer100k``
 
+County- and metro-level hospitalization and ICU actuals are calculated from facility-level data which, suppresses near-zero values due to privacy concerns. As a result, this data may undercount real values, particularly at low levels.
```
smcclure17, line 163: "at low levels" feels vague, but I couldn't think of a better way to phrase this. Open to suggestions, or we can remove it

mikelehen (approved):
Thanks for adding this!  I think it's phrased fine for now.  If we get questions, we can revisit.

PR #1295, opened 2022-04-18
Read parquet file directly from GCS

smcclure17:
Tweak the data pipeline such that the scraper data parquet file does not get persisted to `data/` during each pipeline run


mikelehen (approved):
Looks good, thanks!

PR #1296, opened 2022-04-18
Use gzip to compress `multiregion-wide-dates.csv`

smcclure17:


libs/datasets/timeseries.py
```diff
@@ -1374,9 +1374,8 @@ def write_to_wide_dates_csv(self, path_wide_dates: pathlib.Path, path_static: pa
         # that are now over 100M. Unfortunately this also writes unnecessarily high precision for
         # floats but I don't see an easy solution with to_csv float_format.
         # https://trello.com/c/aDGn57Df/1192-change-combined-data-from-csv-to-parquet will remove
-        # the need to format values as strings.
-        csv_buf = wide_df.to_csv(index=True, float_format="%.9g")
```
smcclure17, line 1378: Pandas can handle gzip compression under the hood without us having to use the `gzip` library, or any sort of file handler, so I've opted for that for simplicity. 

I'm not sure what the specific reason was for writing to a csv object and then using pathlib to write that test to a file, but writing the csv directly using pandas (and gzip compression) seems to work fine.

tests/libs/datasets/timeseries_test.py
```diff
@@ -597,15 +598,18 @@ def test_write_read_wide_dates_csv_compare_literal(tmpdir):
...
+            "                  location_id,variable,provenance,2020-04-03,2020-04-02,2020-04-01\n"
+            "           iso1:us#iso2:us-as,   cases,          ,       300,       200,       100\n"
+            "           iso1:us#iso2:us-as,icu_beds,   pt_src1,         4,         2,         0\n"
+            "iso1:us#iso2:us-ca#fips:06075,   cases,          ,       310,       210,\n"
+            "iso1:us#iso2:us-ca#fips:06075,  deaths,   pt_src2,          ,         2,         1\n".replace(
+                " ", ""
+            )
         )
     )
+    data.to_csv(pointer.path_wide_dates(), index=False, compression="gzip")
```
smcclure17, line 612: Same as above comment, using pandas `to_csv` to take advantage of the `compression="gzip"` argument

tests/libs/datasets/timeseries_test.py
```diff
@@ -584,7 +584,8 @@ def test_write_read_wide_dates_csv_compare_literal(tmpdir):
 
     # Compare written file with a string literal so a test fails if something changes in how the
     # file is written. The literal contains spaces to align the columns in the source.
-    assert pointer.path_wide_dates().read_text() == (
+    wide_dates_data = pd.read_csv(pointer.path_wide_dates(), dtype=object).to_csv(index=False)
```
mikelehen, line 587: I think the comment above this is now incorrect, but given the comment I think maybe this test was intentionally not parsing the CSV with pandas, and so I think we should either:
1. Remove this test if we don't think it's valuable.
2. Read the file and gunzip it, but don't parse it through pandas to preserve the original intention of the test.

I'm fine with either one.

mikelehen (approved):
Looks good with one test nit.

tests/libs/datasets/timeseries_test.py
smcclure17: ah, yeah. I think I'll remove it if we're okay with that 🙏

PR #1298, opened 2022-04-20
Remove DC from HSA and use state-level data to fix hospital metrics

smcclure17:


libs/metrics/top_level_metrics.py
```diff
@@ -297,11 +296,11 @@ def calculate_covid_patient_ratio(data: pd.DataFrame, region: Region):
         cdc_covid_patient_ratio = data[CommonFields.BEDS_WITH_COVID_PATIENTS_RATIO_HSA]
         first_cdc_ratio_index = cdc_covid_patient_ratio.first_valid_index()
     else:
-        cdc_covid_patient_ratio = pd.Series(dtype=float)
+        cdc_covid_patient_ratio = EMPTY_TS
```
mikelehen, line 299: Out of curiosity, does this actually behave differently or was it just cleaner?

libs/metrics/top_level_metrics.py
```diff
@@ -332,16 +331,17 @@ def calculate_weekly_admissions_per_100k(
...
+    if region.is_county():
         # Counties in the Northern Mariana Islands are not mapped to an HSA, so
         # they have no hsaPopulations. For these instances do not try and
-        # calculate a metric.
-        if hsa_population is None:
-            can_admissions_per_100k = pd.Series(dtype=float)
+        # calculate a metric. DC has state-level data, not county-level, so the HSA columns
+        # don't exist for it.
+        if hsa_population is None or region.fips == "11001":
+            can_admissions_per_100k = EMPTY_TS
```
mikelehen, line 344: I'm confused.  Don't we want to hit the else block on line 348 so we use the HSS state-level data for DC?  So line 338 should be `if region.is_county() and region.fips != "11001":` like we did for covid patients ratio ([line 303](https://github.com/covid-projections/covid-data-model/pull/1298/files#diff-6d0e13950a596647213010e005f51de2cf5bd90346eab74741918df924fd35d4R303))?

Or am I missing something?

mikelehen (changes requested):
I may be confused, but I can't quite make sense of the covid admissions logic change.  Also, if you have it handy, can you drop a link to a snapshot with these changes?

libs/metrics/top_level_metrics.py
smcclure17: 🤔 Yeah, that seems to make more sense--and appears to be working locally. I'm creating a snapshot to verify now

libs/metrics/top_level_metrics.py
smcclure17: This behaves the same, I just didn't see that `EMPTY_TS` existed when first making these changes

smcclure17:
@mikelehen here is a new snapshot built off of 728c98034eaa61d8d666907867495d8f976a12e0: https://covid-projections-git-bump-snapshot-3325-2141-covidactnow.vercel.app/?s=33252141

I notice that the DC metro is missing admissions data despite all of its counties having the necessary data--this is something I'll look into, but I think it makes sense to merge this in the meantime (if that sounds alright to you).

libs/metrics/top_level_metrics.py
```diff
@@ -332,16 +331,17 @@ def calculate_weekly_admissions_per_100k(
...
         first_cdc_admissions_index = None
 
     # Use HSA-level data for counties only.
-    if region.level == AggregationLevel.COUNTY:
+    if region.is_county() and region.fips != "11001":
         # Counties in the Northern Mariana Islands are not mapped to an HSA, so
         # they have no hsaPopulations. For these instances do not try and
-        # calculate a metric.
+        # calculate a metric. DC has state-level data, not county-level, so the HSA columns
+        # don't exist for it.
```
mikelehen, line 342: Can you rework this comment now that the DC logic has moved up?

mikelehen (approved):
Looks good with one comment nit.  Yeah, we can figure out the metro thing separately.

mikelehen (approved):
Oh, also, can you revert [data/multiregion-wide-dates.csv](https://github.com/covid-projections/covid-data-model/pull/1298/files#diff-23ed6c77bfb09193551c21a7f826b7e47be133f4bb4b6b4aef70af29b6c31e6e) ?

PR #1304, opened 2022-05-04
Remove data/multiregion-wide-dates.csv

smcclure17:
I should have deleted this in #1296 as it has been replaced by `data/multiregion-wide-dates.csv.gz`


mikelehen (approved):
🎉

PR #1308, opened 2022-05-10
Fix model artifact rename to work across devices.

mikelehen:
This fixes an error I ran into when I was doing some perf testing on a new github actions runner.  When I set up the runner, I gave it a small (10GB) OS drive and a large drive for the action runner / data file stuff.  But this caused the `PYSEIR_ARTIFACT_SNAPSHOT` option to not work because it ended up trying to `rename` a file from one device to another, which apparently isn't supported.

```
OSError: [Errno 18] Invalid cross-device link: '/tmp/tmppf9_1nlo/data/api-results-3355' -> '/mnt/data/covid-data-model/covid-data-model/api-results-3355'
```

### DIFF
--- a/libs/github_utils.py
+++ b/libs/github_utils.py
@@ -3,6 +3,7 @@ import shutil
 import click
 import tempfile
 import io
+import shutil
 
 import zipfile
 import pathlib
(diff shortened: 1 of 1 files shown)


smcclure17 (approved):
huh interesting--thanks for the fix!

PR #1312, opened 2022-05-17
Add PA state and counties to outlier exclusion list

smcclure17:


data/region-overrides.json
```diff
@@ -807,15 +807,6 @@
...
       "context": "NY County, NY big assumed backlog on 5/9/2022"
     },
-    {
-      "include": "region",
-      "metric": "metrics.caseDensity",
-      "blocked": true,
-      "start_date": "2022-05-10",
-      "end_date": "2022-05-10",
-      "region": "PA",
-      "context": "PA staet-level cases spike on 5/11/2022"
```
mikelehen, line 817: Oh, haha. I guess we intentionally removed this as an outlier.  I somehow missed that and assumed it was outlier detection. 🤦

libs/datasets/outlier_detection.py
```diff
@@ -14,8 +14,12 @@
 
 MultiRegionDataset = timeseries.MultiRegionDataset
 
-# Florida biweekly case reporting causes outlier detection to trigger inadvertently.
-STATES_TO_EXCLUDE_CASES_DEATHS = [RegionMask(AggregationLevel.COUNTY, states=["FL"])]
+STATES_TO_EXCLUDE_CASES_DEATHS = [
+    # FL: Biweekly case reporting causes outlier detection to trigger inadvertently.
+    # PA: Switching to weekly case reporting caused outlier detection to trigger inadvertently.
```
mikelehen, line 19: Per the comment above, it's possible we actually don't need to do this.  But I'm fine either way. 🤷

mikelehen (approved):

data/region-overrides.json
smcclure17: oops 😭 looks like I added that but I have no recollection of it--sorry for not realizing that hahah 🤦

libs/datasets/outlier_detection.py
smcclure17: Yeah, agreed this probably isn't necessary--I'll remove all the blocks and create a snapshot to see what that looks like

data/region-overrides.json
mikelehen: Heh. Nah, it's 100% my bad for not noticing it when I was diagnosing the issue of low cases. All good.

PR #1315, opened 2022-06-11
Parallelize Update Combined Datasets

smcclure17:
Update datasets in parallel by partitioning the bulk MultiRegionDataset into individual state shards. Currently, this runs the "Update Combined Datasets" action in ~38 minutes (with refreshing datasets) compared to the previous ~80-100 minutes.

Snapshot built using this PR: https://covid-projections-git-bump-snapshot-3560-2346-covidactnow.vercel.app/internal/compare/?left=3558&locations=0&metric=9&right=3560&sort=2


libs/parallel_utils.py
```diff
@@ -36,7 +36,7 @@ def parallel_map(func: Callable[[T], R], iterable: Iterable[T]) -> Iterable[R]:
         # But that might not be enough. Also make sure we don't spawn more than 32 processes (the
         # build machine is 96-core)
         processes = min(os.cpu_count(), 32)
-        with multiprocessing.Pool(maxtasksperchild=1, processes=processes) as pool:
+        with get_context("spawn").Pool(maxtasksperchild=1, processes=processes) as pool:
```
smcclure17, line 39: I had issues using the `fork` method (both locally and on the runner) where the program would hang indefinitely after completing all the parallel processes, but using `spawn` fixed those issues. 

I see the TODOs and our issue with using `spawn` with pandarallel (https://github.com/nalepae/pandarallel/issues/72), but I _think_ because using `get_context()` only sets the method inside the context manager this should be compatible (and I haven't run into any issues so far.)

cli/data.py
```diff
@@ -88,149 +57,21 @@ def main():
...
 def update(
     aggregate_to_country: bool,
     print_stats: bool,
     refresh_datasets: bool,
-    state: Optional[str],
-    fips: Optional[str],
+    states: Optional[List[str]],
 ):
-    """Updates latest and timeseries datasets to the current checked out covid data public commit"""
-    path_prefix = dataset_utils.DATA_DIRECTORY.relative_to(dataset_utils.REPO_ROOT)
```
smcclure17, line 101: Most of the below code has been moved to `libs/datasets/dataset_orchestrator.py`

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+        multiregion_dataset.to_compressed_pickle(dataset_utils.COMBINED_RAW_PICKLE_GZ_PATH)
+    else:
+        multiregion_dataset = timeseries.MultiRegionDataset.from_compressed_pickle(
+            dataset_utils.COMBINED_RAW_PICKLE_GZ_PATH
+        )
+    return multiregion_dataset
+
+
+def load_datasets_by_field(
```
smcclure17, line 300: There's probably work that could be done to speed up the dataset loading process (which takes ~15-20 min atm), but I didn't look into that in this PR

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+    manual_filter,
+    dataset_utils,
+    timeseries,
+    combined_datasets,
+)
+
+TailFilter = tail_filter.TailFilter
+_log = structlog.get_logger()
+
+# Putting this in dataset_utils caused a circular import with libs.pipeline
```
mikelehen, line 43: This comment could use a bit more context, like, "Ideally this should probably go in dataset_utils, but that causes a circular import with libs.pipeline, so leaving here for now."

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+        return OneStateDataset(
+            multiregion_dataset=dataset, state=pipeline.Region.from_state(states[0])
+        )
+
+
+@dataclass
+class MultiRegionOrchestrator:
+    """Class to orchestrate dataset updates through state-focused parallelization."""
+
+    regions: Iterable[OneStateDataset]
```
mikelehen, line 88: nit: Given we have a "Region" class used pervasively through the codebase, calling this "regions" is a bit confusing.  Could we call it "state_datasets" or something?

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+                f"for a single state. Data for {len(set(states))} found."
+            )
+        return OneStateDataset(
+            multiregion_dataset=dataset, state=pipeline.Region.from_state(states[0])
+        )
+
+
+@dataclass
+class MultiRegionOrchestrator:
+    """Class to orchestrate dataset updates through state-focused parallelization."""
```
mikelehen, line 86: It's hard to guess what this class does from its name.  Based on this comment, perhaps a better name would be `DatasetUpdateOrchestrator` or something?  I don't love the name "orchestrator" (it's not entirely obvious what an orchestrator does), so we could also go `ParallelDatasetUpdater` or similar.  Arguably the fact that it does things in parallel is an implementation detail that the consumer of the class doesn't need to know about, so it could maybe even just be `DatasetUpdater`.  Up to you!

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+    ):
+        processed_regions = parallel_map(self._build_region_timeseries, self.regions)
+        _log.info("Finished processing individual regions...")
+        return self._combine_regions(
+            list(processed_regions),
+            aggregate_to_country=aggregate_to_country,
+            generate_cbsas=generate_cbsas,
+        )
+
+    def _combine_regions(
```
mikelehen, line 127: nit: I might move _build_region_timeseries up above _combine_regions since it's used first in the above code?  Makes it a bit easier to read the code top-to-bottom.

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+
+        if aggregate_to_country:
+            _log.info("Starting country aggregation...")
+            multiregion_dataset = custom_aggregations.aggregate_to_country(
+                multiregion_dataset, reporting_ratio_required_to_aggregate=DEFAULT_REPORTING_RATIO
+            )
+            multiregion_dataset.print_stats("Aggregate to country")
+        return multiregion_dataset
+
+    def _build_region_timeseries(self, region_dataset: OneStateDataset):
```
mikelehen, line 174: nit: Perhaps name this state_dataset if you rename the other variable to state_datasets.

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+        state = region_dataset.state
+        multiregion_dataset = region_dataset.multiregion_dataset
+
+        # TODO: maybe filter overrides to match only the current region?
+        # Not sure if that will help performance but worth a shot.
+        region_overrides_config = manual_filter.transform_region_overrides(
+            self.region_overrides, self.cbsa_aggregator.cbsa_to_counties_region_map
+        )
```
mikelehen, line 182: nit: It looks like we'll be calling `transform_region_overrides` for every state even though the parameters aren't state-dependent.  Should we just calculate this once in the constructor or something?

mikelehen: Though maybe your TODO is hinting that you might want to make it state-dependent...?  Not a big deal either way.

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+        before_manual_filter = multiregion_dataset
+        multiregion_dataset = manual_filter.run(multiregion_dataset, region_overrides_config)
+        manual_filter_touched = manual_filter.touched_subset(
+            before_manual_filter, multiregion_dataset
+        )
+        manual_filter_touched.write_to_wide_dates_csv(
+            dataset_utils.MANUAL_FILTER_REMOVED_WIDE_DATES_CSV_PATH,
+            dataset_utils.MANUAL_FILTER_REMOVED_STATIC_CSV_PATH,
+        )
```
mikelehen, line 191: Hrm...  I think this might end up overwriting these files for every state in parallel.  Since we don't use Tom's data dashboard anymore, we could potentially just get rid of these files.  Else I think the right solution is probably to store the touched records in the state_dataset somehow and then combine / write them to the CSV after all states have been processed.

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+        if self.print_stats:
+            multiregion_dataset.print_stats("manual filter")
```
mikelehen, line 193: Hrm. I assume this will log stats once for each state (50+ times)?  Not 100% sure if we want that.  At minimum we probably would want to include the state name so we can differentiate them?

Same feedback on below print_stats calls.

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+        multiregion_dataset = timeseries.drop_observations(
+            multiregion_dataset, after=datetime.datetime.utcnow().date()
+        )
+
+        multiregion_dataset = outlier_detection.drop_tail_positivity_outliers(multiregion_dataset)
+        if self.print_stats:
+            multiregion_dataset.print_stats(f"drop_tail {multiregion_dataset.location_ids}")
```
mikelehen, line 200: Is this going to be a long list of county FIPS codes or something?  Perhaps just include the state code?

libs/datasets/dataset_orchestrator.py
```diff
@@ -0,0 +1,318 @@
...
+        multiregion_dataset = MultiRegionDataset(
+            timeseries_bucketed=timeseries_df, static=static_df, tag=tag
+        )
+
+        if generate_hsas:
+            _log.info("Starting HSA aggregation...")
+            hsa_aggregator = statistical_areas.CountyToHSAAggregator.from_local_data()
+            multiregion_dataset = hsa_aggregator.aggregate(multiregion_dataset)
+
+        if generate_cbsas:
```
mikelehen, line 158: Out of curiosity do we ever not set generate_hsas and generate_cbsas to True?  Do we need to have these flags? 🤷

mikelehen (approved):
Awesome! 👏 

Looks good except some minor nits and I think we may have an issue with clobbering the manual_filter_removed-wide-dates.csv files with each state. 😬

libs/datasets/dataset_orchestrator.py
smcclure17: Yeah good feedback--I've ripped these out in efa6f92d89d1b944afb592940ea1de51cd46b118 and have replaced them with a single `print_stats` at the end of each state run

libs/datasets/dataset_orchestrator.py
smcclure17: They can be helpful for testing things quickly, as the cbsas/hsa aggregations make up a fair amount of the time of the whole pipeline, so while we don't ever use them when actually running the pipeline I'm biased towards leaving them in

libs/datasets/dataset_orchestrator.py
smcclure17: ooop thank you :( The easiest way I found to deal with this was to just return this filtered dataset alongside the other data, and combine them in `build_and_combine_regions` (see efa6f92d89d1b944afb592940ea1de51cd46b118)

libs/datasets/dataset_orchestrator.py
smcclure17: Good call, I've added this as a cached_property in efa6f92d89d1b944afb592940ea1de51cd46b118, I'm not sure it's worth the effort to try and make this state-dependent at the moment

libs/datasets/timeseries.py
```diff
@@ -459,7 +459,9 @@ def _check_timeseries_wide_vars_index(timeseries_index: pd.MultiIndex, *, bucket
         # timeseries.index order is important for _timeseries_latest_values correctness.
         assert timeseries_index.names == [CommonFields.LOCATION_ID, CommonFields.DATE]
     assert timeseries_index.is_unique
-    assert timeseries_index.is_monotonic_increasing
+    # TODO: The following fails during manual_filter runs, despite the resulting datasets
+    # having monotonic increasing indexes. Maybe related to the TODO in __post_init__?
+    # assert timeseries_index.is_monotonic_increasing
```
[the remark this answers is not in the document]
smcclure17: I'm not 100% sure either, and since the resulting datasets are monotonic increasing anyway, I've opted to just remove it

libs/datasets/dataset_orchestrator.py
mikelehen: Ah, gotcha! 👍

PR #1323, opened 2022-08-30
Disable sentry performance transactions.

mikelehen:
I've been meaning to do this for a while so we don't use up our Sentry transaction quota every month and get email notifications... 🤷


smcclure17 (approved):
Ah, nice! Yeah, it will be nice not to be pinged about this anymore 

Thanks for this! 🙏

PR #1324, opened 2022-08-31
Fix provider.region in serverless.yml and pin serverless version.

mikelehen:
Fixes issue encountered in https://github.com/covid-projections/covid-data-model/runs/8100141218?check_suite_focus=true

It seems like serverless v3 broke us.  I fixed one issue, but hit a different issue, so I'm pinning at 2.72.3.

### DIFF
--- a/api/awsauth/serverless.yml
+++ b/api/awsauth/serverless.yml
@@ -41,6 +41,7 @@ plugins:
 
 provider:
   name: aws
+  region: us-east-1
   runtime: python3.7
 
   stage: ${opt:stage, 'dev'}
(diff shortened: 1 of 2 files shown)


api/awsauth/serverless.yml
```diff
@@ -41,6 +41,7 @@ plugins:
 
 provider:
   name: aws
+  region: us-east-1
```
mikelehen, line 44: This fixes the original `${self:provider.region}` error, but after that was fixed I ran into a permissions error that I didn't want to untangle. (https://github.com/covid-projections/covid-data-model/runs/8105515757?check_suite_focus=true) 

So I just pinned the version.

smcclure17 (approved):

smcclure17:
Ah interesting, pinning the serverless version sounds good to me!

thanks for digging into this 🙏 🙏

PR #1325, opened 2022-08-31
Fix serverless version for prod deploy.

mikelehen:
I missed a spot where we need to pin the serverless version (for the prod deploy).


smcclure17 (approved):
lgtm! thanks for tracking all these down

PR #1329, opened 2022-10-26
Revert duplicates hack

smcclure17:
I've:
- Reverted the database to a backup from Oct. 22nd, 2022 (before the duplicates issue began)
- Ran the HHS facility-level scraper and generated a new parquet file
- Tested an [Update Combined Datasets](https://github.com/covid-projections/covid-data-model/actions/runs/3330302904) run off of my testing branch. This made it past the point where it would previously break


mikelehen (approved):
Looks good! Thanks for dealing with all of this.

PR #1330, opened 2022-10-26
Remove Out of Date Data Disclaimers (as of 26 Oct 2022)

BrettBoval:
I audited the data disclaimers and found the following that should be removed:

- Louisiana state and sub-regions references a hurricane from 13 months ago
- Hawaii has duplicate overrides that are completely redundant
- Minnesota has a disclaimer that is not noticeable in the data (as currently masked)
- Idaho has caught up on the case backlog from earlier in the year
- Wisconsin has caught up on the case backlog from earlier in the year

Notes:

- I tested these changes out in the covid-projections frontend. There is a github action that copies the file from covid-data-model to covid-projections when building a new snapshot. So I am copying those local changes back over to this repo and committing them here to be rolled up in the next update.
- I only looked at non-blocking data disclaimers. To test the ongoing data blocks (such as vaccination data from Hawaii) would require rebuilding an api snapshot, which was out of scope for today.

### Please Check
- [ ] Are tests passing?

### DIFF
--- a/data/region-overrides.json
+++ b/data/region-overrides.json
@@ -215,16 +215,6 @@
     },
     {
       "include": "region-and-subregions",
-      "metric": "metrics.caseDensity",
-      "blocked": false,
-      "start_date": "",
-      "end_date": "",
-      "region": "LA",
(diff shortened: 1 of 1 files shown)


smcclure17 (approved):
Looks great! thanks for digging through these 🙏 🚀

PR #1335, opened 2022-11-30
Replace Hubspot API key

smcclure17:
#766  Updates Hubspot to use a Bearer auth token following: https://developers.hubspot.com/docs/api/migrate-an-api-key-integration-to-a-private-app

Appears to be working per: https://app.hubspot.com/private-apps/8617439/1285491/logs/api


mikelehen (approved):
Looks good to me.  Thanks so much for taking care of this!

PR #1338, opened 2022-12-07
Start / stop the GCE runner before/after use to reduce costs.

mikelehen:
I tested this approach by creating a test action in https://github.com/covid-projections/covid-data-model/pull/1337.  See an example run [here](https://github.com/covid-projections/covid-data-model/actions/runs/3635639729).

### DIFF
--- a/.github/workflows/deploy_api.yml
+++ b/.github/workflows/deploy_api.yml
@@ -1,12 +1,8 @@
 name: Build & Publish API snapshot to data.covidactnow.org
 
-#######################################################################################
-###
-### YOU DON'T NEED TO MODIFY THIS FILE TO KICK OFF A SNAPSHOT FROM YOUR BRANCH ANYMORE.
-###
-### See: https://github.com/covid-projections/covid-data-model/#api-snapshots
-###
-#######################################################################################
+# Use a concurrency group to make sure we don't try to have multiple workflows
+# run with the hosted runner at the same time.
+concurrency: gce-runner
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_api.yml
```diff
@@ -1,12 +1,8 @@
 name: Build & Publish API snapshot to data.covidactnow.org
 
-#######################################################################################
-###
-### YOU DON'T NEED TO MODIFY THIS FILE TO KICK OFF A SNAPSHOT FROM YOUR BRANCH ANYMORE.
-###
-### See: https://github.com/covid-projections/covid-data-model/#api-snapshots
-###
-#######################################################################################
```
mikelehen, line 9: This comment doesn't seem necessary anymore. 😆

.github/workflows/deploy_api.yml
```diff
@@ -50,8 +46,27 @@ env:
...
+          credentials_json: "${{ secrets.GCE_ADMIN_SERVICE_ACCOUNT }}"
+
+      - name: "Set up Cloud SDK"
+        uses: "google-github-actions/setup-gcloud@v1"
+
+      - name: "Start ${{env.GCE_INSTANCE}} VM."
+        run: "gcloud compute instances start --zone ${{env.GCE_ZONE}} ${{env.GCE_INSTANCE}}"
+
   build-and-publish-snapshot:
+    needs: "start-runner"
```
mikelehen, line 69: By default jobs run in parallel but we want them to be sequential.

.github/workflows/deploy_api.yml
```diff
@@ -146,3 +161,19 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.CAN_ROBOT_PERSONAL_ACCESS_TOKEN }}
       run: ./covid-data-model/tools/maybe-trigger-label-api.sh ${{env.SNAPSHOT_ID}} ${{env.COVID_DATA_MODEL_REF}}
+
+  stop-runner:
+    if: ${{ always() }}
```
mikelehen, line 166: We want stop-runner to run even if the build-and-publish-snapshot job fails.

.github/workflows/deploy_api.yml
```diff
@@ -1,12 +1,8 @@
...
-#######################################################################################
-###
-### YOU DON'T NEED TO MODIFY THIS FILE TO KICK OFF A SNAPSHOT FROM YOUR BRANCH ANYMORE.
-###
-### See: https://github.com/covid-projections/covid-data-model/#api-snapshots
-###
-#######################################################################################
+# Use a concurrency group to make sure we don't try to have multiple workflows
+# run with the hosted runner at the same time.
+concurrency: gce-runner
```
mikelehen, line 5: Not 100% sure if we strictly need this.  Since there's only one gce-runner, everything runs serially on the runner, but since now we have some self-hosted jobs within this workflow I think there could be weird race conditions.  Like one VM could be trying to start the runner VM while another one is trying to stop the runner VM or something.  By using a concurrency group, I'm hoping that is avoided.

smcclure17: Yeah, good call 🙏 can't hurt to be safe!

smcclure17 (approved):
Nice! this looks great! 🎉 

After this is merged I imagine we'll want to delete `can-actions-runner-2` and stop `can-actions-runner`?
image

mikelehen:
Yeah.  `can-actions-runner-2` has actually been sitting around for a while. I probably created it and forgot?  I am not sure how much it costs us to have a VM that's not being used. I think it's just the storage.  But I can delete it once I'm confident we're done testing stuff.

I have shut down `can-actions-runner` and kicked off an update datasets + build snapshot job.  Will monitor to make sure it works as expected.

smcclure17:
Oh ok! I might have created it way back when testing something, too, not sure. Either way, yeah I agree, we can get rid of it once we're finished testing 🙏

PR #1345, opened 2022-12-15
Add state testing source to pull in CA data

smcclure17:
Tested locally with output: https://gist.github.com/smcclure17/d38268cf5ad6eb054b43508e64705a95

### DIFF
--- a/libs/datasets/sources/state_provided_testing_dataset.py
+++ b/libs/datasets/sources/state_provided_testing_dataset.py
@@ -6,7 +6,7 @@ from libs.datasets.timeseries import MultiRegionDataset
 
 
 class StateProvidedTestingDataset(data_source.CanScraperBase):
-    """Data source connecting to ."""
+    """Data source connecting to state health department sourced testing data."""
 
     SOURCE_TYPE = "StateProvidedTesting"
 
--- a/libs/datasets/sources/state_provided_testing_dataset.py
+++ b/libs/datasets/sources/state_provided_testing_dataset.py
@@ -13,17 +13,17 @@ class StateProvidedTestingDataset(data_source.CanScraperBase):
     VARIABLES = [
         ccd_helpers.ScraperVariable(
             variable_name="pcr_tests_positive",
-            measurement="cumulative",
+            measurement="new",
             provider="state",
             unit="specimens",
-            common_field=CommonFields.POSITIVE_TESTS,
+            common_field=CommonFields.NEW_POSITIVE_TESTS_VIRAL,
         ),
         ccd_helpers.ScraperVariable(
             variable_name="pcr_tests_total",
-            measurement="cumulative",
+            measurement="new",
             provider="state",
             unit="specimens",
-            common_field=CommonFields.TOTAL_TESTS,
+            common_field=CommonFields.NEW_TOTAL_TESTS_VIRAL,
         ),
     ]
 
@@ -33,8 +33,14 @@ class StateProvidedTestingDataset(data_source.CanScraperBase):
         # Calculate 7-day rolling sum to match CDC test positivity convention.
         # See "Test Percent Positivity Metric" at
         # https://data.cdc.gov/Public-Health-Surveillance/Weekly-COVID-19-County-Level-of-Community-Transmis/dt66-w6m6/
-        ts[CommonFields.TOTAL_TESTS_7D] = ts[CommonFields.TOTAL_TESTS].rolling(window=7).sum()
-        ts[CommonFields.POSITIVE_TESTS_7D] = ts[CommonFields.POSITIVE_TESTS].rolling(window=7).sum()
-        ts = ts.drop(columns=[CommonFields.TOTAL_TESTS, CommonFields.POSITIVE_TESTS])
+        ts[CommonFields.TOTAL_TESTS_VIRAL_7D] = (
+            ts[CommonFields.NEW_TOTAL_TESTS_VIRAL].rolling(window=7).sum()
+        )
+        ts[CommonFields.POSITIVE_TESTS_VIRAL_7D] = (
+            ts[CommonFields.NEW_POSITIVE_TESTS_VIRAL].rolling(window=7).sum()
+        )
+        ts = ts.drop(
+            columns=[CommonFields.NEW_TOTAL_TESTS_VIRAL, CommonFields.NEW_POSITIVE_TESTS_VIRAL]
+        )
 
         return dataclasses.replace(dataset, timeseries_bucketed=ts)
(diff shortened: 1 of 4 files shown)


libs/datasets/sources/state_provided_testing_dataset.py
```diff
@@ -0,0 +1,40 @@
...
+
+    SOURCE_TYPE = "StateProvidedTesting"
+
+    VARIABLES = [
+        ccd_helpers.ScraperVariable(
+            variable_name="pcr_tests_positive",
+            measurement="cumulative",
+            provider="state",
+            unit="specimens",
+            common_field=CommonFields.POSITIVE_TESTS,
```
mikelehen, line 19: I think we could / should be using POSITIVE_TESTS_VIRAL and TOTAL_TESTS_VIRAL since these are labeled as PCR tests.  Any reason you didn't?

Changing it might cause a conflict with the existing `DivisionMethod` in libs/metrics/test_positivity.py though?  So if this is already tested and you want to ship it as-is, I'm fine with that.

libs/datasets/sources/state_provided_testing_dataset.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def make_dataset(self) -> MultiRegionDataset:
+        dataset = super().make_dataset()
+        ts = dataset.timeseries_bucketed
+        # Calculate 7-day rolling sum to match CDC test positivity convention.
+        # See "Test Percent Positivity Metric" at
+        # https://data.cdc.gov/Public-Health-Surveillance/Weekly-COVID-19-County-Level-of-Community-Transmis/dt66-w6m6/
+        ts[CommonFields.TOTAL_TESTS_7D] = ts[CommonFields.TOTAL_TESTS].rolling(window=7).sum()
+        ts[CommonFields.POSITIVE_TESTS_7D] = ts[CommonFields.POSITIVE_TESTS].rolling(window=7).sum()
+        ts = ts.drop(columns=[CommonFields.TOTAL_TESTS, CommonFields.POSITIVE_TESTS])
```
mikelehen, line 38: Oh, wow.  Nice work bending this to match the CDC methodology.  Arguably we could be doing this downstream in libs/metrics/test_positivity.py...  e.g. bake it into `DivisionMethod`.  But if this is already tested and you want to ship it as-is, I'm fine with that.

libs/metrics/test_positivity.py
```diff
@@ -48,11 +48,10 @@ class Method(ABC):
...
     # This method needs a timeseries to have at least one real value within recent_days for it to
     # be considered recent / not stale. Stale timeseries are dropped.
-
-    # TODO 12/13/2022: This is a legacy filtering system. Metrics are excluded from the website
-    # further down the pipeline if the data is stale. Typically, we don't exclude the entire
-    # timeseries if the data is stale.
-    recent_days: int = 60
+    # This is needed b/c the downstream process for choosing test positivity sources looks for the first
+    # non-empty timeseries in the list of sources. We want to choose sources with recent data, so we drop
+    # stale timeseries here.
```
mikelehen, line 53: Oh, interesting.  Thanks for improving the comment!

mikelehen (approved):
Oh, this was a bit more involved than I realized.  Thanks for plumbing it through!

I think this is good as-is, though there may be opportunity to clean it up a tad and use our fields in a more standard way.  Left some comments.  I'm okay with shipping as-is though.

libs/datasets/sources/state_provided_testing_dataset.py
smcclure17: I realized here that I'm taking a rolling sum of cumulative values which doesn't really make sense... So I need to either change this or use the new/daily data we pull from the scraper. In that process I'll look into doing that in the `DivisionMethod` create method if it's straightforward, but if not I might just opt to continue to do that here if we're okay with that

libs/datasets/sources/state_provided_testing_dataset.py
smcclure17: Ah, no reason I just didn't fully understand the defns I think--will update thanks!

libs/datasets/sources/state_provided_testing_dataset.py
mikelehen: Oh, right. Good catch.  Sounds good.

smcclure17: changed this to use daily data in e9aee489a0cdb8f3b975364762431548b76f290b, and will sanity check the output in the AM

smcclure17: Manually calculating the results for LA county with,

```python
   import pandas as pd
   # dataset source 
   df = pd.read_csv("https://data.chhs.ca.gov/dataset/f333528b-4d38-4814-bebb-12db1f10f535/resource/046cdd2b-31e5-4d34-9ed3-b48cdbc4be7a/download/covid19cases_test.csv")
   df = df[["date", "area", "positive_tests", "total_tests"]].set_index(["date", "area"]
   la = df.xs("Los Angeles", level="area")
   la["ptw"] = la["positive_tests"].rolling(window=7).sum()
   la["ttw"] = la["total_tests"].rolling(window=7).sum()
   la["ratio"] = la["ptw"] / la["ttw"]
```

results in: 
```python
la.dropna().tail(7)
```
```
            positive_tests  total_tests      ptw       ttw     ratio
date                                                                
2022-11-29          4672.0      44176.0  21337.0  186160.0  0.114616
2022-11-30          4325.0      37001.0  22707.0  195489.0  0.116155
2022-12-01          3686.0      31674.0  25587.0  219884.0  0.116366
2022-12-02          2917.0      24334.0  25706.0  224236.0  0.114638
2022-12-03           930.0       6883.0  24299.0  215845.0  0.112576
2022-12-04           165.0       1055.0  22351.0  201935.0  0.110684
2022-12-05            56.0        509.0  16751.0  145632.0  0.115023
```

where `ratio` is the test-positivity metric. This output matches the output of the local/test run, so I feel comfortable trusting these calculations. 

It does look like tests are reported by specimen collection date, so the most recent days have fewer observations (I imagine they get updated over time as more results come in) so this could mildly sway the metric, but I think that's okay.

libs/datasets/sources/state_provided_testing_dataset.py
smcclure17: As for the conflict, I think it wouldn't have been an issue (I ended up creating `NEW_POSITIVE/TOTAL_TESTS_VIRAL` to signify we're using new, not cumulative data) because (if I'm understanding correctly) the existing division method is looking for those variables specifically in the 'positiveTestsViral_totalTestsViral` dataset (which doesn't seem to exist anymore, so maybe we should nuke it?) whereas the new method would look in the newly created dataset

mikelehen: Oh, I see. Heh, yeah. I wasn't sure what that `positiveTestsViral_totalTestsViral` was referring to. 👍

PR #1348, opened 2023-01-09
Pipe bivalent booster data through API

smcclure17:
Adds bivalent CDC-sourced vaccine data to the API. [Example output of MA here](https://gist.githubusercontent.com/smcclure17/bbb6a22b359fe90030bf87b8d4cb0c8f/raw/9b0470ca5c5e7ad2b5cc1e0ec77d239d2403489b/MA.timeseries.json)

### DIFF
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -237,7 +237,7 @@ number of people vaccinated with both the first and second dose.
         None,
         description="Number of individuals who are fully vaccinated and have received a booster (or additional) dose.",
     )
-    vaccinationsBivalentDose: Optional[int] = pydantic.Field(
+    vaccinationsFall2022BivalentBooster: Optional[int] = pydantic.Field(
         None, description="Number of individuals who have received a bivalent vaccine dose.",
     )
     vaccinesAdministered: Optional[int] = pydantic.Field(
@@ -328,8 +328,8 @@ class Annotations(base_model.APIBaseModel):
     vaccinationsAdditionalDose: Optional[FieldAnnotations] = pydantic.Field(
         None, description="Annotations for vaccinationsAdditionalDose"
     )
-    vaccinationsBivalentDose: Optional[FieldAnnotations] = pydantic.Field(
-        None, description="Annotations for vaccinationsBivalentDose"
+    vaccinationsFall2022BivalentBooster: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for vaccinationsFall2022BivalentBooster"
     )
     vaccinesAdministered: Optional[FieldAnnotations] = pydantic.Field(
         None, description="Annotations for vaccinesAdministered"
@@ -371,8 +371,8 @@ class Annotations(base_model.APIBaseModel):
     vaccinationsAdditionalDoseRatio: Optional[FieldAnnotations] = pydantic.Field(
         None, description=("Annotations for vaccinationsAdditionalDoseRatio"),
     )
-    vaccinationsBivalentDoseRatio: Optional[FieldAnnotations] = pydantic.Field(
-        None, description="Annotations for vaccinationsBivalentDoseRatio."
+    vaccinationsFall2022BivalentBoosterRatio: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Annotations for vaccinationsFall2022BivalentBoosterRatio."
     )
 
 
@@ -437,7 +437,7 @@ class Metrics(base_model.APIBaseModel):
             "Ratio of population that are fully vaccinated and have received a booster (or additional) dose."
         ),
     )
-    vaccinationsBivalentDoseRatio: Optional[float] = pydantic.Field(
+    vaccinationsFall2022BivalentBoosterRatio: Optional[float] = pydantic.Field(
         None, description=("Ratio of population that have received a bivalent vaccine dose."),
     )
 
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -369,13 +369,10 @@ class Annotations(base_model.APIBaseModel):
         None, description="Annotations for vaccinationsCompletedRatio"
     )
     vaccinationsAdditionalDoseRatio: Optional[FieldAnnotations] = pydantic.Field(
-        None,
-        description=(
-            "Ratio of population that are fully vaccinated and have received a booster (or additional) dose."
-        ),
+        None, description=("Annotations for vaccinationsAdditionalDoseRatio"),
     )
     vaccinationsBivalentDoseRatio: Optional[FieldAnnotations] = pydantic.Field(
-        None, description="Ratio of population have received a bivalent vaccine dose."
+        None, description="Annotations for vaccinationsBivalentDoseRatio."
     )
 
 
@@ -420,12 +417,12 @@ class Metrics(base_model.APIBaseModel):
 
     bedsWithCovidPatientsRatio: Optional[float] = pydantic.Field(
         ...,
-        description="Ratio of staffed hospital beds that are currently in use by COVID patients. For counties, this is calculated using HSA-level data for the corresponding area.",
+        description="Ratio of staffed hospital beds that are currently in use by COVID patients. For counties, this is calculated using HSA-level data for the corresponding area. For more on HSAs, see https://apidocs.covidactnow.org/data-definitions/#health-service-areas",
     )
 
     weeklyCovidAdmissionsPer100k: Optional[float] = pydantic.Field(
         ...,
-        description="Number of COVID patients per 100k population admitted in the past week. For counties, this is calculated using HSA-level data for the corresponding area.",
+        description="Number of COVID patients per 100k population admitted in the past week. For counties, this is calculated using HSA-level data for the corresponding area. For more on HSAs, see https://apidocs.covidactnow.org/data-definitions/#health-service-areas",
     )
 
     vaccinationsInitiatedRatio: Optional[float] = pydantic.Field(
@@ -441,7 +438,7 @@ class Metrics(base_model.APIBaseModel):
         ),
     )
     vaccinationsBivalentDoseRatio: Optional[float] = pydantic.Field(
-        None, description=("Ratio of population have received a bivalent vaccine dose."),
+        None, description=("Ratio of population that have received a bivalent vaccine dose."),
     )
 
     @staticmethod
--- a/api/can_api_v2_definition.py
+++ b/api/can_api_v2_definition.py
@@ -237,6 +237,9 @@ number of people vaccinated with both the first and second dose.
         None,
         description="Number of individuals who are fully vaccinated and have received a booster (or additional) dose.",
     )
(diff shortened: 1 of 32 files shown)


api/data_overview_builder.py
```diff
@@ -21,23 +21,35 @@ class Fields:
...
-CASES_FIELDS = Fields(["cases", "newCases"], ["caseDensity", "infectionRate", "infectionRateCI90"],)
+CASES_FIELDS = Fields(
+    ["cases", "newCases"],
+    ["caseDensity", "infectionRate", "infectionRateCI90", "weeklyNewCasesPer100k"],
+)
 
 DEATHS_FIELDS = Fields(["deaths"], [])
 
-HOSPITALIZATIONS_FIELDS = Fields(["icuBeds", "hospitalBeds"], ["icuCapacityRatio"],)
+HOSPITALIZATIONS_FIELDS = Fields(
```
smcclure17, line 31: I don't think I realized this in the past, but this builds the output of `api/docs/docs/data-definitions.md`. I didn't add the CDC community metrics here when we added them, and instead added them manually. A bit more on this below...

api/docs/docs/data-definitions.md
```diff
@@ -146,21 +146,21 @@ Fields:
...
+### Weekly Covid Admissions Per100K
 
-  Number of COVID patients per 100k population admitted in the past week. For counties, this is calculated using health service area data for the corresponding area. For more information on HSAs see the [Health Service Area section](https://apidocs.covidactnow.org/data-definitions/#health-service-areas)
+  Number of COVID patients per 100k population admitted in the past week. For counties, this is calculated using HSA-level data for the corresponding area.
 
 **Where to access**  
 * CSV column names: ``metrics.weeklyCovidAdmissionsPer100k``
 * JSON file fields: ``metrics.weeklyCovidAdmissionsPer100k``, ``metricsTimeseries.*.weeklyCovidAdmissionsPer100k``
 
-County- and metro-level hospitalization and ICU actuals are calculated from facility-level data which, suppresses near-zero values due to privacy concerns. As a result, this data may undercount real values, particularly at low levels.
```
smcclure17, line 163: This manually-added copy get clobbered when we run `./run.py api update-schemas`. Thoughts on whether we should manually re-add this each time we generate the docs, or should we just opt to remove it? Or try to work it into the `update-schemas` script...

api/docs/open_api_schema.json
```diff
@@ -1451,6 +1451,18 @@
             ],
```
smcclure17, line 1451: generated by `./run.py api update-schemas`

api/schemas_v2/Actuals.json
```diff
@@ -171,6 +171,18 @@
         }
       ]
```
smcclure17, line 172: All these `api/schemas_v2/**` are generated by `./run.py api update-schemas`

api/can_api_v2_definition.py
```diff
@@ -237,6 +237,9 @@ class Actuals(base_model.APIBaseModel):
         None,
         description="Number of individuals who are fully vaccinated and have received a booster (or additional) dose.",
     )
+    vaccinationsBivalentDose: Optional[int] = pydantic.Field(
+        None, description="Number of individuals who have received a bivalent vaccine dose.",
```
mikelehen, line 241: Sorry, I should have flagged this in the scraper PR too, but as we discussed at the kickoff this morning I think we should probably come up with a better name than "bivalent dose".  Most likely we'll have a new booster sometime later this year that could be bivalent or trivalent or quadvalent or whatever.  So to be more future-proof, we should probably do "Fall 2022 booster" or "Fall 2022 bivalent booster" or something.

api/docs/docs/data-definitions.md
```diff
@@ -146,21 +146,21 @@ Fields:
 
 ### Beds With Covid Patients Ratio
 
-  Ratio of staffed hospital beds that are currently in use by COVID patients. For counties, this is calculated using health service area data for the corresponding area. For more information on HSAs see the [Health Service Area section](https://apidocs.covidactnow.org/data-definitions/#health-service-areas)
+  Ratio of staffed hospital beds that are currently in use by COVID patients. For counties, this is calculated using HSA-level data for the corresponding area.
```
mikelehen, line 149: I guess this is auto-generated now?  Is it worth trying to figure out how to get that link back (i.e. put it in the right place in the python doc annotations)?

api/docs/docs/data-definitions.md
mikelehen: I'd definitely avoid manually adding it every time.  If there's not an easy straightforward way to get it auto-generated, then I'm fine with dropping it.

api/can_api_v2_definition.py
```diff
@@ -368,6 +374,9 @@ class Annotations(base_model.APIBaseModel):
             "Ratio of population that are fully vaccinated and have received a booster (or additional) dose."
         ),
     )
+    vaccinationsBivalentDoseRatio: Optional[FieldAnnotations] = pydantic.Field(
+        None, description="Ratio of population have received a bivalent vaccine dose."
```
mikelehen, line 378: should be "population **that** have" I think?

But actually, should this and vaccinationsAdditionalDoseRatio be changed to "Annotations for vaccination...DoseRatio"? I think maybe we copy/pasted an incorrect description.

api/can_api_v2_definition.py
```diff
@@ -431,6 +440,9 @@ class Metrics(base_model.APIBaseModel):
             "Ratio of population that are fully vaccinated and have received a booster (or additional) dose."
         ),
     )
+    vaccinationsBivalentDoseRatio: Optional[float] = pydantic.Field(
+        None, description=("Ratio of population have received a bivalent vaccine dose."),
```
mikelehen, line 444: Same here, "population **that** have"

datapublic/common_fields.py
```diff
@@ -204,6 +204,8 @@ def list(cls):
     VACCINATIONS_INITIATED_PCT = "vaccinations_initiated_pct", FieldGroup.VACCINES
     VACCINATIONS_ADDITIONAL_DOSE = "vaccinations_additional_dose", FieldGroup.VACCINES
     VACCINATIONS_ADDITIONAL_DOSE_PCT = "vaccinations_additional_dose_pct", FieldGroup.VACCINES
+    VACCINATIONS_BIVALENT_DOSE = "vaccinations_bivalent_dose", FieldGroup.VACCINES
+    VACCINATIONS_BIVALENT_DOSE_PCT = "vaccinations_bivalent_dose_pct", FieldGroup.VACCINES
```
mikelehen, line 208: Similarly here we might want a better name...

libs/datasets/sources/cdc_vaccine_dataset.py
```diff
@@ -49,4 +49,11 @@ class CDCVaccinesDataset(data_source.CanScraperBase):
             provider="cdc",
             common_field=CommonFields.VACCINATIONS_ADDITIONAL_DOSE,
         ),
+        ccd_helpers.ScraperVariable(
+            variable_name="total_vaccine_bivalent_dose",
+            measurement="cumulative",
+            unit="people",
+            provider="cdc",
+            common_field=CommonFields.VACCINATIONS_BIVALENT_DOSE,
+        ),
```
mikelehen, line 58: I don't quite remember the context of cdc_vaccine_dataset.py and cdc_new_ vaccine_dataset.py, do we need this in both?

tests/libs/datasets/vaccine_backfill_test.py
```diff
@@ -53,6 +54,7 @@ def test_derive_vaccine_pct():
     )
 
     ds_out = vaccine_backfills.derive_vaccine_pct(ds_in)
+    print(ds_out.timeseries[["vaccinations_bivalent_dose"]])
```
mikelehen, line 57: intentional?

tests/libs/datasets/vaccine_backfill_test.py
```diff
@@ -61,12 +63,14 @@ def test_derive_vaccine_pct():
...
                 CommonFields.VACCINATIONS_COMPLETED_PCT: [None, 1],
                 CommonFields.VACCINATIONS_ADDITIONAL_DOSE_PCT: [1, 5],
+                CommonFields.VACCINATIONS_BIVALENT_DOSE_PCT: [1, 5],
             },
             region_ma: {**ma_timeseries_in, CommonFields.VACCINATIONS_INITIATED_PCT: [94, 95],},
             region_sf: {CommonFields.VACCINATIONS_COMPLETED_PCT: [10, 95],},
         },
         static_by_region_then_field_name=static_data_map,
     )
+    print(ds_expected.timeseries[["vaccinations_bivalent_dose"]])
```
mikelehen, line 73: ditto

mikelehen (approved):
Minor things, but overall looks good.

Only big thing is the naming...  I think we might want to be more future-looking, but I don't feel super strongly.  We could probably get by with "bivalent booster" for this initial bivalent booster and then choose something better ("Fall 2023" or whatever) for the next one.

api/docs/docs/data-definitions.md
smcclure17: Yeah, this will be auto-generated. I've added the HSA link to the field definitions

api/can_api_v2_definition.py
smcclure17: Ah, yeah

I tend to agree--this name doesn't seem great. I'm not 100% sure what a good/future-proof alternative might be--maybe `vaccinationsFirstBivalentDose` / `vaccinations_first_bivalent_dose`?

api/can_api_v2_definition.py
smcclure17: oops, yeah thanks!

libs/datasets/sources/cdc_vaccine_dataset.py
smcclure17: At this point these could be merged, but `CDCVaccinesDataset` gives us state-level data and the other gives us county-level. 

We would need to make a small change in `can-scrapers` to change the county-level dataset source name, then I think we could merge them, but I don't 100% remember why they were split, so maybe there's more there.

tests/libs/datasets/vaccine_backfill_test.py
smcclure17: :( nope, thanks!

libs/datasets/sources/cdc_vaccine_dataset.py
smcclure17: I've added comments explaining this for now

api/can_api_v2_definition.py
smcclure17: `vaccinationsFall2022Booster` / `vaccinationsFall2022BivalentBooster` both sound good to me, too

libs/datasets/sources/cdc_vaccine_dataset.py
mikelehen: Ah, right.  That makes sense.  Thanks for adding comments.  No need to rework.

PR #1351, opened 2023-01-10
Point frontend snapshot trigger to renamed repository

smcclure17:
and other miscellaneous renamings.
 
The `deploy_api.yml` workflow stopped triggering the frontend workflow after the repo name change (see #notifications-daily-deploys in Slack)

### DIFF
--- a/tools/maybe-trigger-web-snapshot-update.sh
+++ b/tools/maybe-trigger-web-snapshot-update.sh
@@ -2,7 +2,7 @@
 # maybe-trigger-web-snapshot-update.sh
 #
 # Helper script called at the end of the deploy_api.yml workflow to potentially
-# trigger the update-snapshot.yml workflow in the covid-projections repo to
+# trigger the update-snapshot.yml workflow in the covid-act-now-website repo to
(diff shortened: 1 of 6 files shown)


tools/maybe-trigger-web-snapshot-update.sh
```diff
@@ -53,9 +53,9 @@ execute () {
   curl -H "Authorization: token $GITHUB_TOKEN" \
       --request POST \
       --data "{ \"ref\": \"${BRANCH}\", \"inputs\": { \"snapshot_id\": \"${SNAPSHOT_ID}\" } }" \
-      https://api.github.com/repos/act-now-coalition/covid-projections/actions/workflows/update-snapshot.yml/dispatches
+      https://api.github.com/repos/act-now-coalition/covid-act-now-website/actions/workflows/update-snapshot.yml/dispatches
```
smcclure17, line 56: This is the only breaking/necessary change here, the rest would redirect automatically, but I figured I'd clean them up while I was here

faihegberg (approved):
LGTM!

PR #1352, opened 2023-01-11
Add demographic data API doc update

smcclure17:
Open to suggestions/edits to copy. This has no related eng/copy changes, so I think we can ship when ready

### DIFF
--- a/api/docs/docs/updates.md
+++ b/api/docs/docs/updates.md
@@ -13,7 +13,7 @@ _2023-01-11_
 
 Unlike the other data in our API, county-level demographic vaccine data is not actively quality assured, and as such, we cannot guarantee that it is of high quality. 
 
-We will continue to provide this data until 04/01/2023 at the earliest, at which point we may consider archiving or removing this data from the API. Please reach out to [data@covidactnow.org](mailto:data@covidactnow.org) with any concerns or questions. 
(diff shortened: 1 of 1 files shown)


api/docs/docs/updates.md
```diff
@@ -7,6 +7,15 @@ description: Updates to the Covid Act Now API.
 
 Updates to the API will be reflected here.
 
+
+### Sunsetting Demographic Vaccine Data
+_2023-01-11_
+
+Unlike the other data in our API, county-level demographic vaccine data is not actively quality assured, and as such, we cannot guarantee that it is of high quality. 
+
+We will continue to provide this data until 04/01/2023 at the earliest, at which point we may consider archiving or removing this data from the API. Please reach out to [data@covidactnow.org](mailto:data@covidactnow.org) with any concerns or questions.
```
mikelehen, line 16: I would probably be more forceful / explicit and say

> We will continue to provide this data until 04/01/2023, at which point we will remove the data from the API.

Nobody will get mad if we don't actually delete it on that date, but it gives us full freedom to do so and the stronger wording will be much more likely to cause somebody to reach out to us if they're using the data.

I'd also be a little tempted to make the date sooner.  Our assumption is that nobody is using it, right?  So it's not like people need time to migrate.

mikelehen (approved):
LGTM with minor suggestions.

api/docs/docs/updates.md
smcclure17: cool, yeah that makes sense -- updated the wording and the date to a ~a month from now

PR #1353, opened 2023-01-17
Reduce extra snapshot builds.

mikelehen:
Compute Engine is our biggest GCP costs so trying to reduce unnecessary runs.  This will also reduce our S3 growth in AWS.  

If for any reason you are not comfortable with removing this, holler.  I don't feel super strongly.

image

### DIFF
--- a/.github/workflows/update_repo_datasets.yml
+++ b/.github/workflows/update_repo_datasets.yml
@@ -5,9 +5,13 @@ name: Update combined datasets
 concurrency: gce-runner
 
 on:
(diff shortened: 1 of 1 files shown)


smcclure17 (approved):
yeah, this sgtm! In the case that this backup is necessary (meaning the scrapers workflow didn't kick off correctly) we typically don't have much new data to ship (because of the failure) so, even in the worst case, I don't think it hurts us too badly. 

Thanks for being proactive with costs!

PR #1355, opened 2023-01-24
use NYT case/death data for jefferson county

smcclure17:


mikelehen (approved):
LGTM.  Don't forget to remove the block too. 😁

smcclure17:
oop, thank you!

PR #1392, opened 2023-12-04
Update to Python 3.10 and bump necessary dependencies

smcclure17:
From [SOW 3](https://docs.google.com/document/d/1WLW3EkmjjS1abJeZGvopjUXNGfMKOO_iLxqeB7fyCuw/edit)

This PR

- Updates us to Python 3.10 (and updates dependencies and code to match)
- Pins the Python versions to 3.10 in our GitHub Action workflows
- Removes the deprecated data-monitoring dash app

Snapshot built off these changes:
- [compare](https://covidactnow-git-bump-snapshot-4861-3608-act-now-coalition.vercel.app/internal/compare/)
- [preview](https://covidactnow-git-bump-snapshot-4861-3608-act-now-coalition.vercel.app/?s=48613608)

PR #1393, opened 2024-02-18
Migrate AWS API Auth app to Python 3.10

smcclure17:
Upgrade the Serverless/AWS app that runs our API from Python 3.7 to Python 3.10

I've split this out from #1392 because the API App is a standalone package/deployment, so it feels right to migrate this separately (and it makes it easier to see what the migration required)


.github/workflows/deploy_aws_api_infrastructure_dev.yml
```diff
@@ -17,21 +17,26 @@ defaults:
   run:
     working-directory: api/awsauth/
 
+env:
```
smcclure17, line 20: This let's us run the action off of branches other than `main` (so we can test the dev environment)

api/awsauth/awsauth/auth_app.py
```diff
@@ -10,9 +10,6 @@
 import pathlib
 import logging
 
-import sentry_sdk
```
smcclure17, line 13: Yank out all the Sentry logging because a) we don't track the alerts anymore, and b) removing Sentry as a dependency helps get us below the max bundle size for CloudFront Lambda deploys

smcclure17:
Definitely-- 

Here is the passing dev API deploy: https://github.com/act-now-coalition/covid-data-model/actions/runs/7945732912

^ This has since been overridden by a deploy from the other branch (#1392), to test that that workflow was passing too.

Here is the latest test run (it ran successfully): https://github.com/act-now-coalition/covid-data-model/actions/runs/7945859366/job/21692966825

